PR #29, opened 2016-09-23
Add a few more R relevant extensions

jimhester:
Not sure if there was a reason the r key was duplicated, if so I can revert that change.


BurntSushi:
I think some other tool (ack? ag?) was using `rr` so I copied that, but I agree that `r` makes more sense, especially since we have a dedicated flag for specifying types. In any case, thank you!

PR #42, opened 2016-09-24
Files with matches

### LINKED ISSUE #26: add -l / --files-with-matches option?
I use the `-l` option to grep / ack / ag / etc quite a lot to list files containing matches for further processing in a shell pipeline. Does it seem reasonable as a thing to add to ripgrep?

andschwa:
Resolves #26 (or at least starts to).

- [x] Needs tests
- [x] Needs docs
- [x] Needs to terminate the search at the first match
- [ ] Needs to be refactored after a review

### DIFF
--- a/src/search_stream.rs
+++ b/src/search_stream.rs
@@ -80,6 +80,7 @@ pub struct Options {
     pub after_context: usize,
     pub before_context: usize,
     pub count: bool,
+    pub files_with_matches: bool,
     pub eol: u8,
     pub invert_match: bool,
     pub line_number: bool,
@@ -92,11 +93,21 @@ impl Default for Options {
             after_context: 0,
             before_context: 0,
             count: false,
+            files_with_matches: false,
             eol: b'\n',
             invert_match: false,
             line_number: false,
             text: false,
         }
+    }
+
+}
+
+impl Options {
+    /// Both --count and --files-with-matches options imply that we should not
+    /// display matches at all.
+    pub fn skip_matches(&self) -> bool {
+        return self.count || self.files_with_matches;
     }
 }
 
@@ -158,6 +169,14 @@ impl<'a, R: io::Read, W: Terminal + Send> Searcher<'a, R, W> {
         self
     }
 
+    /// If enabled, searching will print the path instead of each match.
+    ///
+    /// Disabled by default.
+    pub fn files_with_matches(mut self, yes: bool) -> Self {
+        self.opts.files_with_matches = yes;
+        self
+    }
+
     /// Set the end-of-line byte used by this searcher.
     pub fn eol(mut self, eol: u8) -> Self {
         self.opts.eol = eol;
@@ -193,7 +212,7 @@ impl<'a, R: io::Read, W: Terminal + Send> Searcher<'a, R, W> {
         self.line_count = if self.opts.line_number { Some(0) } else { None };
         self.last_match = Match::default();
         self.after_context_remaining = 0;
-        loop {
+        while !self.terminate() {
             let upto = self.inp.lastnl;
             self.print_after_context(upto);
             if !try!(self.fill()) {
@@ -202,7 +221,7 @@ impl<'a, R: io::Read, W: Terminal + Send> Searcher<'a, R, W> {
             if !self.opts.text && self.inp.is_binary {
                 break;
             }
-            while self.inp.pos < self.inp.lastnl {
+            while !self.terminate() && self.inp.pos < self.inp.lastnl {
                 let matched = self.grep.read_match(
                     &mut self.last_match,
                     &mut self.inp.buf[..self.inp.lastnl],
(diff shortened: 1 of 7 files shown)


andschwa:
@BurntSushi So I *think*, if I'm understanding this code correctly, to terminate early for mmap searching I could just break out of

```rust
        for m in self.grep.iter(self.buf) {
            if self.opts.invert_match {
                self.print_inverted_matches(last_end, m.start());
            } else {
                self.print_match(m.start(), m.end());
            }
            last_end = m.end();
        }
```

t the first match. Is this loop "for matches in lazy iter"? I could use some pointers for sure 😉 (pun totally intended).

For now I'm going to eat some dinner and get to those video games, but this was fun!

BurntSushi:
>  Is this loop "for matches in lazy iter"? I could use some pointers for sure :wink: (pun totally intended).

You got it, that's right. Breaking out of that loop is perfect.

Note that you'll also need to do this in `search_stream.rs`, which is a little less nice.

BurntSushi:
@andschwa Thanks for working on this by the way! Enjoy the video games. :-)

src/search_stream.rs
```diff
@@ -203,6 +213,9 @@ impl<'a, R: io::Read, W: Terminal + Send> Searcher<'a, R, W> {
                 break;
             }
             while self.inp.pos < self.inp.lastnl {
+                if self.opts.files_with_matches && self.match_count > 0 {
+                    break;
```
andschwa: @BurntSushi in my experimentation, this seemed to work correctly to terminate after one match. However, I expected to be able to just break at the first match further down in the loop... but when I tried that I found that the match_count was > 1 at the top of the loop, which was unexpected. I'm missing something.

src/search_stream.rs
```diff
@@ -281,7 +297,7 @@ impl<'a, R: io::Read, W: Terminal + Send> Searcher<'a, R, W> {
 
     #[inline(always)]
     fn print_before_context(&mut self, upto: usize) {
-        if self.opts.count || self.opts.before_context == 0 {
+        if self.opts.count || self.opts.files_with_matches || self.opts.before_context == 0 {
```
andschwa: @BurntSushi I think it might not be a bad refactor to add an `opts.just_file = self.opts.count || self.opts.files_with_matches`, which can replace four instances of this or statement; eh?

src/search_stream.rs
BurntSushi: `match_count` could increase by more than than you might expect if you're doing `inverted` matching. For regular searching, `match_count` should only be incremented by `1` at most on each iteration.

I would probably move this check to right after the call to `read_match`. i.e., `if self.opts.files_with_matches && matched { break; }`.

andschwa: It can't be *right* after it because the first match needs to be handled. But what's odd is that if I put it at the end of the while loop, I get this:

```rust
loop {
            println!("{:?} {}", self.path, self.match_count);
...
while ... {
...
        if self.opts.files_with_matches && matched {
            break;
        }
    }
}
```
```
"src/args.rs" 1
"src/args.rs" 2
"src/args.rs" 3
"src/args.rs" 3
```

I think I have to break out of both loops maybe.

BurntSushi: Oh, yes, indeed you do. The outer loop corresponds to refilling the buffer with data from the file and the inner loop corresponds to iterating over all matches within that buffer. Breaking out of the inner loop won't cause the outer loop to stop.

You might consider changing `loop` to `while !self.opts.files_with_matches || self.match_count == 0` or something similar. (With the case unfortunately repeated inside the `while` loop as well.)

andschwa: Added a small `terminate()` function; it's working just grand.

src/search_stream.rs
andschwa: Added `opts.skip_matches()`, definitely feels better.

andschwa:
Just FYI I'm already using this; I wanted to rename a function I'd plumbed through:

```sh
./target/debug/rg -l just_file | xargs gsed -i 's/just_file/skip_matches/g'
```

🎉

andschwa:
So either `--invert-match` or there's a bug: when I combine it with `--count` (i.e. without these changes), I still get files that have matches listed (is this expected?). So I've removed `--files-without-match` as I was using the same logic and so had the same bug.

andschwa:
@BurntSushi okay this is ready for review!

andschwa:
Hm my test failures are due to a few runners swapping the order in which the output was written.

BurntSushi:
Run with -j1.

andschwa:
Sadly, while `-j1` does make it deterministic, it does not account for ordering being different across platforms, which caused most of the failures in [build #168](https://travis-ci.org/BurntSushi/ripgrep/builds/162517709) (although a couple failed with very odd panics on tests these changes didn't touch).

Since I don't actually need to test with multiple files... I'm going the simpler route and just testing with `sherlock`. Slightly less robust; much less prone to pointless failure.

BurntSushi:
Sounds good to me!

andschwa:
So, this is admittedly strange, but Travis CI appears to be caching the built targets (and so left around my created test file `file` from [build #168](https://travis-ci.org/BurntSushi/ripgrep/builds/162517709) on the next run [build #169](https://travis-ci.org/BurntSushi/ripgrep/builds/162518787), which didn't expect it), causing failures. This appears to be what's happening though because when I added a `cargo clean` to the CI scripts, they're all now passing [build #170](https://travis-ci.org/BurntSushi/ripgrep/builds/162519607). I can squash that into this commit if you agree.

BurntSushi:
That's fine sure, although I haven't seen that happen before!

BurntSushi:
I just reviewed it and this is high quality work. Thank you so much. :-)

Your code refactoring is also going to help fix #77. Yay!

andschwa:
Thank you! I really wanted this feature, and missed programming in Rust a lot. Cheers!

PR #69, opened 2016-09-24
Add support for the Nim programming language file type

dloss:


BurntSushi:
Ah, surprised I forgot about Nim! Awesome, thank you.

PR #71, opened 2016-09-25
[Fixes #46] Use 1 less worker thread than number of threads

### LINKED ISSUE #46: switch default thread count to `cpus - 1` instead of `cpus`
It makes sense to not use one thread more than the number of CPUs you have. Namely, `-j` launches `N` workers for searching, while the main thread does directory traversal.

Ideally, `cpus` would be the number of physical cores (i.e., not counting hyper threading).

catchmrbharath:
The main thread does directory traversal. Hence
number of threads = main Thread + number of worker threads.
We should have atleast one worker thread.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -28,6 +28,7 @@ use std::process;
 use std::result;
 use std::sync::{Arc, Mutex};
 use std::thread;
+use std::cmp;
 
 use deque::{Stealer, Stolen};
 use grep::Grep;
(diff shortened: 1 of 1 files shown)


BurntSushi:
Could you add "Fixes #46" to the commit message so Github links them? Otherwise this looks good, thanks!

BurntSushi:
Thanks!

PR #78, opened 2016-09-25
Update AUR package name

SevereOverfl0w:
Seems that the package was renamed.


Arvamer:
It was [moved to community](https://www.archlinux.org/packages/community/x86_64/ripgrep/).

BurntSushi:
Indeed, `ripgrep` is now in community. I'll update the README in a bit (there are some other things I'd like to say, particularly around SIMD support and such).

Thanks though!

PR #100, opened 2016-09-26
Recognise cljc and cljx extensions as Clojure(script)

emlyn:
(see https://github.com/clojure/clojurescript/wiki/Using-cljc)


BurntSushi:
Is Clojurescript a distinct language from Clojure?

(Apologies for my ignorance.)

emlyn:
I think that's debatable, some might argue that they are distinct, but for this purpose they can probably be considered the same.

BurntSushi:
Good enough for me!

BurntSushi:
Thanks!

PR #111, opened 2016-09-27
Max depth option

### LINKED ISSUE #109: Directory ignore command line
Do you think it would be valuable to add an option to skip traversing directories? I know I can use an `.rgignore` file but it's not convenient to copy it from repository to repository.

If so, I think I could try and implement it as well.

gsquire:
Closes #109

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -136,6 +136,10 @@ Less common options:
     -L, --follow
         Follow symlinks.
 
+    --maxdepth NUM
+        Descend at most NUM directories below the command line arguments.
+        A value of zero only searches the starting-points themselves.
+
     --mmap
         Search using memory maps when possible. This is enabled by default
         when ripgrep thinks it will be faster. (Note that mmap searching
@@ -222,6 +226,7 @@ pub struct RawArgs {
     flag_invert_match: bool,
     flag_line_number: bool,
     flag_fixed_strings: bool,
+    flag_maxdepth: Option<usize>,
     flag_mmap: bool,
     flag_no_heading: bool,
     flag_no_ignore: bool,
@@ -272,6 +277,7 @@ pub struct Args {
     invert_match: bool,
     line_number: bool,
     line_per_match: bool,
+    maxdepth: Option<usize>,
     mmap: bool,
     no_ignore: bool,
     no_ignore_parent: bool,
@@ -399,6 +405,7 @@ impl RawArgs {
             invert_match: self.flag_invert_match,
             line_number: !self.flag_no_line_number && self.flag_line_number,
             line_per_match: self.flag_vimgrep,
+            maxdepth: self.flag_maxdepth,
             mmap: mmap,
             no_ignore: no_ignore,
             no_ignore_parent:
@@ -681,7 +688,10 @@ impl Args {
 
     /// Create a new recursive directory iterator at the path given.
     pub fn walker(&self, path: &Path) -> Result<walk::Iter> {
-        let wd = WalkDir::new(path).follow_links(self.follow);
+        let mut wd = WalkDir::new(path).follow_links(self.follow);
+        if let Some(maxdepth) = self.maxdepth {
+            wd = wd.max_depth(maxdepth);
+        }
         let mut ig = Ignore::new();
         // Only register ignore rules if this is a directory. If it's a file,
         // then it was explicitly given by the end user, so we always search
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -831,6 +831,21 @@ sherlock\x00can extract a clew from a wisp of straw or a flake of cigar ash;
     assert_eq!(lines, expected);
 });
 
+// See: https://github.com/BurntSushi/ripgrep/issues/109
+clean!(max_depth, "far", ".", |wd: WorkDir, mut cmd: Command| {
+    wd.create_dir("one");
+    wd.create("one/pass", "far");
+    wd.create_dir("one/too");
+    wd.create("one/too/many", "far");
+
+    cmd.arg("--maxdepth").arg("2");
+
+    let lines: String = wd.stdout(&mut cmd);
+    let expected = path("one/pass:far\n");
+
+    assert_eq!(lines, expected);
+});
+
 #[test]
 fn binary_nosearch() {
     let wd = WorkDir::new("binary_nosearch");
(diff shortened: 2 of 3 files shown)


gsquire:
I'm not sure of how to write a unit test because I think I would just be testing `walkdir` if I did :)

Let me know if you are more clever than me in this regard.

src/args.rs
```diff
@@ -136,6 +136,9 @@ Less common options:
     -L, --follow
         Follow symlinks.
 
+    --maxdepth NUM
+        Descend at most NUM directories below the command line arguments.
+
```
BurntSushi: Could you add this flag to `doc/rg.1.md` and run the convert script to create a new `doc/rg.1` file?

BurntSushi: Maybe clarify that a value of `0` searches the current directory only.

src/args.rs
```diff
@@ -681,7 +687,12 @@ impl Args {
 
     /// Create a new recursive directory iterator at the path given.
     pub fn walker(&self, path: &Path) -> Result<walk::Iter> {
-        let wd = WalkDir::new(path).follow_links(self.follow);
+        let wd =
+            if self.maxdepth > 0 {
+                WalkDir::new(path).follow_links(self.follow).max_depth(self.maxdepth)
+            } else {
+                WalkDir::new(path).follow_links(self.follow)
+            };
```
BurntSushi: I think I like this better, stylistically:

```
let mut wd = WalkDir::new(path).follow_links(self.follow);
if let Some(maxdepth) = self.maxdepth {
    wd = wd.max_depth(maxdepth);
}
```

src/args.rs
```diff
@@ -681,7 +687,12 @@ impl Args {
 
     /// Create a new recursive directory iterator at the path given.
     pub fn walker(&self, path: &Path) -> Result<walk::Iter> {
-        let wd = WalkDir::new(path).follow_links(self.follow);
+        let wd =
+            if self.maxdepth > 0 {
```
BurntSushi: I think we want to allow a value of `0` to mean "search current directory." In this case, `maxdepth` should be an `Option<usize>`, which docopt will handle correctly.

BurntSushi:
@gsquire Could you write an integration test please? It would be good to follow the pattern I've been using whenever I introduce new command line flags: https://github.com/BurntSushi/ripgrep/blob/master/tests/tests.rs#L781-L820

Other than that and a few nits, this looks good, thanks!

gsquire:
I rebased onto master and squashed. Is that the style of integration test you expected?

tests/tests.rs
```diff
@@ -831,6 +831,17 @@ sherlock\x00can extract a clew from a wisp of straw or a flake of cigar ash;
...
 
+// See: https://github.com/BurntSushi/ripgrep/issues/109
+clean!(max_depth, "far", ".", |wd: WorkDir, mut cmd: Command| {
+    wd.create_dir("one");
+    wd.create_dir("one/too");
+    wd.create("one/too/many", "far");
+
+    cmd.arg("--maxdepth").arg("1");
+
+    wd.assert_err(&mut cmd);
```
BurntSushi: Could you add a file `one/foo` with `far` in it, and assert that you get results? That seems like a slightly better test since it makes sure search still works, but that it also respects the depth setting.

BurntSushi:
@gsquire That's perfect! I had one last suggestion and then this is ready to go. Thanks so much!

gsquire:
@BurntSushi Doing this causes the test to fail:

```rust
wd.create_dir("one");
wd.create("one/foo", "far");
wd.create_dir("one/too");
wd.create("one/too/many", "far");

cmd.arg("--maxdepth").arg("1");

let lines: String = wd.stdout(&mut cmd);
let expected = path("one/pass:far\n");

assert_eq!(lines, expected);
```

But changing the depth to 2 passes...am I misunderstanding the depth used in `walkdir`?

BurntSushi:
I goofed. Looks like your initial implementation is correct. I was off by
one. Could you just make sure we are consistent with GNU find? The docs
will need to be updated too. Sorry for short response. On mobile.

gsquire:
It's okay, just wanted to verify.

Quoting from the [GNU find man page](http://man7.org/linux/man-pages/man1/find.1.html):

> Descend at most levels (a non-negative integer) levels of directories below the starting-points. 
-maxdepth 0 means only apply the tests and actions to the starting-points themselves.

So no need to update anything? I have to say, I thought having max depth of one and having a file in a directory below `.` would be found. This is why I asked you to confirm. No rush on this, I want to make sure I understand it correctly is all.

tests/tests.rs
BurntSushi: OK, so I think with `one/foo`, you do indeed need to set `maxdepth` to `2`. The argument, `.`, corresponds to depth `0`. `one` is depth `1` and `one/foo` is depth `2`.

BurntSushi:
@gsquire Ah OK. Looks good. I think just updating the test to make sure it both ignores a deeper directory and gets search results from a shallower directory should be sufficient. Also, I think the docs need to be updated. Probably saying "current directory" is wrong and we should say "starting points" like `find` does. Thanks for sticking with me!

tests/tests.rs
gsquire: Interesting, I thought the count applied to directories and not their contents. I'll update it :)

BurntSushi: The count applies to whether to descend into a directory or not. At depth `0`, you don't descend ever, you just yield whatever was given. So a walkdir of `./` at depth `0` yields `./`. At depth `1`, you descend once, so that a walkdir of `./` at depth `1` yields `one`, but doesn't descend into `one`. Finally, at depth `2`, a walkdir of `./` yields `./one` and then `./one/too` and also `./one/foo`, but doesn't descend into `./one/too`.

gsquire:
Sweet, I know the phrasing is important. I updated that and the test. Hope it's good to go!

tests/tests.rs
gsquire: That makes sense, thanks for a detailed explanation. I suppose it's really a "tree" after all.

gsquire:
Looks like the build failure happened on master too...strange.

BurntSushi:
Yup this looks great. The failure looks unrelated. Thanks so much for doing this and putting up with me! :-)

gsquire:
I'm happy to be able to contribute, you're a Rust machine :)

Congrats on this project!

PR #115, opened 2016-09-27
Update brew 0.2.1 package hashes

nickstenning:
https://github.com/BurntSushi/ripgrep/commit/9fa38c62327b5e72111bc68b4c299fbf970d3f7c updated the package version in the Homebrew formula but didn't update the SHA256 hashes.


BurntSushi:
That's... weird. I could have sworn I did it. Really need to automate that.

Anyway, thanks and sorry about the goof!

PR #121, opened 2016-09-28
if --color always, always print with color, even when --vimgrep is given

lilydjwg:


src/args.rs
```diff
@@ -355,7 +355,9 @@ impl RawArgs {
                 self.flag_threads
             };
         let color =
-            if self.flag_vimgrep {
+            if self.flag_color == "always" {
+              true
+            } else if self.flag_vimgrep {
                 false
             } else if self.flag_color == "auto" {
                 atty::on_stdout() || self.flag_pretty
```
BurntSushi: I think you can change the `else` in this expression to be `false` now, right?

lilydjwg: Yes, I've changed that.

BurntSushi:
Thanks!

PR #122, opened 2016-09-28
colorize filepath at the beginning of line too

lilydjwg:


BurntSushi:
Looks good to me, thanks so much!

PR #125, opened 2016-09-28
[WIP] Implement -o, --only-matching.

### LINKED ISSUE #34: Add an option similar to -o, --only-matching
With grep you can print only the matched parts of the files. The option is described like this in the grep manpage:

```
-o, --only-matching
       Print only the matched (non-empty) parts of a matching line, with each such part on a separate output line.
```

A more powerful option, would something similar to `--replace`, but that doesn't print the non matched part of the text.

utkarshkukreti:
Fixes #34.

@BurntSushi The tests pass, but let me know if anything is wrong or can be improved!

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -165,6 +165,10 @@ Less common options:
         a list of matching files such as with --count, --files-with-matches
         and --files.
 
+    -o, --only-matching
+        Only print the text that matches the regex, with each match on a
+        separate line.
+
     -p, --pretty
         Alias for --color=always --heading -n.
 
@@ -236,6 +240,7 @@ pub struct RawArgs {
     flag_no_mmap: bool,
     flag_no_filename: bool,
     flag_null: bool,
+    flag_only_matching: bool,
     flag_pretty: bool,
     flag_quiet: bool,
     flag_regexp: Vec<String>,
@@ -283,6 +288,7 @@ pub struct Args {
     no_ignore_parent: bool,
     no_ignore_vcs: bool,
     null: bool,
+    only_matching: bool,
     quiet: bool,
     replace: Option<Vec<u8>>,
     text: bool,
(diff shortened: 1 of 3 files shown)


BurntSushi:
@utkarshkukreti Thanks! This looks good as a first pass. One of the reasons why I hadn't just done this myself is because I wasn't sure about its interaction with other flags. But we can figure that out here! Some questions:

* What if the user wants line counts? What about column numbers?
* What if `--vimgrep` is passed? (I think having one flag override the other is fine btw, we just need to say it.)
* What about colors?
* How does it interact with the context options? (GNU grep's output here is... strange, but I can see how it makes sense. Namely, it doesn't print any context lines, but does still print context separators.)

Pretty much all of these interactions need to be tested too. Doing some of these things may not be very nice in the current printer code.

utkarshkukreti:
@BurntSushi thanks for the initial review! I'll try to address everything as soon as I can. As for (3), both `ag` and `grep` color all the text with `-o`, so I think we should do the same here if colors are enabled.

BurntSushi:
@utkarshkukreti That's a reasonable choice.

In general, we should probably put a strong prefer towards answering my questions by looking at what other tools do. I've tried to give `grep` behavior the higher priority over any other tool. (With that said, there can be good reasons for doing something different than `grep`, we should just try to justify it.)

utkarshkukreti:
Unfortunately, I didn't get time to work on this and probably won't for a few weeks. I'm closing this but if anyone wants to use these changes as a starting point for their own attempt, please feel free to!

BurntSushi:
@utkarshkukreti No problem! Thanks for the heads up! I understand this was probably trickier than it appeared, in large part because the printing code isn't in the best shape. I expect to try to add this when I refactor the search/print code.

PR #133, opened 2016-09-29
AppVeyor: Change release description to fit Travis binaries

akien-mga:
That will avoid all your future automatically deployed releases being described as "Windows release" when they contain binaries for Linux, OSX and Windows.


BurntSushi:
Oh, that's cute. I never even saw "Windows release" on the releases page! Thanks!

Issue #10, opened 2016-09-21
add to package repositories


BurntSushi:
It'd be nice to at least get it into Ubuntu and homebrew. Sadly, I think either one of those will be quite difficult since Rust isn't packaged in either one of them.

The lowest hanging fruit is the Archlinux User Repository (AUR).

Fedora is getting Rust packaged soon, so that may be plausible.

Sadly, we may need to live with binaries distributed here for the time being.

BurntSushi:
I now have a [brew formula](https://github.com/BurntSushi/ripgrep/blob/master/pkg/brew/ripgrep.rb) and an [Archlinux `PKGBUILD`](https://github.com/BurntSushi/ripgrep/blob/master/pkg/archlinux/PKGBUILD). I also added [ripgrep to the Archlinux user repository](https://aur.archlinux.org/packages/ripgrep/).

Sadly, the brew formula isn't actually in Homebrew, and it downloads a binary, which is really bad form. I'd like to fix it, but I'm not quite sure how.

tekacs:
rust does appear to be packaged in Homebrew (http://braumeister.org/formula/rust) - just a heads-up.

BurntSushi:
@tekacs Oh, cool, somehow I missed that! I'll check that out, because I definitely want to get `ripgrep` into `brew` proper if that's possible.

mwean:
I tried to create a formula for it:

```ruby
class Ripgrep < Formula
  desc "ripgrep combines the usability of The Silver Searcher with the raw speed of grep"
  homepage "https://github.com/BurntSushi/ripgrep"
  url "https://github.com/BurntSushi/ripgrep/archive/#{version}.tar.gz"
  version "0.1.17"
  sha256 "b558b6650bfa9cf0fd0fa58a8617cafc7c819ee25a26d15ca2ce39979dd18a18"

  depends_on "rust" => :build

  def install
    system "cargo", "build"
  end

  test do
    system bin/"rg", "--help"
  end
end
```

But couldn't get it to build. The first time I tried, it was stuck on `make` for over an hour before I killed it. Then I got this error:

```
Last 15 lines from /Users/Matt/Library/Logs/Homebrew/rust/02.make:
[ 22%] Building CXX object lib/Support/CMakeFiles/LLVMSupport.dir/SearchForAddressOfSpecialSymbol.cpp.o
[ 23%] Building CXX object lib/Support/CMakeFiles/LLVMSupport.dir/Signals.cpp.o
[...]

PR #139, opened 2016-09-30
Make the repo a Homebrew Tap

moshen:
Because there have been several comments on the builds provided by @BurntSushi [having `simd-accel` enabled](https://github.com/BurntSushi/ripgrep/issues/10#issuecomment-250612979), vs [HomeBrew using rust stable to build](https://github.com/Homebrew/homebrew-core/pull/5268).  I thought I would create this PR to enable the BurntSushi/ripgrep repo to be used as a Homebrew tap.  This would allow those who are tracking these releases (built with rust nightly?) to have easy updates, at least until the official Homebrew build can enable `simd-accel`.

Another aside, this PR removes the non-64-bit releases from the `Formula`.  This is because everything Mac OS 10.7 and newer has to be running on an x86_64 cpu.  If you want to continue supporting non-64-bit installs, I can add it back to this PR.

### DIFF
--- a/HomebrewFormula
+++ b/HomebrewFormula
@@ -1,0 +1,1 @@
+pkg/brew
--- a/README.md
+++ b/README.md
@@ -97,12 +97,11 @@ but you'll need to have the
 [Microsoft VC++ 2015 redistributable](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
 installed.
 
-If you're a **Homebrew** user, then you can install it with a custom formula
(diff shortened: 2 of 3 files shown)


BurntSushi:
This is awesome, thank you.

> Another aside, this PR removes the non-64-bit releases from the Formula. This is because everything Mac OS 10.7 and newer has to be running on an x86_64 cpu. 

I'm not a Mac user so I hadn't realized this. I'll also remove the binary from the Github releases too. (An easy decision, given that it has been causing problems lately.)

Issue #134, opened 2016-09-29
processing a large ignore file is very slow


Deewiant:
Testcase: pristine git clone of [mysql/mysql-server](https://github.com/mysql/mysql-server) at the current HEAD commit, 71f48ab393bce80a59e5a2e498cd1f46f6b43f9a. Quick comparison of `git grep` and `rg` on a fixed-string pattern that is not found anywhere in the repository, after warming up the filesystem cache with `git grep` a few times:

```
$ time git grep thiswillnotbefound
0.29suser 0.33skernel 0.135elapsed 459%CPU 26Mmaxmem
0inputs+0outputs (0major+8952minor)pagefaults (39406voluntary+52in)switches

$ time rg thiswillnotbefound
3845.59suser 0.78skernel 8:01.86elapsed 798%CPU 130Mmaxmem
0inputs+0outputs (0major+5543minor)pagefaults (51voluntary+133031in)switches
```

Somehow ripgrep is more than three orders of magnitude slower than `git grep` here, at 8 minutes versus 0.1 seconds. This is of course not a very good sample of results but I think with a difference that big you don't really need the details. (Feel free to ask if you can't reproduce it though.)

This is with ripgrep 0.2.1, git 2.10.0.

One-shot results from The Silver Searcher 0.33.0 and sift 0.8.0 for further comparison:

```
$ time ag thiswillnotbefound
1.64suser 0.75skernel 1.324elapsed 180%CPU 4Mmaxmem
0inputs+0outputs (0major+29085minor)pagefaults (66106voluntary+108in)switches

$ time sift --git thiswillnotbefound
Error: 4 files skipped due to very long lines (>= 262144 bytes). See options --blocksize, --err-show-line-length and --err-skip-line-length.
24.31suser 0.54skernel 7.715elapsed 322%CPU 26Mmaxmem
0inputs+0outputs (0major+4039minor)pagefaults (77269voluntary+348in)switches
```

oconnor663:
I can repro this. It looks like the problem is something to do with `.gitignore` handling. If we disable that, `rg` is actually faster:

```
$ cd mysql-server
$ time git grep thiswillnotbefound
noglob git grep thiswillnotbefound  0.12s user 0.15s system 257% cpu 0.107 total
$ time rg thiswillnotbefound --no-ignore
rg thiswillnotbefound --no-ignore  0.16s user 0.10s system 292% cpu 0.090 total
$ time rg thiswillnotbefound
[...]

PR #141, opened 2016-10-01
avoid using an Arc for quiet_matched

lilydjwg:
I'm sorry, the result isn't stable, and I can't tell whether it's faster or slower....

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -102,10 +102,11 @@ fn run(args: Args) -> Result<u64> {
     }
 
     let out = Arc::new(Mutex::new(args.out()));
-    let quiet_matched = QuietMatched::new(args.quiet());
+    let quiet_matched = QuietMatched::new();
     let mut workers = vec![];
 
     let workq = {
+        let quiet = args.quiet();
         let (workq, stealer) = deque::new();
         for _ in 0..threads {
             let worker = MultiWorker {
@@ -120,7 +121,7 @@ fn run(args: Args) -> Result<u64> {
                     match_count: 0,
                 },
             };
-            workers.push(thread::spawn(move || worker.run()));
+            workers.push(thread::spawn(move || worker.run(quiet)));
         }
         workq
     };
@@ -271,9 +272,9 @@ struct Worker {
 }
 
 impl MultiWorker {
-    fn run(mut self) -> u64 {
+    fn run(mut self, quiet: bool) -> u64 {
         loop {
-            if self.quiet_matched.has_match() {
+            if quiet && self.quiet_matched.has_match() {
                 break;
             }
             let work = match self.chan_work.steal() {
@@ -294,7 +295,7 @@ impl MultiWorker {
             outbuf.clear();
             let mut printer = self.worker.args.printer(outbuf);
             self.worker.do_work(&mut printer, work);
-            if self.quiet_matched.set_match(self.worker.match_count > 0) {
+            if quiet && self.quiet_matched.set_match(self.worker.match_count > 0) {
                 break;
             }
             let outbuf = printer.into_inner();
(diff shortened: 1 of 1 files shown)


BurntSushi:
Yeah, I wouldn't expect this to make a difference. An Arc should only impose overhead when it's cloned, not when it's accessed. We only clone it once for each worker, which means the overhead is likely imperceptible.

Can you explain more about your benchmark?

lilydjwg:
I just use perf to see where it takes the most time.

```
sudo perf record -F 99 -g -- rg xxx >/dev/null
```

then
```
sudo perf report
```

and press <kbd>a</kbd> on the function that uses the most of time to see assembly. Not very easy to read, but there is a `mfence` instruction, which clearly corresponds to the `fence(SeqCst)` call inside the `steal()` function.

a

Today rg takes more than one minute to complete a search, so I tried to find out why. The time reduces to several seconds when I build the new master.

BurntSushi:
Profiles can be hard to read, and in this case, it is probably misleading you. The profile reveals a lot of time spent in the work queue because it is *spinning* waiting for work. That is, the directory iterator can't produce work fast enough for the search workers.

The time reduction on master almost certainly means that the directory you're searching has a very large gitignore file. I implemented a crude stopgap measure to make this faster, but I'm still working on it.

If I'm right, then you should notice a large improvement by running with the `-u` flag.

One way to extract a less confusing profile is to pass `-j1`. This will cause single threaded search to, which doesn't use a work queue at all.

lilydjwg:
So there is a spinning lock? It makes my fan spin loudly....I prefer it to block instead.

There is no gitignore file, though it's in a (huge) git repo---the "community" repo from Arch Linux. `-u` indeed makes it much faster.

I take the source code outside the repo, and it's slow the same.

With `-j1` rg uses only one core and it takes less time! Though it's still  much slower than `ag --workers 1`.

BurntSushi:
Could you please tell me how you're running your test? What command did you run to get the data? What `rg` commands are you running?

> So there is a spinning lock? It makes my fan spin loudly....I prefer it to block instead.

This is how lock free queues work. We should fix the producer.

lilydjwg:
I run this in the `percona-server-5.7.14-8` directory
```
time ./rg utf8mb4 >/dev/null
```

> This is how lock free queues work. We should fix the producer.

Does this mean that, if I search on slow disks, it takes a lot of CPU time to wait?

BurntSushi:
Could you please tell me how to get the data you're searching? What command should I run?

> Does this mean that, if I search on slow disks, it takes a lot of CPU time to wait?

No, it doesn't mean that. If the disks are slow and the corpora isn't memory, then the search workers will also be slower.

lilydjwg:
Extract [this tarball](http://www.percona.com/downloads/Percona-Server-5.7/Percona-Server-5.7.14-8/source/tarball/percona-server-5.7.14-8.tar.gz).

> No, it doesn't mean that. If the disks are slow and the corpora isn't memory, then the search workers will also be slower.
I see, thanks for the explanation :-)

BurntSushi:
> There is no gitignore file, though it's in a (huge) git repo---the "community" repo from Arch Linux. -u indeed makes it much faster.

Could you show me how to get this data as well? And the `rg` command you're running?

(I'm asking because it looks like there's more going on here. The only difference between `rg PAT` and `rg -u PAT` is that the latter doesn't look at ignore files.)

BurntSushi:
For `percona-server`, it seems clear to me that you're running into #134. There is a *giant* `.gitignore` file in the root directory of that tarball. This is where `rg` is spending all of its time. Since directory traversal is single threaded, it can't keep up with the search workers, so the search workers are constantly asking for more work, causing the queue to spin and use up CPU.

lilydjwg:
Oh, I see! I used `ag -g` to search for `gitignore`, but it didn't search for hidden files either!

BurntSushi:
Indeed. On `percona-server`, I get these timings:

```
[andrew@Cheetah percona-server-5.7.14-8] time rg utf8mb4 | wc -l
3114

real    0m0.813s
user    0m6.163s
sys     0m0.183s
[andrew@Cheetah percona-server-5.7.14-8] time ag utf8mb4 | wc -l
3210

real    0m1.687s
user    0m2.263s
sys     0m0.930s
```

(The counts are off because I'm pretty sure `ag` is getting the `.gitignore` file wrong in some places.)

If we ask both tools to skip ignore files, then they both get a lot faster, but `rg` still wins:

```
[andrew@Cheetah percona-server-5.7.14-8] time rg -u utf8mb4 | wc -l
3117

real    0m0.118s
user    0m0.323s
sys     0m0.350s
[andrew@Cheetah percona-server-5.7.14-8] time ag -u utf8mb4 | wc -l
3214

real    0m0.399s
user    0m0.617s
sys     0m1.053s
```

BurntSushi:
@lilydjwg FYI, I made a few (important) edits to my previous comment.

PR #143, opened 2016-10-03
Fix brew formula name to not conflict with core

moshen:
Since the homebrew-core formula was accepted, we should differentiate
the prebuilt formula available in this tap

Only after my previous PR was accepted (and I tried it again), did I notice that there may be an issue with conflicting formula names:

image

This is failing because the cache I have from installing the `homebrew-core` version of ripgrep doesn't match the sha of the prebuilt download.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,7 +101,7 @@ If you're a **Homebrew** user, then you can install it with a custom tap:
 
 ```
 $ brew tap burntsushi/ripgrep https://github.com/BurntSushi/ripgrep.git
-$ brew install burntsushi/ripgrep/ripgrep
+$ brew install burntsushi/ripgrep/ripgrep-bin
 ```
 
 If you're an **Arch Linux** user, then you can install `ripgrep` from the official repos:
--- a/pkg/brew/ripgrep-bin.rb
+++ b/pkg/brew/ripgrep-bin.rb
@@ -1,9 +1,11 @@
-class Ripgrep < Formula
+class RipgrepBin < Formula
   version '0.2.1'
   desc "Search tool like grep and The Silver Searcher."
   homepage "https://github.com/BurntSushi/ripgrep"
   url "https://github.com/BurntSushi/ripgrep/releases/download/#{version}/ripgrep-#{version}-x86_64-apple-darwin.tar.gz"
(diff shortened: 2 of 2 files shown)


BurntSushi:
Is `-prebuilt` an idiom that the Homebrew ecosystem uses? If not, is there one?

moshen:
Good question.  There is really only `bottle` from what I can tell.  This actually seems like more of a problem with Homebrew since it doesn't namespace the cache by tap name.

BurntSushi:
Hmm. I would like to make sure we get this name right. For example, I can imagine being convinced that the tap should compile from source instead of distributing prebuilt binaries. However, that's unlikely to happen unless something like `rustup` gets packaged because we'd need a nightly compiled to enable SIMD.

In Archlinux, we'd call this `ripgrep-git`, but that's not accurate because this tap doesn't track master. We also have things like `ripgrep-bin`, but that has the same problem as `ripgrep-prebuilt`.

Maybe I'm overthinking this. (I do like `ripgrep-bin` better than `ripgrep-prebuilt` though.)

moshen:
It appears that [multirust](https://github.com/Homebrew/homebrew-core/blob/d4109a06e6b3957a7091ced6de1dd9d4de801952/Formula/multirust.rb) is available (which uses rustup?).  However, there are no formula which depend on it.. so I'm not sure if that's allowed.  [I have a change that is using it to build with rust nightly](https://github.com/Homebrew/homebrew-core/pull/5563).  I guess this PR could just hang out depending on what happens there.  If the Homebrew PR is accepted and bottles created from the nightly build, I can revise this PR further.

BurntSushi:
`multirust` is on the way out unfortunately (to be replaced by `rustup`). The Homebrew folks probably made the right call by sticking with `rust`. In general, we really shouldn't be packaging nightly Rust at all. It just so happens that `ripgrep` benefits from it. I realize it sucks, but it's a good motivator for us (the Rust community) to get SIMD stabilized.

With that in mind, I'm just going to accept this and call it done. Thanks so much for straightening this out!

moshen:
Gotcha.  No problem!  Thanks for ripgrep! 🍻

PR #144, opened 2016-10-04
Added *.cmake extension to cmake file type.

bitshifter:
.cmake is a common extension used by CMake modules and includes.

For example, the include macro searches for .cmake files - https://cmake.org/cmake/help/v3.5/command/include.html?highlight=include.


BurntSushi:
Thanks!

Issue #145, opened 2016-10-04
I'm getting "invalid argument" with trying to add new types


samuelcolvin:
Best described with terminal output

```shell
> rg --type-add 'jinja:*.{jinja,jinja2}'
Invalid arguments.

Usage: rg [options] -e PATTERN ... [<path> ...]
       rg [options] <pattern> [<path> ...]
       rg [options] --files [<path> ...]
       rg [options] --type-list
       rg [options] --help
       rg [options] --version
> echo $?
1
> rg --version
0.2.1
```

However `--type-add` works fine with combined with a search, eg. `rg --type-add 'jinja:*.{jinja,jinja2}'  -tjinja foobar`, but that doesn't add the type to the global list and has to be used on every search.

OS: ubuntu 16.04, kernel `4.4.0-36-generic`.

BurntSushi:
> but that doesn't add the type to the global list and has to be used on every search.

There is no global list or persistent configuration that `ripgrep` writes. I would very much appreciate it if you could tell me where the docs have led you astray. :-( How can I fix them?

samuelcolvin:
Docs are clear now I look more properly, just me being dumb.

Would you accept a pull request to add jinja files to the types?

BurntSushi:
@samuelcolvin Absolutely. Thanks. :-)

PR #146, opened 2016-10-04
add jinja type for *.jinja and *.jinja2

samuelcolvin:
see #145.

I guess this is quite niche, it's just one templating language (although popular) but without the option for an override or extension locally I guess this is the best option.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -42,6 +42,7 @@ const TYPE_EXTENSIONS: &'static [(&'static str, &'static [&'static str])] = &[
     ("haskell", &["*.hs", "*.lhs"]),
     ("html", &["*.htm", "*.html"]),
     ("java", &["*.java"]),
+    ("jinja", &["*.jinja", "*.jinja2"]),
     ("js", &[
         "*.js", "*.jsx", "*.vue",
     ]),


src/types.rs
```diff
@@ -82,6 +82,7 @@ const TYPE_EXTENSIONS: &'static [(&'static str, &'static [&'static str])] = &[
     ("xml", &["*.xml"]),
     ("yacc", &["*.y"]),
     ("yaml", &["*.yaml", "*.yml"]),
+    ("jinja", &["*.jinja", "*.jinja2"]),
```
BurntSushi: You you put this in alphabetical order? Trying to keep things tidy. :-)

BurntSushi:
> but without the option for an override or extension locally I guess this is the best option.

Not sure I follow. `--type-add` will let you do this. e.g., `rg --type-add 'jinga:*.{jinga,jinga2}' -tjinga QUERY_PATTERN` will only search Jinga files. If you do `alias rg="rg --type-add 'jinga:*.{jinga,jinga2}'"` then `rg -tjinga QUERY_PATTERN` will do the right thing.

(I'd still like to accept this PR too. Jinga is pretty popular.)

samuelcolvin:
(Note, it's jinja, not "jinga".)

> `alias rg="rg --type-add 'jinga:*.{jinga,jinga2}'"`

Makes sense, and I guess that's leaner and more canonical than `rg` having it's own config file.

I had originally assumed that since ripgrep broadly followed git's example wrt to ignore files it might do the same wrt config settings where users are likely to have very specific and esoteric preferences, but bash aliases should suffice.

BurntSushi:
@samuelcolvin Gotya. And yes, `ripgrep` may wind up with a config file or something similar, but I want to see how far we can get with just aliases first. :-)

In any case, thank you!

PR #92, opened 2016-09-25
ripgrep is now in [community]

svenstaro:
The README should reflect that.

PR #148, opened 2016-10-05
Change Arch Linux instructions

munyari:
The `-Syu` flag will do a full system upgrade and then install the package, which is not necessarily the desired behavior. Only the `-S` flag is necessary to install a single package.
See https://wiki.archlinux.org/index.php/Pacman#Installing_specific_packages
https://wiki.archlinux.org/index.php/Pacman#Upgrading_packages

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ $ brew install burntsushi/ripgrep/ripgrep-bin
 If you're an **Arch Linux** user, then you can install `ripgrep` from the official repos:
 
 ```
-$ pacman -Syu ripgrep
+$ pacman -S ripgrep
 ```
 
 If you're a **Rust programmer**, `ripgrep` can be installed with `cargo`:


BurntSushi:
The `-Syu` was added by @svenstaro in #92. I also found it kind of odd. @svenstaro Thoughts?

svenstaro:
It's true that you only need -S but then again telling people to do -Syu is the safer bet because we do not support partial upgrades and we regularly get problems with that in IRC. You can take it out but someone will find a way to do a partial upgrade.

munyari:
Perhaps, but if that's considered best practice, shouldn't it be in the arch wiki?

svenstaro:
Perhaps so but I'm sure it says that somewhere. Anyway, I think @BurntSushi should accept this patch and then we'll just hope that no idiot turns up that manages to break something.

BurntSushi:
All right, let's do it. Thanks for working it out!

munyari:
@svenstaro no need to refer to anybody as an idiot mate. It's not the most obvious problem.

lugubris:
sorry for necrobumping, got refrenced from reddit

- partial upgrades are not supported so -Sy should never be used, to install packages -S should be used [0].

So assuming the user never runs -Sy instead of -Syu, using -S to install software does no harm.

[0] -> https://wiki.archlinux.org/index.php/Pacman#Installing_packages

PR #155, opened 2016-10-07
Adding extra .md filetype for ease of access to Markdown filetypes

theamazingfedex:
`-tmarkdown` is somewhat lengthy and unwieldy in practice, we could easily use `-tmd` in addition for ease of access.


BurntSushi:
I agree, `md` is clearly superior than `markdown` for this. That was a gaffe on my part. Thanks!

PR #157, opened 2016-10-09
Always follow symlinks on explicit file arguments

### LINKED ISSUE #137: Unexpected interaction with symlinks
I noticed that `rg` ignores symlinks by default, even when I invoke it solely to search a single symlinked file. For me, that's very surprising behavior. The `-debug` flag also said nothing about why it was skipping my file.

So I'm suggestion to either change that behavior (unless maybe it's made this way for a good reason?), or inform the user that his symlinks are being ignored.

CannedYerins:
Fixes #137.  I have no prior knowledge, but it looks like Args.walker() is the correct place to locate this...

Let me know if you think this should be documented anyplace.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -708,7 +708,9 @@ impl Args {
 
     /// Create a new recursive directory iterator at the path given.
     pub fn walker(&self, path: &Path) -> Result<walk::Iter> {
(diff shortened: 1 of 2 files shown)


BurntSushi:
This seems OK, thank you for fixing it.

I have several style nits, but I'm just going to merge them and fix them myself.

PR #168, opened 2016-10-11
clarify docs for --threads

### LINKED ISSUE #167: Default for -j is not clear
`rg --help` says this:

```
    -j, --threads ARG
        The number of threads to use. Defaults to the number of logical CPUs
        (capped at 6). [default: 0]
```

I have four cores here. Given this documentation I have no idea what the default parallelishness is :)

durka:
Fixes #167.

### DIFF
--- a/doc/rg.1
+++ b/doc/rg.1
@@ -277,7 +277,7 @@ This is overridden by either \-\-case\-sensitive or \-\-ignore\-case.
 .TP
 .B \-j, \-\-threads \f[I]ARG\f[]
 The number of threads to use.
-Defaults to the number of logical CPUs (capped at 6).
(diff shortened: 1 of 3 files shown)


BurntSushi:
darwin builds take f..o..r..e..v..e..r..

durka:
Speaking of Darwin, `sed -i` works differently on OSX and [there's no real way to write it portably](http://stackoverflow.com/a/38595160/1114328), so `convert-to-man` doesn't work out of the box.

BurntSushi:
Screw it. Tired of waiting. Thanks!

(And yeah, I'm not particularly happy about that `convert-to-man` script for a number of reasons. Another problem for another day.)

PR #170, opened 2016-10-12
Add .gradle as a groovy extension pattern

sfackler:
Gradle is a build system for JVM-based languages that is configured via a Groovy DSL.


BurntSushi:
Thanks! :-)

PR #183, opened 2016-10-16
Add instructions for installation on Fedora 24+ and RHEL/CentOS 7.

carlwgeorge:
Please note that the referenced copr repository should just be a temporary home while the Fedora/EPEL package review is pending.

https://bugzilla.redhat.com/show_bug.cgi?id=1380442


BurntSushi:
Awesome! Thank you so much for writing that up. :D

PR #187, opened 2016-10-19
Reset the terminal when Ctrl-C is pressed

### LINKED ISSUE #119: rg doesn't reset terminal color when interrupted
I was grepping a large log file, and accidentally wrote a regex that matched a very large number of lines. After waiting for a minute of it just printing out lines to stdout, I killed `rg` with <kbd>Ctrl</kbd>-<kbd>C</kbd>. However, that left the `time` output and my shell prompt blue since apparently I'd interrupted it in the process of printing out a line number.

`rg` should probably catch SIGINT and reset the terminal color (or any other terminal changes) before exiting.

lambda:
If a user hits Ctrl-C to exit out of a search in the middle of printing
a line, we don't want to leave the terminal colors screwed up for them.
Catch Ctrl-C using the ctrlc crate, obtain a stdout lock to ensure that
other threads don't continue writing after we do so, reset the terminal,
and exit the program.

Closes #119

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,6 +2,7 @@
 name = "ripgrep"
 version = "0.2.3"
 dependencies = [
+ "ctrlc 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "deque 0.3.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "docopt 0.6.86 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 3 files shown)


lambda:
This PR is a bit of a strawman; I figured I'd try the simplest possible thing that seemed to work. I first tried to see whether I should thread this into `ColoredTerminal` somehow, but that seemed complicated, and this approach seemed simpler.

I haven't tested this at all on Windows, so don't know if this will work properly on there.

On my system, for some reason, this suppresses the shell prompt printed after exiting. I haven't quite figure out why that is. But it does have the effect of resetting the colors appropriately, so if I just hit "return", I do get the prompt properly. Writing out extra newlines and flushing doesn't seem to help this.

I had expected that this would have the effect of losing the information about the process being killed by a signal, just returning 1 instead, since I didn't clear the signal handler and re-raise the signal, but `echo $?` gives me 130 as you'd expect from an interrupted process. So I guess that's good, if unexpected.

BurntSushi:
@lambda Nice! Thanks so much for doing this. I'll pull this down and try to play with it soon before merging. (It may not happen until this weekend, sorry!)

BurntSushi:
OK, I tried this out and it does indeed fix the issue for me. Interestingly, I'm not experiencing the shell prompt suppression. I get my shell prompt back, but I have no idea what the forces are at play here. (FWIW, I'm running `bash` in `konsole` as my terminal emulator on Linux.)

I merged this in 811fcc1fe80e32916350888f4e6923cc2488901a (after a rebase).

Thanks so much!

PR #195, opened 2016-10-26
Mention Nix package in README

8573:
In the `README.md` document, where said document documents the
availability of pre-built packages of ripgrep, document the
availability of such a package from the package management system Nix.

(@rolodato from #10: you may be interested, if you haven't already
noticed the package.)

### DIFF
--- a/README.md
+++ b/README.md
@@ -132,6 +132,14 @@ $ yum-config-manager --add-repo=https://copr.fedorainfracloud.org/coprs/carlgeor
 $ yum install ripgrep
 ```
 
+If you're a **Nix** user, you can install `ripgrep` from
+[nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/text/ripgrep/default.nix):
+
+```
+$ nix-env --install ripgrep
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -132,6 +132,14 @@ $ yum-config-manager --add-repo=https://copr.fedorainfracloud.org/coprs/carlgeor
 $ yum install ripgrep
 ```
 
+If you're a **Nix** user, you can install `ripgrep` from
+[nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/tools/text/ripgrep/default.nix):
+
+```
+$ nix-env --install ripgrep
+$ # (Or using the attribute name, which is also `ripgrep`.)
```
BurntSushi: Can you explain what this line means? Is it necessary?

BurntSushi:
@8573 Thanks! :-) I have one question but this otherwise looks good!

README.md
BurntSushi: That's good enough for me! (Nix sounds really cool btw.)

Issue #196, opened 2016-10-27
Support persistent configuration (`.rgrc`?)


SimenB:
I'd love to keep different types for instance, and not have to specify `--type-add` each time.

My specific case is type zsh which searches zsh config files (`.zshrc` `*.zsh`, and some custom `.zsh*`).

Maybe it could support more stuff like `no-heading`, `--heading` etc as well, which would allow people to customize rg more.

(I know I can just `alias rgzsh="rg --type-add 'zsh:*.zsh' --type-add 'zsh:.zshrc' --type zsh"`, but meh)

Loving the tool, btw!

BurntSushi:
Can you make do with creating an alias? Or even a script in your `~/bin`? For example, I have `~/bin/rgp`:

```
#!/bin/sh

exec rg -p "$@" | less -RFX
```

SimenB:
I just updated the OP with alias stuff 😄 

And yes, I could, but then I have to remember my aliases (woe is me, right?). Doing `rg -tzsh` is more natural when I've gotten used to types than doing `rgzsh` or whatever.

BurntSushi:
Why don't you just alias `rg` itself? This works, for example:

```
alias rg="rg --type-add 'zsh:*.zsh' --type-add 'zsh:.zshrc' --type zsh"
```

Then you can do `rg -tzsh`.

(Also, `zsh` seems like a common enough type that it should be part of ripgrep proper! :-))

SimenB:
Aliasing rg seems pretty good!

I can provide a PR for zsh type (just add an entry here I suppose: https://github.com/BurntSushi/ripgrep/blob/master/src/types.rs)

But I'd still really like for that to be defined for rg, and not drowned amongst all of my other aliases (I still do `alias | rg whatever` since I forget. I'd like to think `rg --type-list` is nicer (and yes, I realize `rg --type-list` would work with your suggested `rg` alias)).

BurntSushi:
> I can provide a PR for zsh type (just add an entry here I suppose: https://github.com/BurntSushi/ripgrep/blob/master/src/types.rs)

Sounds good!

> But I'd still really like for that to be defined for rg, and not drowned amongst all of my other aliases (I still do alias | rg whatever since I forget. I'd like to think rg --type-list is nicer (and yes, I realize rg --type-list would work with your suggested rg alias)).

You can also do `command -V rg` to see the alias.

SimenB:
[...]

PR #197, opened 2016-10-27
Add zsh type

SimenB:
Like discussed in #196 

Filenames are the ones from oh-my-zsh (`*.zsh`) and the ones from here: https://www-s.acm.illinois.edu/workshops/zsh/startup_files.html


BurntSushi:
Darwin is holding things up, as usual... Passing on Linux is good enough for me for this PR. Thanks!

SimenB:
Sweet! What's your release schedule?

BurntSushi:
@SimenB I don't have one. I just do it when it feels right. Maybe some day I'll adopt a schedule, but not today.

PR #198, opened 2016-10-28
update python types to include Cython files

sanga:
Cython a tool to compile python (or possibly 'pseudo' python i.e. python with a few type annotations thrown in) to down to C. Commonly, Cython files have the extension pyx (see https://en.wikipedia.org/wiki/Cython). Personally I think it makes sense to include these files in the python type filter. Cython is widely used in the scientific python community.


BurntSushi:
SGTM! Thanks!

PR #211, opened 2016-11-01
Snapcraft definition for building ripgrep snaps.

tasdomas:
Snaps are a new application packaging approach that allows for better isolation. This PR introduces a snapcraft.yaml file that allows building ripgrep snaps using the [snapcraft](http://snapcraft.io) tool.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@ target
 /grep/Cargo.lock
 /globset/Cargo.lock
 /ignore/Cargo.lock
+parts/
+prime/
+stage/
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,16 @@
+name: rg
+version: 0.2.6
+summary: a command line search tool
+description: |
+  ripgrep is a command line search tool that
+  combines the usability of The Silver Searcher
+  (an ack clone) with the raw speed of GNU grep.
+confinement: strict
+apps:
+  rg:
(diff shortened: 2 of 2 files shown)


BurntSushi:
Hmm. I'm honestly not sure when I should merge this or not. I know that at least as of right now, I have a hard enough time maintaining the brew formula (I'm not a Mac user), so I worry that this will bitrot easily. In particular, I've never used or heard of this tool.

Is there an established way of maintaining snaps out-of-tree? I think that's what I would prefer at the moment.

anuragsoni:
@tasdomas @BurntSushi Since snapcraft (the snap package builder) can pull source from git it can live out-of-tree. I was actually trying out building a snap package today before I saw this pull request :)

I believe the only change required to the yml file would be the following.
```
parts:
  rg:
    plugin: rust
    source: git://github.com/BurntSushi/ripgrep
    source-tag: 0.2.6

```

anuragsoni:
Also, i'm not sure how common is it for snap packages to be used outside of Ubuntu for now, but I guess now they offer installation on arch, fedora and opensuse so that should cover a lot of different flavors!

tasdomas:
@BurntSushi, @anuragsoni - having the snap build out of tree is fine.

BurntSushi:
@tasdomas @anuragsoni Thanks for the info. I think I'd rather take that route for now!

PR #213, opened 2016-11-02
add asciidoc filetype and update markdown filetype

tjdgus3537:


ignore/src/types.rs
```diff
@@ -81,6 +81,7 @@ use {Error, Match};
 
 const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("agda", &["*.agda", "*.lagda"]),
+    ("asciidoc", &["*.adoc", "*.asc", "*.asciidoc"]),
```
BurntSushi: Is there a shorter abbreviation we can use for this? I feel like `ad` is too short and `ascii` might be too general however, so maybe this is the best we can do.

tjdgus3537: agree that

BurntSushi:
Thanks!

BurntSushi:
Thanks!

PR #218, opened 2016-11-03
adding cs filetype for ease of use.

theamazingfedex:


BurntSushi:
What file type is this?

theamazingfedex:
.cs for csharp

BurntSushi:
OK, so you're proposing that we have two names for the same file type. I'd like to avoid this, but we do have precedent for it, so I guess we can merge.

theamazingfedex:
I'm not sure if Rust will allow for this, but if there were a way to have multiple keys for each type, we could avoid having two distinct entries for it. 
Probably something like the following:
``` Rust
const DEFAULT_TYPES: &'static [(&'static [&'static str], &'static [&'static str])] = &[
  (&["cs", "csharp"], &["*.cs"]),
  (&["md", "markdown"], &["*.markdown", "*.md", "*.mdown", "*.mkdn"])
]
```

BurntSushi:
I'm actually fine with the duplication in source code for now. What concerns me more is the UX. That is, having multiple ways to refer to the same thing.

> If this is possible, it would also require some slight refactoring of the rg type checking module.

At least the `Type` matcher itself could be made to duplicate the definitions, even if they aren't duplicated in the source code, so I think that's fine.

PR #219, opened 2016-11-03
adding .pdf filetype to available types.

theamazingfedex:


BurntSushi:
Uh, could you explain the use case for this? PDF is a binary format. It feels weird to watch to search it. ripgrep should, in most cases, detect the binary format and skip it altogether.

theamazingfedex:
I'm not entirely sure what the exact case was, but I kept encountering situations when trying to search through my work repository that inundated me with results from a .pdf file. I figure that if a user does end up enabling binary searching, then perhaps it would be useful to include some of the most common binary filetypes.

BurntSushi:
> but I kept encountering situations when trying to search through my work repository that inundated me with results from a .pdf file

Was this happening with ripgrep?

> I figure that if a user does end up enabling binary searching, then perhaps it would be useful to include some of the most common binary filetypes.

That's a good point.

PR #232, opened 2016-11-12
Avoid heap allocations in lazy_static when using nightly

cesarb:
On nightly, the lazy_static crate is able to use `const fn` to avoid a heap allocation for each `lazy_static!`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,6 +45,7 @@ kernel32-sys = "0.2"
 winapi = "0.2"
 
 [features]
+nightly = ["lazy_static/nightly"]
 avx-accel = ["bytecount/avx-accel"]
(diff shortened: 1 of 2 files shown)


BurntSushi:
Is there a real world use case the benefits from this?

cesarb:
It should in theory make ripgrep start slightly faster (less allocations in docopt and in the two `lazy_static!` uses within ripgrep itself), though it's in the noise for me.

BurntSushi:
Yeah, I don't at all buy that this has any impact on startup time. Because of that, I don't think it's worth adding a new feature for it. It will be nice once we get `const fn` on stable Rust, but we can wait for that. Thanks anyway!

Issue #230, opened 2016-11-10
Provide bash completion


mvitz:
It would be nice to have bash completion.

BurntSushi:
I agree, but I'm unlikely to do this myself. I would however maintain one. But it would have to be automatically generated somehow.

kbknapp:
@BurntSushi Related to #136 clap now supports generating bash, zsh, fish, and powershell completion scripts :wink:

Issue #210, opened 2016-11-01
Support non-utf-8 file names


bluss:
File names on linux are just bags of bytes. Regular shell tools work with them no problem:

```
$ cat broken� 
hi
$ ls -d broken�
broken?
$ grep "hi" -n broken� 
1:hi
```

The main annoyance is that it poisons the whole directory with ripgrep:

```
$ rg hi *
Argument 'broken�' is not valid UTF-8. Use hex escape sequences to match arbitrary bytes in a pattern (e.g., \xFF).
```

Doing the right thing should be easy - an argument is always an OsStr when it enters the rust program, which is losslessly converted to a `Path`.

BurntSushi:
Yup, this is a silly bug. Note that if you run `rg hi` instead of `rg hi *`, then it works. The specific problem is this very ill-advised code in `src/args.rs`:

```rust
    pub fn parse() -> Result<Args> {
        // Get all of the arguments, being careful to require valid UTF-8.
        let mut argv = vec![];
        for arg in env::args_os() {
            match arg.into_string() {
                Ok(s) => argv.push(s),
                Err(s) => {
                    errored!("Argument '{}' is not valid UTF-8. \
                              Use hex escape sequences to match arbitrary \
                              bytes in a pattern (e.g., \\xFF).",
                              s.to_string_lossy());
                }
            }
        }
        let mut raw: RawArgs =
            Docopt::new(USAGE)
                .and_then(|d| d.argv(argv).version(Some(version())).decode())
                .unwrap_or_else(|e| e.exit());
```

The specific reason why I did this was to catch invalid UTF-8 in *patterns*, but of course, this catches invalid UTF-8 everywhere.

Finally, Docopt has this:

```
    fn get_argv() -> Vec<String> {
        // Hmm, we should probably handle a Unicode decode error here... ---AG
        ::std::env::args().skip(1).collect()
    }
```

... and the argv parser is built around `&str` instead of `OsStr`.

Sigh. I should probably fix this inside of Docopt, but I also want to switch to clap. See #136.

bluss:
Yes, clap makes it possible to do this correctly, it's nice that way.

bluss:
The reason I got to this point was that I'm trying to make a way to list matches in rg in latest-modified-first order. First trying to just feed it files in that order.

BurntSushi:
Oh interesting. You'll also need to pass `-j1`. If you do that, then the results should come back in the order you gave them.

Issue #189, opened 2016-10-19
Consider showing less output for `rg -h`


djc:
This might be a bit subjective, but for me the result of `rg -h` was somewhat suboptimal in giving me much more output than expected, thus quickly scrolling the most useful stuff off the screen.

I think I would like it better if `rg -h` did not show the `Less common options` and `File management options` by default, instead showing a line at the bottom that refers to some extra switch to make `-h/--help` more verbose.

BurntSushi:
FWIW, `rg --help | less` should help, and it's what I normally do when the output of `--help` is large.

I'm not opposed to doing this, but I'm not 100% sure we should because I'm not a fan of making flags harder to discover.

kballard:
`ag --help` shows a one-line (or sometimes two-line) description of all flags. Maybe `rg --help` should do the same, and then something like `rg --help-options` to show the more verbose output? That would make it easier to skim the options to find what you want.

djc:
Yeah, that's another option that would maybe help discovery without making flags harder to discover.

BurntSushi:
I do like that idea better.

devsnek:
I'm not sure if something has happened since this, but the help output is once again enormous, and usually end up doing `rg --help | rg "thing i'm looking for"`.

BurntSushi:
`rg --help` has been basically "the man page" for years. Nothing has changed recently. If you want a denser view, then use `rg -h`.

devsnek:
Oh interesting. I'm not used to `-h` and `--help` doing different things. Perhaps they should be merged and manpages can be generated separately?

BurntSushi:
No. man pages aren't available everywhere. The behavior is staying as is.

lilyball:
`rg -h` still has a large header block that looks identical to the one from `rg --help`. Perhaps we could slim down the `rg -h` header block for just a couple of lines? It seems odd that even the USAGE section isn't immediately visible in `rg -h | less` on a 80x24 terminal.

BurntSushi:
Indeed, I'd be open to that. The initial header block has indeed slowly grown larger over the years.

PR #233, opened 2016-11-13
Switch from Docopt to Clap.

### LINKED ISSUE #136: slowdown with many files on the command line, compared to no arguments
Example using the https://github.com/keybase/client repo:

```
$ ls go/**/*.go | wc -l
1352

$ time rg blahblahblah go/       
rg blahblahblah go/  0.03s user 0.01s system 210% cpu 0.017 total

# <<< this is the slow one >>>
$ time rg blahblahblah go/**/*.go
rg blahblahblah go/**/*.go  0.14s user 0.01s system 104% cpu 0.137 total

$ time grep blahblahblah -r go/  
grep --color=auto blahblahblah -r go/  0.01s user 0.01s system 92% cpu 0.014 total

$ time grep blahblahblah go/**/*.go
grep --color=auto blahblahblah -r go/**/*.go  0.01s user 0.01s system 93% cpu 0.018 total
```

It looks like `rg` is slower when given a (long) list of files on the command line, than it is when it has to find those files itself. `grep` doesn't have this problem, so I don't think it's some confounder like my shell taking a long time to expand the glob.

BurntSushi:
There were two important reasons for the switch:

1. Performance. Docopt does poorly when the argv becomes large, which is
   a reasonable common use case for search tools. (e.g., use with xargs)
2. Better failure modes. Clap knows a lot more about how a particular
   argv might be invalid, and can therefore provide much clearer error
   messages.

While both were important, (1) made it urgent.

Note that since Clap requires at least Rust 1.11, this will in turn
increase the minimum Rust version supported by ripgrep from Rust 1.9 to
Rust 1.11. It is therefore a breaking change, so the soonest release of
ripgrep with Clap will have to be 0.3.

There is also at least one subtle breaking change in real usage.
Previous to this commit, this used to work:

    rg -e -foo

Where this would cause ripgrep to search for the string `-foo`. Clap
currently has problems supporting this use case
(see: https://github.com/kbknapp/clap-rs/issues/742),
but it can be worked around by using this instead:

    rg -e [-]foo

or even

    rg [-]foo

and this still works:

    rg -- -foo

This commit also adds Bash, Fish and PowerShell completion files to the
release, fixes a bug that prevented ripgrep from working on file
paths containing invalid UTF-8 and shows short descriptions in the
output of `-h` but longer descriptions in the output of `--help`.

Fixes #136, #189, #210, #230

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,13 +30,10 @@ matrix:
       env: TARGET=x86_64-apple-darwin
     # Minimum Rust supported channel.
     - os: linux
-      rust: 1.9.0
-      env: TARGET=x86_64-unknown-linux-musl
-    - os: linux
-      rust: 1.9.0
+      rust: 1.11.0
       env: TARGET=x86_64-unknown-linux-gnu
     - os: osx
-      rust: 1.9.0
+      rust: 1.11.0
       env: TARGET=x86_64-apple-darwin
 
 before_install:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,8 +3,8 @@ name = "ripgrep"
 version = "0.2.9"
 dependencies = [
  "bytecount 0.1.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "clap 2.18.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "ctrlc 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "docopt 0.6.86 (registry+https://github.com/rust-lang/crates.io-index)",
  "env_logger 0.3.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "grep 0.1.4",
  "ignore 0.1.5",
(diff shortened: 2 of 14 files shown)


BurntSushi:
cc @kbknapp

kbknapp:
I'm on mobile right now so I still want to read through this in more detail but it looks good so far! And I really, really like some of the things you did that I haven't thought of doing!

I'll update here once kbknapp/clap-rs#742 has been implemented.

BurntSushi:
@kbknapp Thanks! I was hoping you'd take a peek. ^_^

BurntSushi:
Woohoo! @kbknapp Thanks for clap! ^_^

kbknapp:
@BurntSushi I saw you already noticed kbknapp/clap-rs#755 but this should address a few of the items you mentioned. For the use of a hyphen in specific option and not command wide there are plenty of edge cases that don't have 100% solutions (i.e. many people may want many different things). So I've implemented what I think to be the 95% solution.

Namely, edge cases arise when one of two things happen, 

* The option in question accepts multiple values
* The value the user wishes to supply partially or fully matches a valid flag

For the primary edge case example, assume `-e PAT...` where `PAT` can start with a hyphen. Also assume `-v`, `-a`, and `-l` are all valid short flags. Using `-e valid_pattern -val` can cause ambiguities. This can be solved by forcing `-e` to accept only value at a time `Arg::number_of_values(1)` and `Arg::multiple(true)`. This allows `-e PAT -e PAT` but disallows `-e PAT PAT`.

So there are workarounds, I just wanted to ensure you're aware some testing may be required :wink:

Also the ZSH completions should be fixed for you.

BurntSushi:
@kbknapp Ah that sounds perfect! I think I've banned all uses of `--flag val1 val2` anyway since it feels incredibly non-standard to me. :-) Thanks so much for the quick turnaround!

PR #235, opened 2016-11-15
Added elixir to types

jinyeow:
While using ripgrep, I noticed that elixir wasn't one of the types included, so I added it into the types.rs.


BurntSushi:
Thanks!

PR #238, opened 2016-11-18
`Ignore` refactorings

jacwah:
These changes makes the code more readable in my opinion.

### DIFF
--- a/ignore/src/dir.rs
+++ b/ignore/src/dir.rs
@@ -75,7 +75,7 @@ struct IgnoreOptions {
 
 impl IgnoreOptions {
     /// Returns true if at least one type of ignore rules should be matched.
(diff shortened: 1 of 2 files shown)


ignore/src/dir.rs
```diff
@@ -75,7 +75,7 @@ struct IgnoreOptions {
 
 impl IgnoreOptions {
     /// Returns true if at least one type of ignore rules should be matched.
-    fn should_ignores(&self) -> bool {
+    fn has_any_ignore_options(&self) -> bool {
```
jacwah: As mentioned in the commit message, I'm not sure whether this or simply `any` is a better name.

BurntSushi:
LGTM, thank you!

Issue #89, opened 2016-09-25
"--null" or "--print0" option?


gwerbin:
Is there any chance you'd be interested in implementing a feature like Ag's `--null` option, and Ack's (not to mention find's) `--print0` option? If you did, it'd be really easy to make [tag](http://github.com/aykamko/tag/) work with Ripgrep, not to mention more supportive of machine-parsing in general.

BurntSushi:
`grep` has this too. Can we write the specification first? Here's my first attempt:

```
--null
    Whenever a file name is printed, proceed it with a NUL byte.
    This includes printing filenames before matches, and when printing
    a list of matching files such as with --count, --files-with-matches
    and --files.
```

kaushalmodi:
I would like to have this option too..

For now, my woraround is to do..

```
rg foo | tr '\n' '\0'
```

But hopefully internal support of null-separated output makes this faster.

My actual use case is

```
rg -g '' --files | tr '\n' '\0' >! proj_file_list.txt
```

BurntSushi:
@kaushalmodi Unrelated: your `-g ''` is I think completely superfluous.

gwerbin:
@BurntSushi LGreatTM, that's a more detailed spec than any of the other tools that use `--print0`, including both BSD and GNU Find. But maybe it should be both `--print0` and `--null`. Ag has them aliased to each other, and Ack only supports `--print0`. Meanwhile Pt and Git Grep _only_ support `--null`.

kaushalmodi:
@BurntSushi 

> Unrelated: your -g '' is I think completely superfluous.

Thanks for letting me know.. I'll switch to `rg --files` then.

@gwerbin 

There's `-0` too. So `-0` or `--null` or `--print0`.

BurntSushi:
In general, when there's been variation in flag options, I've tried to go with what grep does. I'm not a big fan of lots of aliases for doing the same thing, although I have done it in a few places. For a more niche option like this, I'd rather not increase the number of flags.

gwerbin:
@BurntSushi good to be consistent, then.

@kaushalmodi I avoided `-0` because I always thought of that as an _input_ parsing option as per, e.g. xargs.

emlyn:
Just a nit about the wording of the spec (`Whenever a file name is printed, proceed it with a NUL byte`). I've not heard "proceed" used in this way before, is it a misspelling of "precede", or does it mean "follow"?

BurntSushi:
[...]

PR #239, opened 2016-11-20
Add --files-without-matches flag.

### LINKED ISSUE #138: Feature request: --files-without-matches
One flag that occasionally comes useful is to identify files that _don't_ match a certain expression. Both `ag` and `ack` have `-L`/`--files-without-matches` (effectively the opposite of `-l`/`--files-with-matches`).

One minor problem here is that `rg -L` is already taken for following symlinks.

Curiously, `ag -lv` appears to be a synonym of `-L`. I'm not sure whether this is intentional or not, but it seems like a perfectly reasonable approach to me — in fact, years ago, before I knew of `-L`, this is actually what I first tried with `ack`, and got bogus results (it interprets this combo as "list all files that have at least one line that does not match the given expression", like ripgrep does today). Personally, I don't think I've ever had a use case for this.

mernen:
Performs the opposite of `--files-with-matches`: only shows paths of files that contain zero matches.

Closes #138.

---

This is a rather quick-and-dirty patch: I had zero familiarity with the codebase, and I basically searched for `files.with.matches` to guide me where changes should be made. I've tried to match the overall code style, but I may have missed something.

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -339,6 +339,14 @@ sherlock!(files_with_matches, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
     assert_eq!(lines, expected);
 });
 
+sherlock!(files_without_matches, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
+    wd.create("file.py", "foo");
+    cmd.arg("--files-without-matches");
+    let lines: String = wd.stdout(&mut cmd);
+    let expected = "file.py\n";
+    assert_eq!(lines, expected);
+});
+
 sherlock!(after_context, |wd: WorkDir, mut cmd: Command| {
     cmd.arg("-A").arg("1");
     let lines: String = wd.stdout(&mut cmd);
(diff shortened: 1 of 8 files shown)


tests/tests.rs
```diff
@@ -1059,6 +1067,16 @@ sherlock!(feature_89_files_with_matches, "Sherlock", ".",
 });
 
 // See: https://github.com/BurntSushi/ripgrep/issues/89
+sherlock!(feature_89_files_without_matches, "Sherlock", ".",
```
BurntSushi: I think this should say feature_138.

BurntSushi:
This looks awesome, thank you! I had one little nit but this otherwise looks good to go!

tests/tests.rs
mernen: I actually meant issue #89, since the test is about the interaction between `--null` and this feature.

BurntSushi: Ah, I see, missed that. Sounds good!

BurntSushi:
Thanks!

BurntSushi:
After playing with this, I realized that the equivalent option in grep is `--files-without-match` instead of `--files-without-matches`. That's my bad. I'll fix that!

BurntSushi:
Done in 03f76053228b646a7febed3853ab28f42b83b1ad

mernen:
Oops, my bad! Every other tool uses `--files-without-matches`, and I never bothered to check grep. Wouldn't it be better to also rename functions and fields, though?

BurntSushi:
@mernen No worries, it's totally my fault since I didn't check grep either until just now! It's unfortunate that other tools diverge, but I tend to give grep the highest precedence with this sort of thing because of its ubiquity.

I'm not too picky about the names in the code. There doesn't really need to be a precise correspondence between the CLI flags and the struct fields, so we can take liberties to make them more consistent. Moreover, much of that search code is going to get moved into a separate crate eventually, which will pretty firmly divorce it from whatever CLI interface ripgrep puts in front of it.

PR #240, opened 2016-11-20
Completely re-work colored output and tty handling.

### LINKED ISSUE #37: ^O characters in coloured output
Viewing with `--color always` piped to `less -R`, colours appear correctly but there are ^O characters (\x0F) inserted regularly, which appear in less as inverted `^O`. When removed with `sed 's/\x0f//g'` output appears like screenshot.

Tested release binary version 0.1.16 on Debian Linux, with TERM=screen.linux.

### LINKED ISSUE #51: Color customization
We should definitely try to support customizing colors in the output similar to how `ag` does it. Currently they support 3 color customizations, as per [this issue](https://github.com/ggreer/the_silver_searcher/issues/90):
- `--colors-match`
- `--colors-path`
- `--colors-line-number`

Implementing the part where we set colors based on an argument is not particularly hard (just a matter of translating user input to `term::Attr` enums). Parsing input in the bash terminal format like `ag` does was not difficult to implement. However, this limits Windows usage because we don't have an easy translation between those colors.

How should we go about supporting these color customizations for both Linux, Mac and Windows?

### LINKED ISSUE #94: windows doesn't detect stdin automatically
Namely, running `rg pat < file` will recursively search the current directory instead of `file`. A workaround is to do `rg pat - < file`, which will search file.

This is a consequence of fixing #19. Maybe there is a better way to detect whether stdin is a pipe automatically, but I don't know it. [@vadz in particular points out that it is quite hairy](https://github.com/BurntSushi/ripgrep/issues/19#issuecomment-249453628):

> Oh, sorry, I should have thought about this, this is actually a pretty well-known issue -- but without any good solution, unfortunately. The problem is that Windows doesn't have any concept of PTY, so when a program is running in any kind of terminal emulator, and not the standard console window, its stdin is always connected to a pipe, as far as Windows is concerned. Cygwin applications can distinguish between "real" pipes and normal input from the terminal, but I don't know of any way to do this without linking to cygwin1.dll.

I think you should still be able to detect whether stdin is a TTY when running inside the native console and you should be able to check for this (running inside console, I mean). But everything console-related in Windows is pretty hairy, just look at [our own code](https://github.com/wxWidgets/wxWidgets/blob/v3.1.0/src/msw/app.cpp#L288) for detecting whether we can write to a console...

### LINKED ISSUE #117: Color doesn't work in windows mintty (git bash)
I can't get rg to output color in mintty, ie the default term that comes with git for windows and MSYS2. Tried `--color always` and `-p`, to no avail. Color works fine (by default) in standard cmd.exe.

Color (`--color=auto`) works with standard grep in mintty, or at least the grep build that comes with git for windows.

### LINKED ISSUE #182: --color does not work under Emacs
I'm using rg 0.2.1 and Emacs 25. It looks like `--color` doesn't have effect under Emacs, while other tools such as grep works.
- `M-! echo hello | grep --color=always hello` gives `[01;31m[Khello[m[K`
- `M-! echo hello | rg --color=always hello` gives `hello` (where is the escape code?)

see also https://github.com/emacs-helm/helm/issues/1624#issuecomment-253879188

### LINKED ISSUE #231: choose buffering strategy
Would like to do something like:

tail -f log.txt | rg term1 | rg term2 | etc.

and have data emitted right when data that would match hits the logfile.
Is this possible?

BurntSushi:
This commit completely guts all of the color handling code and replaces
most of it with two new crates: wincolor and termcolor. wincolor
provides a simple API to coloring using the Windows console and
termcolor provides a platform independent coloring API tuned for
multithreaded command line programs. This required a lot more
flexibility than what the `term` crate provided, so it was dropped.
We instead switch to writing ANSI escape sequences directly and ignore
the TERMINFO database.

In addition to fixing several bugs, this commit also permits end users
to customize colors to a certain extent. For example, this command will
set the match color to magenta and the line number background to yellow:

    rg --colors 'match:fg:magenta' --colors 'line:bg:yellow' foo

For tty handling, we've adopted a hack from `git` to do tty detection in
MSYS/mintty terminals. As a result, ripgrep should get both color
detection and piping correct on Windows regardless of which terminal you
use.

Finally, switch to line buffering. Performance doesn't seem to be
impacted and it's an otherwise more user friendly option.

Fixes #37, Fixes #51, Fixes #94, Fixes #117, Fixes #182, Fixes #231


retep998:
cc @Stebalien you may be very interested in the techniques used here to detect what sort of TTY is available on Windows.

BurntSushi:
#94 is where the goods are, including [some code](https://github.com/BurntSushi/ripgrep/issues/94#issuecomment-261761687). Arguably, that might belong in the `atty` crate. cc @softprops

BurntSushi:
I filed an issue against `atty`: https://github.com/softprops/atty/issues/10

How awesome would it be if the entire Rust ecosystem could automatically do correct tty detection on Windows? Oh boy!

Issue #129, opened 2016-09-28
maximum line length


BurntSushi:
I'd like `ripgrep` to have the ability to either hide or trim lines that are very long. Some lines take up my entire screen and are borderline useless to look at. It's possible that finding an intelligent way to shorten them would be best, since my guess is that the actual matched text is much smaller than the full line. However, this is harder to implement.

I don't think this should be enabled by default. It seems a little surprising for `ripgrep` to hide lines like that. In general, I like the work flow of, "run a search, see huge lines, confirm that I don't care about them and run ripgrep again with an option to hide them." It may however be plausible to enable this limit if results are being dumped to a terminal (we already enable colors, line numbers and file headings).

kaushalmodi:
A very good use case is searching in minified js/css.

BurntSushi:
@kaushalmodi Yup. I'm pretty sure that was precisely the thing I hit. :-)

tjayrush:
I'd rather see a Linux-like approach to this. Pipe the file through something like 'tr' first, then pipe the output through 'cut'  Like this: cat file.js | tr 'x' 'y' | ripgrep whatever | cut -c1-100

BurntSushi:
@tjayrush What is the purpose of `tr` there?

In any case, `rg` has the opportunity to show better output here, for example, by only showing the match and a few characters around it.

tjayrush:
I meant for 'tr' to convert some character into new lines (I forgot to finish editing my comment) to alleviate the long line problem. If 'rg' can already limit to the context that's better. 

The larger point was that instead of adding a new command line option, pipe the output (or input) into another already-existing command line program. 

--------
Thomas Jay Rush
jrush@greathill.com

BurntSushi:
I think my comment still stands.

Also, rg searches recursively by default. In that case, it doesn't make
sense to modify the input.

I don't think you need to argue in favor of piping. I'm all for it. But it
isn't always sufficient.

BurntSushi:
OK, in the interest of moving forward, here is a proposed specification.

Add a new flag, `-M/--max-columns NUM`, which is disabled by default. When enabled, lines with more than `NUM` columns are suppressed from output. Columns are counted as the number of bytes in a line.

-----

I mentioned above it might be cool to only show the context around each match (within a line), but I think that's a complicated feature that deserves its own issue/motivation/specification. I think the above flag solves the most pressing problem.

hpsu:
ag has
[...]

Issue #162, opened 2016-10-11
libripgrep


BurntSushi:
This is a tracking issue for the concept known as "libripgrep."

In fact, "libripgrep" probably won't ever be a thing unto itself. Instead, there will be several independent crates that together make up the thing one might think of as "libripgrep." Most of this work is already done. Here's a list:
- [memchr](https://crates.io/crates/memchr) - Fast single byte search.
- [walkdir](https://crates.io/crates/walkdir) - Recursive directory iterator.
- [utf8-ranges](https://crates.io/crates/utf8-ranges) - Generate utf8 automata.
- [regex-syntax](https://crates.io/crates/regex-syntax) - A regex parser (including literal extraction helpers).
- [regex](https://crates.io/crates/regex) - The regex engine itself.
- [globset](https://crates.io/crates/globset) - Regex based glob matching, including glob-set matching.
- [grep](https://crates.io/crates/grep) - Blazing fast line-by-line search. This is where all of the inner literal optimizations happen. (INCOMPLETE.)
- [ignore](https://crates.io/crates/ignore) - Recursively traverse a directory while respecting ignore files and file types.
- parallel directory iterator - Iterate over files in parallel, applying ignore rules. (**This is now done and part of the `ignore` crate.**)

`grep` is incomplete because it exposes only the most basic line-by-line
search. For example, it doesn't handle inverted matching, context handling,
line counts or anything other than "which line matches this regex." Still,
there is considerable regex-specific work in here around literal matching and
other optimizations. The rest of the aforementioned work is done in `ripgrep`
proper. Ideally, all of that should get moved out into the `grep` crate. This
is predominantly a _design_ task, because searching is inextricably tied to
printing, which means the API for all of these features is quite complex.

`gitignore` is a reasonably light wrapper around the `globset` crate. The hope
is that a future `hgignore` crate would be too. Nevertheless, the semantics are
tricky enough that it's worth it to put it into its own crate.

A parallel directory iterator is something I want because so much time is spent
in handling all of the ignore rules. If we can apply parallelism to that, I
kind of expect that we'll get a nice win. Whether a parallel directory iterator
can be design out-of-tree or not still isn't clear. In particular, it's not
clear how coupled this needs to be with respecting ignore rules.

---

In any case, I don't really have the bandwidth to do the design work required to manage contributions from others toward this goal quite yet. I kind of think they're big enough that I would like to at least do the initial work on them.

BurntSushi:
[...]

PR #245, opened 2016-11-22
Limit line lengths

forgottenswitch:
Possible fix for https://github.com/BurntSushi/ripgrep/issues/129
Python to generate a test:
```
with open("long_line_test", "w") as f:
    for n in range(1, 10):
        for i in range(1, 100):
            i_z = str(i).zfill(5)
            f.write("_{}:{}_,".format(n, i_z))
        f.write("\n")
```

### DIFF
--- a/src/atty.rs
+++ b/src/atty.rs
@@ -48,18 +48,24 @@ pub fn on_stdout() -> bool {
 
 /// Returns tty width
 #[cfg(unix)]
-pub fn width() -> usize {
+pub fn width() -> Option<usize> {
     use libc;
 
     let wsz = libc::winsize {
         ws_row: 0, ws_col: 0,
         ws_xpixel: 0, ws_ypixel: 0,
     };
-    let w: u16;
+    let rc: usize;
 
-    unsafe { libc::ioctl(libc::STDOUT_FILENO, libc::TIOCGWINSZ, &wsz); }
-    w = wsz.ws_col;
-    return w as usize;
+    unsafe {
+        rc = libc::ioctl(libc::STDOUT_FILENO, libc::TIOCGWINSZ, &wsz) as usize;
+    }
+    if rc == 0 {
+        let w = wsz.ws_col;
+        return Some(w as usize);
+    } else {
+        return None;
+    }
 }
 
 /// Returns true if there is a tty on stdin.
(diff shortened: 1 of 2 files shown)


BurntSushi:
@forgottenswitch Thanks for the PR! I really appreciate you taking the time to try to fix #129! :-)

Unfortunately, I think there are significant problems with the patch as-is:

1. In #129, I proposed a specification for fixing the issue, but I don't see any evidence that this was considered here. I would much rather start with a simpler solution than jump to something as complex as this.
2. Doing stuff based on the tty width seems a bit too magical for my taste. (Moreover, I don't know how to do it on Windows, and the patch as-is looks like it won't compile on Windows.)
3. The printer code has gotten *significantly* more complicated and I don't understand most of the changes you've made. I *can't* merge code that I don't understand.

Before continuing, I think we should focus entirely on my first objection, (1), and avoid getting distracted with (2) or (3) until we actually decide that this is what we want.

forgottenswitch:
As for (1), one solution would be to make sure each match has even context before and after,
and output each on a dedicated line.
However, what to do if there are dozens of matches on a single line?

What this implements, is try to print as much as possible before the match, and all that fits after.

BurntSushi:
> The proper solution would be to make sure each match has even context before and after.
(except left- and right- most).

It seems very complex and error prone to do this for multiple matches on one very long line. Before I even *consider* supporting something like that, I would absolutely demand a well designed abstraction around it. I'd suggest we iterate on it in the issue tracker, and not move to the PR phase until we have a design that we're both happy with.

I'd also suggest we wait to do this until I get more of #162 done, which I hope to include pulling the printer out of ripgrep proper and into its own crate.

BurntSushi:
I would strongly urge you not to spend more time on this PR until we hammer out a lot more details and a proper specification. This feature is simply too complicated to do otherwise.

forgottenswitch:
Spec
------
Make length of printed lines limit-able width COLUMNS envvar.
If the latter is not a parseable integer, query the pty or Console width.

Every output line should consist of a contiguous input.

Print at least one match on every *output* line.
If the match itself does not fit, drop its end.

On every *output* line, amount of context before first match should equal to that after last one.

------
What this implements, is try to print as much as possible before the match, and all that fits after.
This should be a mistake.

PR #246, opened 2016-11-22
Less and Sass types

mvitz:
Just noticed that Less and Sass are not supported by default. This PR adds types for both.


BurntSushi:
Ah cool, thanks!

PR #250, opened 2016-11-25
Re-add zsh types

SimenB:
Seems like #197 got lost in #202 

I added the missing `.zlogout`, and also search the ones without a leading `.` (which should be in `/etc`). Is there syntax for making a char optional (like `?` in regex)?
The entries are sorted in the order they're read by zsh, with the addition of a `*.zsh` at the end


BurntSushi:
Oh interesting! Not sure how they got lost, sorry about that!

In globs, I don't think there is an equivalent to `?` from regexes. You need to use some other syntax like `{foo,bar}`. ... But the way you wrote it here is better because it's easier for the glob match optimizer to see the literals. :-)

PR #253, opened 2016-11-25
Add stylus type

SimenB:
Seeing as #246 was added, it'd be a shame if the last big dialect wasn't included as well 😄


BurntSushi:
Never heard of it, but it looks popular!

PR #257, opened 2016-11-29
Resolve symbolic links for all input files

### LINKED ISSUE #256: Cannot search in symlink
`db` is a symlink to a directory.

```
$ rg --version
ripgrep 0.3.1
$ rg test db
db: Is a directory (os error 21)
$ rg test db/
(works as expected)
```

I would expect it to traverse the symlink in the first example, or at least to show an error message that is less confusing ;)

jFransham:
Closes #256.

Question: is this the right place to do this? This gives `paths` an IO/filesystem dependency that it didn't already have, which makes this function impure when you might expect it to return the same thing each time.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -344,6 +344,16 @@ impl<'a> ArgMatches<'a> {
         if paths.is_empty() {
             paths.push(self.default_path());
         }
+
+        for path in &mut paths {
+            use std::fs::canonicalize;
+            use std::mem::replace;
+
+            if let Ok(canon) = canonicalize(&path) {
+                replace(path, canon);
+            }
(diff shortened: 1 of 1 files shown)


BurntSushi:
Sorry for the late reply! Unfortunately, this bug is a bit more complex to fix. Canonicalizing all of the paths like this is a non-starter, since the output will use all of the canonicalized paths instead of their logical paths. The right solution here is to read file paths given at the command line as if they were normal files, which required fixing the underlying recursive directory iterators. For example, see: https://github.com/BurntSushi/walkdir/pull/15

I'll be pushing a fix soon!

Issue #193, opened 2016-10-24
[Question] how to search by filenames only?


hh9527:
`--files` will list all files which respect ignore files. Is there a flag to list filenames which match the given filename pattern?

BurntSushi:
From `rg --help`:

```
    -g, --glob GLOB ...        Include or exclude files for searching that
                               match the given glob. This always overrides any
                               other ignore logic. Multiple glob flags may be
                               used. Globbing rules match .gitignore globs.
                               Precede a glob with a '!' to exclude it.
```

YPCrumble:
@BurntSushi this is related to the issue I just raised, #284. The I believe the question here is "How can I search for a pattern in the filename, and return the filenames that include that pattern in the path/filename?"

It appears that the -g flag limits my search _for text within the file_ to the filenames matching the -g pattern. Does that explain the nuance? 

Example: 

If there is a file relative to my directory, `./stories/story.txt` that contains the text "Once upon a time", I want to be able to run `rg -g story` and return `./stories/story.txt`. It appears that the -g flag would return nothing in this query, but it would return "Once upon a time" if I try `rg Once -g story.*`.

hraban:
I have the following alias to shim this in my ~/.bashrc:

```sh
alias rgf='rg --files | rg'
```

This allows you to do:

```sh
$ rgf 2018
lib/lib.es2018.full.d.ts
lib/lib.es2018.promise.d.ts
lib/lib.es2018.intl.d.ts
lib/lib.es2018.regexp.d.ts
lib/lib.es2018.asynciterable.d.ts
lib/lib.es2018.d.ts
```

johnyradio:
> ```shell
> --files
> ```

how would you do that without the alias?

hraban:
@johnyradio just type the command directly:

    rg --files | rg 2018

A Bash alias just substitutes the command for whatever the contents of the alias is, then continues evaluating as usual. That's all an alias is:

    alias rgf="rg --files | rg"
    rgf 2018

=

    rg --files | rg 2018

johnyradio:
Many thanks! Super helpful.

dzhang-b:
`rg --files | rg <pattern> <directory>`  or `rgf <pattern> <directory>`won't work. It will search all contents in the directory, not the filename in that directory.

Piping:
@hraban Thanks for the tip, I updated the logic a bit for future newcomer.
```sh
rf() {
  if [ -z "$2" ]
  then
[...]

Issue #91, opened 2016-09-25
Searching filenames (`-g` option in `ag`)


Valloric:
Ag supports using `-g` to recursively search filenames matching the provided pattern. Ripgrep doesn't have this feature (to the best of my knowledge; I've read the `--help` output three times looking for it); it's the only ag feature holding me back from fully switching to ripgrep. :)

Since ripgrep already has something unrelated mapped to `-g`, the option name should be something else. The incompatibility with ag will be unfortunate, but survivable.

BurntSushi:
I'm pretty sure this has been reported at least several times now. :-) `rg` has `--files` for listing files and `-g` for glob matching.

Valloric:
> I'm pretty sure this has been reported at least several times now. :-) 

Sorry about that! It might be indicative of a documentation "bug"; it seems people are having trouble locating this feature.

Valloric:
Ah, I get it now. It's a combination of two flags: `--files` lists all the files that would have been searched and `-g` can be used to only search files matching a pattern.

This is very non-obvious. It requires users to fully understand `two` separate, unrelated options and to realize that when combined, they can be used to solve their problem.

Since ripgrep will have lots of users coming from ag who will expect a simple single-flag solution, it might be a good idea to have a single flag (much like ag) for this use-case, if for any reason, than to save you the headache of closing all this issue reports. :)

BurntSushi:
I'd like to solve this with better documentation. It's clear this is an important feature to many, but it's decidedly subservient to ripgrep's primary focus, so I'd like to avoid adding extra flags for it.

gabrielmagno:
That would be a great feature. But it is important to notice the differences.

Let's say I have this directory tree:
```
.
└── path
    └── to
        ├── file
        │   ├── pattern (file)
        │   └── pattern.txt (file)
        ├── pattern (file)
        └── patterns
```

- `ag` returns all the files that have "pattern" in **any part** of its **path**
```
$ ag -g "pattern"
path/to/pattern/fileA.txt
path/to/patterns/fileB.txt
path/to/file/pattern
path/to/file/pattern.txt
```

- `rg` returns all the files that have "pattern" as its **name**
```
$ rg --files -g "pattern"
path/to/file/pattern
```

So, ideally, if such a feature is implemented in ripgrep to reproduce what The Silver Search does, it should check the path (not only the name) and also math any part of the string (not an exact match, not a glob).

BurntSushi:
[...]

Issue #75, opened 2016-09-25
Add `-g` option from ag


rubik:
The `-g` option searches only in the file names. Example:

```sh
$ ag -g init | head
scipy/sparse/linalg/eigen/arpack/__init__.py
scipy/sparse/linalg/isolve/__pycache__/__init__.cpython-35.pyc
scipy/sparse/linalg/isolve/__init__.py
scipy/sparse/linalg/__pycache__/__init__.cpython-35.pyc
scipy/sparse/linalg/__init__.py
scipy/sparse/linalg/dsolve/__pycache__/__init__.cpython-35.pyc
scipy/sparse/linalg/dsolve/__init__.py
scipy/sparse/__pycache__/__init__.cpython-35.pyc
scipy/sparse/__init__.py
scipy/sparse/csgraph/__pycache__/__init__.cpython-35.pyc
```

It's quite useful in combination with [`ctrlp.vim`](https://github.com/ctrlpvim/ctrlp.vim), to replace the Vim file finder with a faster alternative. Does it seem like a sensible addition to rg?

BurntSushi:
It's already there. Use `--files`. (Note that there's a bug with it accepting file path arguments in the latest release that is fixed on master. I'm getting a new release out today.)

rubik:
I've seen that option, but it's not equivalent:
```sh
$ rg --files -u | rg -N django
notebook/static/components/codemirror/mode/django/django.js
$ ag -g django
notebook/static/components/codemirror/mode/django/django.js
```

`ag -g` actually searches the file names, while `rg --files` only lists all the files.

BurntSushi:
@rubik Check out `rg --help`, please. `rg --files -g 'django'`. Note that `-g` is short for `--glob`, so you'll need to do `rg --files -g 'django*'` to get the same results. And use `rg --files -g '**/django/**'` to match anywhere in the path.

rubik:
Thanks, I actually read `man rg` but got confused.

pdbradley:
hey @rubik I also use Ctrlp, and I set up ripgrep as a replacment for silversearcher (ag), but one problem I'm having is that using ripgrep, the list of search results includes the full path to the file.  I don't like showing the full path, it clutters things.

Somehow using ag, the results only included the relative path of the project:

app/models/user.rb (for example)

and using rg, the result would show:

/User/myusername/projects/someproject/app/models/user.rb

[...]

Issue #54, opened 2016-09-24
`rg -g foobar` ?


steckerhalter:
I'm used to this from `ag` as a more convenient `find`. Could that be supported? Thanks

BurntSushi:
Try `rg -g foo --files`.

steckerhalter:
Thanks, a bit more to type though :)

Noted just now that `ag` seems to use wildcards by default:

```sh
$ ag -g grandshell
etc/grandshell-theme/README.md
etc/grandshell-theme/grandshell-theme-pkg.el
etc/grandshell-theme/grandshell-theme.el
etc/grandshell-theme/grandshell-theme.png
etc/etc/emacs/grandshell-theme.el
```

and it searches the whole path:

```sh
$ rg -g grandshell --files
$ rg -g *grandshell* --files
./etc/grandshell-theme/grandshell-theme-pkg.el
./etc/grandshell-theme/grandshell-theme.el
./etc/grandshell-theme/grandshell-theme.png
./etc/etc/emacs/grandshell-theme.el
```

BurntSushi:
The goal here isn't too do everything exactly like ag. For example, it would be nice to have one glob syntax. If you want the whole path, try `**/grandshell/**`.

steckerhalter:
Of course, I'm just noticing and expecting `rg` to be just the same. And maybe you can use this information for some purpose...

BurntSushi:
Gotya. Thanks for the bug report, it's useful to be aware of the difference, but I think I'm going to stick with what we have.

PR #285, opened 2016-12-21
Update docs to explain use of -g and --files to search for paths.

### LINKED ISSUE #284: [QUESTION/SUGGESTION] Does/should ripgrep support ag/ack's -g flag for searching filenames?
Per the ack documentation:

> -g PATTERN Print files where the relative path + filename matches PATTERN.  This option can be combined with --color to make it easier to spot the match.

I've searched the documentation and can't find this feature. Is this something that you'd be interested in for ripgrep? I'd be happy to submit a PR for it - I'm not a Rust expert but would love to jump in.

Thanks for writing such a great search tool!

YPCrumble:
@BurntSushi this updates docs to resolve #284, #193, #54, #75, #91, #48. Explains the combined use of `-g` and `--files` flags to replicate the `-g` option in ag/ack.

Looks like I used a later version of pandoc to compile the docs that perhaps did things a little differently. It complained about using sed's `-i` flag (`sed: -i may not be used with stdin`).

Please let me know if any issues/suggestions from your end.

Thank you for your patience and thanks again for building ripgrep!


doc/rg.1.md
```diff
@@ -136,6 +144,10 @@ Project home page: https://github.com/BurntSushi/ripgrep
 --files
 : Print each file that would be searched (but don't search).
 
+    Combine with the -g flag to return matched paths, for example:
+
+    rg -g '<pattern>' --files
```
BurntSushi: As above, change this to `<glob>`.

BurntSushi:
Oh, and could you also put the string `Fixes #284` at the bottom of your commit message? I'm trying to keep things tidy. :-) Thanks!

YPCrumble:
@BurntSushi thank you for your review! Made those changes. Let me know if what I did with the commit message isn't what you meant? Also I escaped the < and > characters around `<glob>` in `rg.1.md` because it appeared that not escaping them made them not appear in `rg.1`. Please me know if this was a mistake?

BurntSushi:
@YPCrumble Looks good to me! I verified that `rg.1` looks right by running `man -l doc/rg.1`.

> Let me know if what I did with the commit message isn't what you meant?

Almost. Ideally, simple PRs like this should just be a single commit. You can do this by amending your previous commit and force pushing your branch, and that commit message is what should contain `Fixes #284` in it. I just went ahead and did that for you though. Thanks again!

YPCrumble:
@BurntSushi got it - thanks for the explanation and making that adjustment for me. Didn't realize I can just take a look at the docs without compiling the code. Glad to have been able to contribute!

PR #287, opened 2016-12-23
Update Cargo.lock to bring in clap 2.19.2 fix for ZSH completions.

lilianmoraru:
Brings in [this fix](https://github.com/kbknapp/clap-rs/issues/771).


BurntSushi:
Thanks!

PR #292, opened 2016-12-25
Add Fedora package

fszymanski:


BurntSushi:
I'm sorry, but I have no hope of maintaining this, and I therefore can't accept it. In the future, it would be better to discuss these changes in a separate issue to avoid doing this work.

---

OK, time to elaborate a bit. The reason why the `pkg` directory exists was because I was initially trying to get ripgrep into distributions for people to use. I personally use Archlinux, so I knew how to do that. I also wanted to get this into the hands of Mac users as easily as possible, so I set up a brew package for it. Since that point, others have taken over the maintenance burden of those packages. (ripgrep is now part of Archlinux and homebrew proper.) There is a homebrew package still in this repo because it permits homebrew users to install a version of ripgrep compiled from Rust nightly, which is faster than the one compiled on Rust stable because of SIMD support. Once SIMD becomes available on Rust stable, it is my intention to remove that homebrew package.

PR #297, opened 2016-12-30
Add types for handling SWIG files.

### LINKED ISSUE #296: Add swig type
That’s [swig](http://swig.org/) which uses `*.i` and `*.def` files. Equivalent of
```
--type-add 'swig:*.i' --type-add 'swig:*.def'
```

mcepl:
SWIG from http://swig.org/

Fixes #296


BurntSushi:
Awesome, thanks!

PR #301, opened 2017-01-03
Add powershell as a known file type

lzybkr:


ignore/src/types.rs
```diff
@@ -142,6 +142,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("pdf", &["*.pdf"]),
     ("php", &["*.php", "*.php3", "*.php4", "*.php5", "*.phtml"]),
     ("pod", &["*.pod"]),
+    ("powershell", &["*.ps1", "*.psd1", "*.psm1"]),
```
BurntSushi: Would it be better to use `ps` here? I feel like `powershell` is a bit long. Is `ps` is accepted abbreviation of PowerShell that folks would recognize?

lzybkr: Indeed, it felt long when I tested it.

`ps` is a frequently used abbreviation *within* PowerShell (e.g. as a prefix for reserved variables or module names), but I'm not sure it's used enough with non-PowerShell tools that it should be considered the best abbreviation.

I also had a slight concern with potential confusion with postscript - one reason PowerShell doesn't use the extension `ps`.

`posh` might be a more widely accepted abbreviation, but I'm not a fan of that abbreviation, partly because that is an actual word.

This would be the second longest type, and there are many somewhat long types, e.g.:

```
markdown
asciidoc
taskpaper
vimscript
powershell
coffeescript
```

At any rate, I posted a survey on [Twitter](https://twitter.com/lzybkr/status/816421545421615105), so we'll see what the feedback looks like.

BurntSushi: > I also had a slight concern with potential confusion with postscript - one reason PowerShell doesn't use the extension ps.

Yeah, I thought about that too, but postscript could use `post` I think.

> This would be the second longest type, and there are many somewhat long types, e.g.:

Yeah, I haven't been terribly conscientious with vetting each name that's added. We do at least have `md` for `markdown`.

> At any rate, I posted a survey on Twitter, so we'll see what the feedback looks like.

Funny. I retweeted it. :-)

BurntSushi:
Looks great! Thanks so much!

PR #303, opened 2017-01-05
Provide a mechanism to compose type definitions

### LINKED ISSUE #83: add a way for file types to include rules from other file types
This would be useful for defining file types that are aggregates of other file types, for example:

```
--type-add src:cpp,py
```

would be equivalent of specifying:

```
--type-add src:*.C,*.H,*.cc,*.cpp,*.cxx,*.h,*.hh,*.hpp,*py
```

CannedYerins:
This extends the syntax of the --type-add flag to allow including the globs of
other already defined types.

Fixes #83.

Hi @BurntSushi, this is pretty much a straight implementation of your spec from #83, except that we are also allowing Unicode numbers in type names in addition to Unicode letters.  Letters were not enough for the default types (curse you, m4).

I am no rust expert so I'd appreciate specific feedback on my add_def changes (or anything, for that matter).  And let me know if I missed any places to put tests or documentation (or if I put the documentation in too many places; it did feel a bit repetitive).

Thanks!


ignore/src/types.rs
```diff
@@ -504,15 +524,48 @@ impl TypesBuilder {
...
+                let types = types_string.split(',');
+                // Check ahead of time to ensure that all types specified are
+                // present and fail fast if not.
+                if types.clone().any(|t| !self.types.contains_key(t)) {
+                    return Err(Error::InvalidDefinition);
+                }
+                for type_name in types {
+                    let globs = self.types.get(type_name).unwrap().globs.clone();
+                    for glob in globs {
+                        self.add(name, &glob)?;
```
BurntSushi: Sorry, but this is going to fail CI because ripgrep's minimum version is currently Rust 1.12. You'll need to use `try!` here.

src/app.rs
```diff
@@ -445,11 +445,24 @@ lazy_static! {
...
+              comma) that have been defined and its rules will automatically \
+              be imported into the type specified. For example, to create a \
+              type called src that matches C++, Python and Markdown files, one \
+              can use:\n\n\
+              --type-add 'src:include:cpp,py,md'\n\n\
+              Additional glob rules can still be added to the src type by \
+              using the --type-add flag again:\n\n\
+              --type-add 'src:include:cpp,py,md' --type-add 'src:*.foo'\n\n\
+              Note that type names must consist only of Unicode letters or \
+              numbers. Punctuation characters are not allowed.");
```
BurntSushi: These docs are great. Can you also add them to the man page in `doc/rg.1.md`? (You'll need to `cd doc && ./convert-to-man` to actually produce the `man` file, which can be viewed with `man -l rg.1`. If you can't get `pandoc` installed easily, that's OK, I can do the final conversion step.)

BurntSushi:
Oh, could you also add an integration test in `tests/tests.rs`? Thanks!

CannedYerins:
I was able to install pandoc but I could not view the man file.  Probably some OSX incompatibility?  Either way, you might want to take a look yourself.

BurntSushi:
Merged in https://github.com/BurntSushi/ripgrep/commit/ed01e80a796b7f5cbc83c63517d884d85f1cebd7

BurntSushi:
Thanks! The doc formatting was a little off. I fixed that and also squashed the commits. Nice work!

PR #307, opened 2017-01-08
Minor code and doc fixes

mernen:
The first commit is an oversight I spotted immediately after having #239 accepted.

The second one is just a minor divergence between the documentation and the actual code ([`Args::threads()`](/BurntSushi/ripgrep/blob/851799f42be4409d4e2498960725e11d0514a0e7/src/args.rs#L667)). I'm assuming the code has the actual intended cap, as it's not too uncommon to find desktop machines with 12+ logical cores, it's still low enough not to accidentally overwhelm a modest (albeit modern) CPU, and it seems to be the more recent commit.

### DIFF
--- a/doc/rg.1
+++ b/doc/rg.1
@@ -364,7 +364,7 @@ single thread.
 .TP
 .B \-j, \-\-threads \f[I]ARG\f[]
 The number of threads to use.
-0 means use the number of logical CPUs (capped at 6).
+0 means use the number of logical CPUs (capped at 12).
 [default: 0]
 .RS
 .RE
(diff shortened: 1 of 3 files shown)


BurntSushi:
@mernen Thanks! :-)

> I'm assuming the code has the actual intended cap, as it's not too uncommon to find desktop machines with 12+ logical cores, it's still low enough not to accidentally overwhelm a modest (albeit modern) CPU, and it seems to be the more recent commit.

Thanks for noticing this. The docs did indeed get out of sync. Note though that the default isn't necessarily 12, it's just [*capped* at 12](https://github.com/BurntSushi/ripgrep/blob/master/src/args.rs#L667).

PR #329, opened 2017-01-17
Add stderr support to wincolor.

pkgw:
Might as well get this bit in — hopefully this one isn't controversial?


BurntSushi:
This PR is in `wincolor 0.1.2` on crates.io. Thanks again!

PR #334, opened 2017-01-18
File Types: Add .eex under Elixir

robi-wan:
.eex is the default file ending for templates using Elixir's template engine EEx.


BurntSushi:
Great, thank you!

PR #339, opened 2017-01-23
Implement Hash for Glob

stuhood:
Implement `Hash` for `Glob`, and re-implement `PartialEq` using only non-redundant struct fields.

I've been using the globset crate as a library (thank you!), and ran across the fact that `Glob` was un-hashable while deduping user inputs.


BurntSushi:
Thanks!

PR #344, opened 2017-01-30
add '*.sass' to sass type

samuelcolvin:
Kind of self explanatory, for some reason `.sass` itself was missed from the sass type. I guess because `sass` is nowadays slightly less popular than `scss` but still very much in use.


BurntSushi:
Awesome, thank you!

PR #356, opened 2017-02-11
Add option --max-columns to truncate long lines

### LINKED ISSUE #129: maximum line length

RalfJung:
This implements what we discussed in <https://github.com/BurntSushi/ripgrep/issues/129#issuecomment-276946474>.

One open question from my side: how bad is it that by always using `CountingReplacer`, we lose the optimization offered by returning something from `Replacer::no_expansion`? In principle, we could use `CountingReplacer` only if `max_columns` is `Some`, therefore only incurring a performance cost if `--max-columns` is set. Is that worth the additional effort?

TODO: Add some tests.

Fixes #129.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -165,6 +165,9 @@ fn app<F>(next_line_help: bool, doc: F) -> App<'static, 'static>
              .short("j").value_name("ARG").takes_value(true)
              .validator(validate_number))
         .arg(flag("vimgrep"))
+        .arg(flag("max-columns")
+             .value_name("NUM").takes_value(true)
+             .validator(validate_number))
         .arg(flag("type-add")
              .value_name("TYPE").takes_value(true)
              .multiple(true).number_of_values(1))
@@ -454,6 +457,10 @@ lazy_static! {
              "Show results with every match on its own line, including \
               line numbers and column numbers. With this option, a line with \
               more than one match will be printed more than once.");
+        doc!(h, "max-columns",
+             "Don't print lines wider than NUM lines.",
+             "Don't print lines wider than NUM lines.  Longer lines are \
+              omitted, and only the nubmer of matches in that line is printed.");
 
         doc!(h, "type-add",
              "Add a new glob for a file type.",
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -46,6 +46,8 @@ pub struct Printer<W> {
     colors: ColorSpecs,
     /// The separator to use for file paths. If empty, this is ignored.
     path_separator: Option<u8>,
+    /// Restrict lines to this many columns.
+    max_columns: Option<usize>
 }
 
 impl<W: WriteColor> Printer<W> {
@@ -65,6 +67,7 @@ impl<W: WriteColor> Printer<W> {
             with_filename: false,
             colors: ColorSpecs::default(),
             path_separator: None,
+            max_columns: None,
         }
     }
 
@@ -141,6 +144,12 @@ impl<W: WriteColor> Printer<W> {
     /// When set, each match is prefixed with the file name that it came from.
     pub fn with_filename(mut self, yes: bool) -> Printer<W> {
         self.with_filename = yes;
+        self
+    }
+
+    /// Configure the max. number of columns used for printing matching lines.
+    pub fn max_columns(mut self, max_columns: Option<usize>) -> Printer<W> {
+        self.max_columns = max_columns;
         self
     }
 
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -3,11 +3,25 @@ use std::fmt;
 use std::path::Path;
 use std::str::FromStr;
 
-use regex::bytes::Regex;
+use regex::bytes::{Regex, Replacer, Captures};
 use termcolor::{Color, ColorSpec, ParseColorError, WriteColor};
 
 use pathutil::strip_prefix;
 use ignore::types::FileTypeDef;
+
+/// CountingReplacer implements the Replacer interface for Regex,
+/// and counts how often replacement is being performed.
+struct CountingReplacer<'r> {
+    replace: &'r [u8],
+    count: &'r mut usize,
+}
+
+impl<'r> Replacer for CountingReplacer<'r> {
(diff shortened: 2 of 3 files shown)


src/app.rs
```diff
@@ -454,6 +457,10 @@ lazy_static! {
              "Show results with every match on its own line, including \
               line numbers and column numbers. With this option, a line with \
               more than one match will be printed more than once.");
+        doc!(h, "max-columns",
+             "Don't print lines wider than NUM lines.",
+             "Don't print lines wider than NUM lines.  Longer lines are \
```
BurntSushi: We should try to keep a consistent style. One space after periods please. :-)

src/printer.rs
```diff
@@ -263,31 +286,57 @@ impl<W: WriteColor> Printer<W> {
             self.write(b":");
         }
         if self.replace.is_some() {
-            let line = re.replace_all(
-                &buf[start..end], &**self.replace.as_ref().unwrap());
+            let mut count = 0;
+            let line = {
+                let replacer = CountingReplacer { count: &mut count, replace : &**self.replace.as_ref().unwrap() };
```
BurntSushi: I think we should write code that is consistent with the style of the surrounding code. Could you please adhere to a 79 column limit (inclusive)? Thanks.

src/printer.rs
```diff
@@ -263,31 +286,57 @@ impl<W: WriteColor> Printer<W> {
...
+                re.replace_all(&buf[start..end], replacer)
+            };
+            if let Some(max_columns) = self.max_columns {
+                if line.len() > max_columns {
```
BurntSushi: `if self.max_columns.map_or(false, |m| line.len() > m)`

src/printer.rs
```diff
@@ -263,31 +286,57 @@ impl<W: WriteColor> Printer<W> {
...
+                    let _ = self.wtr.set_color(self.colors.line());
+                    self.write(format!(" [Omitted long line with {} replacements]", count).as_bytes());
```
BurntSushi: I don't think there should be a leading space here.

src/printer.rs
```diff
@@ -263,31 +286,57 @@ impl<W: WriteColor> Printer<W> {
...
-            let line = re.replace_all(
-                &buf[start..end], &**self.replace.as_ref().unwrap());
+            let mut count = 0;
+            let line = {
+                let replacer = CountingReplacer { count: &mut count, replace : &**self.replace.as_ref().unwrap() };
+                re.replace_all(&buf[start..end], replacer)
+            };
+            if let Some(max_columns) = self.max_columns {
+                if line.len() > max_columns {
+                    let _ = self.wtr.set_color(self.colors.line());
```
BurntSushi: I think I would rather see this use the match color instead of the line number color.

src/printer.rs
```diff
@@ -263,31 +286,57 @@ impl<W: WriteColor> Printer<W> {
...
-        }
-        if buf[start..end].last() != Some(&self.eol) {
-            self.write_eol();
+            // write_matched_line guarantees to write a newline.
         }
     }
 
     fn write_matched_line(&mut self, re: &Regex, buf: &[u8]) {
+        if let Some(max_columns) = self.max_columns {
+            if buf.len() > max_columns {
```
BurntSushi: `if self.max_columns.map_or(false, |m| buf.len() > m)`

src/printer.rs
```diff
@@ -263,31 +286,57 @@ impl<W: WriteColor> Printer<W> {
...
+                let count = re.find_iter(buf).count();
+                let _ = self.wtr.set_color(self.colors.line());
+                self.write(format!(" [Omitted long line with {} matches]", count).as_bytes());
```
BurntSushi: I don't think we need the leading space here.

src/printer.rs
```diff
@@ -263,31 +286,57 @@ impl<W: WriteColor> Printer<W> {
...
-            self.write_eol();
+            // write_matched_line guarantees to write a newline.
         }
     }
 
     fn write_matched_line(&mut self, re: &Regex, buf: &[u8]) {
+        if let Some(max_columns) = self.max_columns {
+            if buf.len() > max_columns {
+                let count = re.find_iter(buf).count();
+                let _ = self.wtr.set_color(self.colors.line());
```
BurntSushi: "match" color instead of "line" color.

src/printer.rs
```diff
@@ -312,6 +361,15 @@ impl<W: WriteColor> Printer<W> {
         if let Some(line_number) = line_number {
             self.line_number(line_number, b'-');
         }
+        if let Some(max_columns) = self.max_columns {
+            if end - start > max_columns {
```
BurntSushi: `if self.max_columns.map_or(false, |m| end - start > m)`

src/printer.rs
```diff
@@ -312,6 +361,15 @@ impl<W: WriteColor> Printer<W> {
...
+                let _ = self.wtr.set_color(self.colors.line());
+                self.write(format!(" [Omitted long context line]").as_bytes());
```
BurntSushi: Drop leading space.

BurntSushi:
@RalfJung Thank you for working on this. :-)

BurntSushi:
I'd like to push a new release out soon and you got this PR so close to completion that I picked it up in #406. Thanks so much!

RalfJung:
Ah, so sorry for not coming back to this earlier. Busy times. I was hoping to do something this week. Thanks for completing this!

PR #378, opened 2017-02-23
Add .log as FileType

deepy:


BurntSushi:
Sounds reasonable to me, thanks!

Issue #369, opened 2017-02-18
Ignoring files based on their size


mensfeld:
Hey, is there an option to use ripgrep and exclude files based on their size? I have a lot of +10MB text files that I don't want to grep and would love to be able to exclude them all. Thank you.

BurntSushi:
No. Can you use some other attribute of the text files? You can specify paths to ignore at the command line with `-g '!dir/to/ignore'` or `-g '!*.txt'`.

mensfeld:
Hello @BurntSushi. Thank you for such fast reaction. Unfortunately I can't. I use rg as a part in a "chain" where I get some locations and I need to do search based on that. If it is not possible, I will just wrap it around with a shell script with a list of dirs/files to ignore.

BurntSushi:
It is indeed not possible right now. It would have to be explicitly added to ripgrep as a new flag. Which I'd be OK with, since it should be relatively easy to add. It will have a performance cost, but that seems fine since the end user would have to specifically ask for it anyway.

mensfeld:
@BurntSushi - yes it would have a perf. impact but only when needed which shouldn't be that bad. For now I will just find those files on my own and exclude them. Thank you

tiehuis:
I'll probably end up looking at implementing this over the next few days. Just to outline the intended approach from a quick glance before beginning.

 - Add a new size matcher into `Ignore` and co in the `ignore` crate (in file `dir.rs`)
 - Add the size check into `matched` (in file `dir.rs`)
 - Add the top-level options into all the required locations
 - Add some tests

There probably should be some nice intuitive values expected for the command line option, too (i.e. `--limit-size=2K` `--limit-size=1G`).

BurntSushi:
> Add a new size matcher into Ignore and co in the ignore crate (in file dir.rs)

I think the `ignore` crate is probably the right place, but it feels like the filter should be implemented in `walk.rs`, not `dir.rs`. The stuff in `dir.rs` is pretty tightly coupled to glob matching, where as the stuff in `walk.dir` is generalized to recursive directory walking. Note that you will have to implement the check on both the single threaded and parallel walkers.

> There probably should be some nice intuitive values expected for the command line option, too (i.e. --limit-size=2K --limit-size=1G).

That seems like nice extra credit. Feel free to punt on that if you like. :-)

BurntSushi:
Fixed in https://github.com/BurntSushi/ripgrep/pull/385

PR #385, opened 2017-02-28
Add Max Filesize Option

tiehuis:
See #369.

There is now a build dependency on `regex`/ This is due to using it to validate the expect input format for the new option. I did seriously consider just doing this manually since it isn't very complicated but since there is a very similar function when actual parsing the arguments, I've opted for minimality.


ignore/src/walk.rs
```diff
@@ -1253,6 +1278,34 @@ fn check_symlink_loop(
     Ok(())
 }
 
+fn skip_filesize(
+    max_filesize: Option<u64>,
+    path: &Path,
+    ent: &Option<Metadata>,
+    is_dir: bool
+) -> bool {
```
BurntSushi: This function cannot accept a `Metadata` because that implies *every* file is stat'd regardless of whether a max filesize option is set or not. You need to move the metadata access to occur only after you know there is a file size filter.

src/app.rs
```diff
@@ -491,3 +502,24 @@ lazy_static! {
 fn validate_number(s: String) -> Result<(), String> {
     s.parse::<usize>().map(|_|()).map_err(|err| err.to_string())
 }
+
+fn validate_max_filesize(s: String) -> Result<(), String> {
+    let re = Regex::new(r#"^(\d+)([KMG])?$"#).unwrap();
```
BurntSushi: `\d` is Unicode aware, which includes a lot more than `[0-9]`, so I think you want to change this to `[0-9]`.

BurntSushi: I don't think you need the heredoc syntax either, since you aren't using any `"` inside the regex. `r"^([0-9]+)([KMG])?$"` will suffice.

src/app.rs
```diff
@@ -491,3 +502,24 @@ lazy_static! {
...
+    let caps = try!(re.captures(&s)
+                      .ok_or("invalid format for max-filesize argument"));
+
+    let value = caps.get(1);
+    let suffix = caps.get(2).map(|x| x.as_str());
+
+    match value {
```
BurntSushi: You don't need this match. You can unwrap the `value` since it is guaranteed to match if the entire regex matched. e.g., `try!(caps[1].parse::<u64>().map_err(|err| err.to_string()))`.

src/args.rs
```diff
@@ -779,6 +782,33 @@ impl<'a> ArgMatches<'a> {
         btypes.build().map_err(From::from)
     }
 
+    /// Parses the max-filesize argument option into a byte count.
+    fn max_filesize(&self) -> Result<Option<u64>> {
```
BurntSushi: This duplication isn't acceptable, sorry.

Please just get rid of the clap validator. The error can be reported from here. That will also get rid of the build dep on regex.

src/args.rs
```diff
@@ -779,6 +782,33 @@ impl<'a> ArgMatches<'a> {
...
+        use regex::Regex;
+
+        let max_filesize = match self.value_of_lossy("max-filesize") {
+            Some(x) => x,
+            None => return Ok(None)
+        };
+
+        let re = Regex::new(r#"^(\d+)([KMG])?$"#).unwrap();
+        let caps = try!(re.captures(&max_filesize)
+                          .ok_or("invalid format for max-filesize argument"));
```
BurntSushi: See comments above about the regex itself and handling the captures after-the-fact.

tests/tests.rs
```diff
@@ -432,6 +432,18 @@ sherlock!(context_line_numbers, "world|attached",
...
+    cmd.arg("--max-filesize").arg("44444444444444444444");
+    wd.assert_err(&mut cmd);
+});
+
+sherlock!(max_filesize_parse_error_suffix, "Sherlock", ".",
+|wd: WorkDir, mut cmd: Command| {
+    cmd.arg("--max-filesize").arg("45k");
+    wd.assert_err(&mut cmd);
+});
+
```
BurntSushi: Can you add a test that checks the functionality of `--max-filesize` itself? I realize you have some unit tests, but checking that it works in an integration test will cover some important pieces of code, like the filesize parser and whether the limit is actually passed to the directory iterator.

tiehuis:
Thanks for the quick response. Here is an updated version.

A note about the `skip_filesize` function. I've currently just added appropriate checks prior to calling to ensure that it is necessary to even call. The two places it is called have slightly differing types `walkdir::DirEntry` and `walk::DirEntry` with no common impl for the metadata function. For the moment this seems the cleanest way to me and I've added a note to the function regarding the performance concern in actually calling it.

BurntSushi:
@tiehuis I will see if I can sort out the `skip_filesize` issue you mention next time I'm in that code. For now it looks OK. Thanks. :-)

BurntSushi:
Urk. I missed that this hadn't been squashed. In the future, I will try to do better to notice and remind folks.

PR #392, opened 2017-03-04
Add enclosing group to alternations in globs

### LINKED ISSUE #391: set globs have incorrect behavior
Repro:

```
$ git clone git://github.com/bag-man/dotfiles
$ cd dotfiles
$ rg --no-ignore --hidden --follow --files --glob '!{.git,node_modules,plugged}/**' --glob '*.{js,json,php,md,styl,scss,sass,pug,html,config,py,cpp,c,go,hs}'
```

The output includes files in the `.git` directory. Interestingly, removing some of the extensions in the second `--glob` impacts the output in unexpected ways. Changing the first glob to not use a recursive glob also seems to make it work.

If I look at the debug output, it looks like the generated regex for the second glob is:

```
(?-u)^(?:/?|.*/).*\\.hs|c|cpp|py|config|html|pug|sass|scss|styl|md|php|json|js$
```

Which is just plain wrong. The alternation should be in a group.

tiehuis:
Fixes #391.

### DIFF
--- a/globset/src/glob.rs
+++ b/globset/src/glob.rs
@@ -659,9 +659,18 @@ impl Tokens {
                     for pat in patterns {
                         let mut altre = String::new();
                         self.tokens_to_regex(options, &pat, &mut altre);
-                        parts.push(altre);
+                        if !altre.is_empty() {
+                            parts.push(altre);
+                        }
                     }
-                    re.push_str(&parts.join("|"));
+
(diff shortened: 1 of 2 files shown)


tiehuis:
Seems like the CI tests are failing within globset. They pass for me locally so will have another look into this and see if its obvious.

PR #400, opened 2017-03-09
Copy _rg.ps1 to windows zip files

dstcruz:
Please help me verify that the `_rg.ps1` file in the appveyor script is correct.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -40,6 +40,7 @@ before_deploy:
   - cargo build --release
   - mkdir staging
   - copy target\release\rg.exe staging
+  - copy target\release\_rg.ps1 staging
   - cd staging
     # release zipfile will look like 'rust-everywhere-v1.2.3-x86_64-pc-windows-msvc'
   - 7z a ../%PROJECT_NAME%-%APPVEYOR_REPO_TAG_NAME%-%TARGET%.zip *


BurntSushi:
I fear that this may not be quite right. Did you confirm this works in a local build? At least on Darwin and Linux, the generated completion files are nestled a bit further down: https://github.com/BurntSushi/ripgrep/blob/master/ci/before_deploy.sh#L23

BurntSushi:
(Unfortunately, CI isn't going to test this for us because this code doesn't trigger until a tagged release. If it's too much trouble, I can probably just take this on when I do the next release and iterate until it works.)

dstcruz:
I must admit to being completely ignorant about all rust things. I did install rust/cargo on my machine and ran the `cargo build --release` command that I saw in the AppVeyor script. However, I did not get an `rg.exe` file in `target\release\`. I can see than the shell completion file is in `target\release\build\ripgrep-[some-hash]\out\'. I probably wrongly assumed that they would be at the same level as the `rg.exe`.

I see that the AppVeyor script is [currently running?](https://ci.appveyor.com/project/BurntSushi/ripgrep/build/1.0.313/job/p6714xn58pnfylgh). Can we wait to see what error that might throw? Better yet, is there any way to explore the filesystem from that run?

dstcruz:
Ugh. Sorry. Just noticed that the AppVeyor script is only running tests.

dstcruz:
Ok, just a sec. I think I got this fixed.

dstcruz:
Since I can't commit a fix to this pull request, I will close it and add a corrected version.

BurntSushi:
@dstcruz You can update existing PRs by either adding commits or just amending your commit and force pushing. (The latter is preserved since it will yield a more serviceable history.)

dstcruz:
Thanks. As you can see... I haven't used GitHub much either. I'm getting there...

PR #406, opened 2017-03-13
Add new -M/--max-columns option.

### LINKED ISSUE #129: maximum line length

BurntSushi:
This permits setting the maximum line width with respect to the number
of bytes in a line. Omitted lines (whether part of a match, replacement
or context) are replaced with a message stating that the line was
elided.

Fixes #129

### DIFF
--- a/doc/rg.1.md
+++ b/doc/rg.1.md
@@ -203,6 +203,10 @@ Project home page: https://github.com/BurntSushi/ripgrep
 -L, --follow
 : Follow symlinks.
 
+-M, --max-columns *NUM*
+: Don't print lines longer than this limit in bytes. Longer lines are omitted,
+  and only the number of matches in that line is printed.
+
 -m, --max-count *NUM*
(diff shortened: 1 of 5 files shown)


BurntSushi:
cc @RalfJung

RalfJung:
I have to say I preferred the look of this with a space between the colon and the "line omitted" mark, but oh well. Thanks for adopting my PR :)

PR #408, opened 2017-03-14
Add powershell completions to build artifacts

leafgarland:
Use a `ps:` (powershell) command to copy the completions file so that we
can use directory globbing to find the file.


dstcruz:
Thanks for looking into this!

BurntSushi:
@leafgarland My hero! Thank you! :-)

@dstcruz Thank you for all your effort on this too of course. :-)

PR #412, opened 2017-03-19
Removed debug info from release builds

Aaronepower:
Having debug info enabled on release causes bloated binaries on Unix based systems. This affects people who use `cargo install ripgrep`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -50,6 +50,3 @@ lazy_static = "0.2"
 [features]
 avx-accel = ["bytecount/avx-accel"]
 simd-accel = ["bytecount/simd-accel", "regex/simd-accel"]
-
-[profile.release]
-debug = true


BurntSushi:
What is the actual problem here? I have debug enabled on release binaries purposefully.

Aaronepower:
From IRC: `ripgrep's executable is 26MB, 24MB of which is debug information`. This seems a bit much, what is the reasoning for leaving it on in release, profiling?

BurntSushi:
I understand that it makes the binary big. What i would like to know is *why* that's a problem.

Aaronepower:
Other than the inclination to not waste disk space if you don't need to. I think the main problem is the inconsistency as if I was to use one of the package managers using the GitHub releases I would have a binary that is ~1-5MB, whereas with cargo I get 20+MB.

BurntSushi:
@Aaronepower I've moved this discussion to #413 because it really shouldn't be tied to a particular implementation strategy.

kpp:
Well, you can manually strip debug info using `strip`.

BurntSushi:
I decided to just apply `strip` in CI, which I did in https://github.com/BurntSushi/ripgrep/commit/30ca3ecca65d06ebc1d0794f3968397709aff9a8

PR #421, opened 2017-03-28
Add -0 shortcut for --null #419

### LINKED ISSUE #419: Add -0 shortcut for --null
I've use ag with `-0` a lot to safely pipe it into xargs. Any chance you could add the same alias so I don't have to type as much when switching to rg?

kpp:
Fixes #419


BurntSushi:
@kpp Could you also rebase with master? I just pushed a fix to hopefully get CI working.

kpp:
@BurntSushi fixed commit msg + rebase

PR #427, opened 2017-03-30
Apply binary name to clap

### LINKED ISSUE #426: Autogenerated usage mismatches binary name
```
$ rg -h | head -n14 | tail -n 1
USAGE:ripgrep [OPTIONS] <pattern> [--] [path]...
```

It should be `USAGE:rg [OPTIONS] <pattern> [--] [path]...`

kpp:
Fixes #426 

Example:
```
$ cargo run -- -h | head -n14 | tail -n 1
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/rg -h`
USAGE: rg [OPTIONS] <pattern> [--] [path]...
```

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -24,7 +24,7 @@ const TEMPLATE: &'static str = "\
 {author}
 {about}
 
-USAGE:{usage}
+USAGE: {usage}
 
 ARGS:
 {positionals}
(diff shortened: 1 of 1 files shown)


BurntSushi:
I don't think this is correct. The docs of `bin_name` say:

> Overrides the system-determined binary name. This should only be used when absolutely necessary, such as when the binary name for your application is misleading, or perhaps not how the user should invoke your program.

clap should be able to pick up the binary name automatically. So I think this is a bug in clap.

Issue #377, opened 2017-02-23
Specify color setting of column number


ShikChen:
Currently (0.4.0) the `--colors` option only supports types path, line and match. There is no way to specify the color of column number now. So when using with `--column`, the column part is uncolored and has the same style of matching line, which is a little unpleasant.

BurntSushi:
@ShikChen Out of curiosity, what do you use the column information for?

ShikChen:
I use it for vim, which needs the column information to jump into exact location. The above statement is also true for option `--vimgrep`. More precisely, I use it with [fzf.vim](https://github.com/junegunn/fzf.vim).

BurntSushi:
@ShikChen Sorry, I wasn't clear. Color information is *for humans*. What do *you* use the column information for? I understand that automated tools may read it to do useful things, but automated tools don't need colors.

ShikChen:
But I will see the column information in the UI of fzf.vim, if I want to jump with exact location. When the matching line starts without indent (white spaces), my eyes cannot easily parse boundary between the column number and matching line, especially when that line is starting with numbers.

I agree that most of time we don't need line/column numbers. But it's still possible that someone is manipulating tabular data like csv, and need line/column information to distinguish between multiple matches.

I can understand if you decide not to add color to column number. The default search tool in fzf is [Ag](https://github.com/ggreer/the_silver_searcher), which doesn't support it either. Nobody complains about that so maybe I am too sensitive.

Inspired from your comment, now I have some vimscripts that hiding the line/column part in fzf.

BurntSushi:
FWIW, I do actually think line numbers are useful to be able to read because it helps orient our minds to where something is in a file (and possibly relative to other things as well). But column numbers seem less useful since lines are generally pretty short and the matches themselves are highlighted with color.

I'll leave this open for now and mull on it some more.

ShikChen:
I found another scenario that I want the column number recently, although it's a little complex and weird. I have mapped a hotkey in vim to search the current word under the cursor in ripgrep with fzf, and I can interactively filter the results in fzf.  I sometimes use the column number as filter. For example, when tracing the code of Bash, I use the column number `:1:` to select the function definition. (I know that I can use `^` to match the prefix in normal ripgrep/fzf usage, but it's just harder to use in my scenario.)

kpp:
The case of reading columns is `-o` flag.

![Image of example](http://oi65.tinypic.com/9gzeqq.jpg)

[...]

PR #428, opened 2017-03-30
Improves Printer, adds ColorSpecs for column and separator

kpp:
I added coloring for columns and separators. Also I fixed some bugs, when paths in context were not colored.

Image of usage

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -204,22 +204,14 @@ impl<W: WriteColor> Printer<W> {
     pub fn path<P: AsRef<Path>>(&mut self, path: P) {
         let path = strip_prefix("./", path.as_ref()).unwrap_or(path.as_ref());
         self.write_path(path);
-        if self.null {
-            self.write(b"\x00");
-        } else {
-            self.write_eol();
-        }
+        self.write_path_eol();
     }
 
     /// Prints the given path and a count of the number of matches found.
     pub fn path_count<P: AsRef<Path>>(&mut self, path: P, count: u64) {
         if self.with_filename {
             self.write_path(path);
-            if self.null {
-                self.write(b"\x00");
-            } else {
-                self.write(b":");
-            }
+            self.write_path_sep(b':');
         }
         self.write(count.to_string().as_bytes());
         self.write_eol();
@@ -227,13 +219,11 @@ impl<W: WriteColor> Printer<W> {
 
     /// Prints the context separator.
     pub fn context_separate(&mut self) {
-        // N.B. We can't use `write` here because of borrowing restrictions.
         if self.context_separator.is_empty() {
             return;
         }
-        self.has_printed = true;
         let _ = self.wtr.write_all(&self.context_separator);
-        let _ = self.wtr.write_all(&[self.eol]);
+        self.write_eol();
     }
 
     pub fn matched<P: AsRef<Path>>(
@@ -280,16 +270,17 @@ impl<W: WriteColor> Printer<W> {
     ) {
         if self.heading && self.with_filename && !self.has_printed {
             self.write_file_sep();
-            self.write_heading(path.as_ref());
+            self.write_path(path);
+            self.write_path_eol();
         } else if !self.heading && self.with_filename {
-            self.write_non_heading_path(path.as_ref());
+            self.write_path(path);
+            self.write_path_sep(b':');
         }
         if let Some(line_number) = line_number {
             self.line_number(line_number, b':');
         }
         if let Some(c) = column {
-            self.write((c + 1).to_string().as_bytes());
-            self.write(b":");
+            self.column_number(c + 1, b':');
         }
         if self.replace.is_some() {
             let mut count = 0;
@@ -299,11 +290,8 @@ impl<W: WriteColor> Printer<W> {
                 re.replace_all(&buf[start..end], replacer)
             };
             if self.max_columns.map_or(false, |m| line.len() > m) {
-                let _ = self.wtr.set_color(self.colors.matched());
-                let msg = format!(
-                    "[Omitted long line with {} replacements]", count);
-                self.write(msg.as_bytes());
-                let _ = self.wtr.reset();
+                let msg = format!("[Omitted long line with {} replacements]", count);
+                self.write_colored(msg.as_bytes(), |colors| colors.matched());
                 self.write_eol();
                 return;
             }
@@ -320,10 +308,8 @@ impl<W: WriteColor> Printer<W> {
     fn write_matched_line(&mut self, re: &Regex, buf: &[u8]) {
         if self.max_columns.map_or(false, |m| buf.len() > m) {
             let count = re.find_iter(buf).count();
-            let _ = self.wtr.set_color(self.colors.matched());
             let msg = format!("[Omitted long line with {} matches]", count);
-            self.write(msg.as_bytes());
-            let _ = self.wtr.reset();
+            self.write_colored(msg.as_bytes(), |colors| colors.matched());
             self.write_eol();
             return;
         }
@@ -333,9 +319,7 @@ impl<W: WriteColor> Printer<W> {
             let mut last_written = 0;
             for m in re.find_iter(buf) {
                 self.write(&buf[last_written..m.start()]);
-                let _ = self.wtr.set_color(self.colors.matched());
-                self.write(&buf[m.start()..m.end()]);
-                let _ = self.wtr.reset();
+                self.write_colored(&buf[m.start()..m.end()], |colors| colors.matched());
                 last_written = m.end();
             }
             self.write(&buf[last_written..]);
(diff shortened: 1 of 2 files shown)


src/app.rs
```diff
@@ -249,12 +249,12 @@ lazy_static! {
...
-              style. A special format, {type}:none, will clear all color \
-              settings for {type}.\n\nFor example, the following command will \
-              change the match color to magenta and the background color for \
-              line numbers to yellow:\n\n\
+              should be one of path, line, column, match or separator. \
+              {attribute} can be fg, bg or style. {value} is either a color \
+              (for fg and bg) or a text style. A special format, {type}:none, \
+              will clear all color settings for {type}.\n\nFor example, the \
+              following command will change the match color to magenta and \
+              the background color for line numbers to yellow:\n\n\
```
BurntSushi: These docs also need to be added to `doc/rg.1.md` and the `doc/rg.1` file needs to be regenerated. (If it's hard for you to install pandoc, then I can do the regeneration.)

src/printer.rs
```diff
@@ -204,36 +204,26 @@ impl<W: WriteColor> Printer<W> {
...
         self.write(count.to_string().as_bytes());
         self.write_eol();
     }
 
     /// Prints the context separator.
     pub fn context_separate(&mut self) {
-        // N.B. We can't use `write` here because of borrowing restrictions.
-        if self.context_separator.is_empty() {
-            return;
+        if !self.context_separator.is_empty() {
```
BurntSushi: Why change the coding style here? I much prefer early returns.

src/printer.rs
```diff
@@ -204,36 +204,26 @@ impl<W: WriteColor> Printer<W> {
...
+            let separator = self.context_separator.clone();
```
BurntSushi: The previous code called out the borrowing restriction, but you removed that comment and added a `clone`. The whole point of the previous code was to avoid the `clone`. ;-)

src/args.rs
```diff
@@ -731,8 +731,10 @@ impl<'a> ArgMatches<'a> {
         let mut specs = vec![
             "path:fg:magenta".parse().unwrap(),
             "line:fg:green".parse().unwrap(),
+            "column:fg:yellow".parse().unwrap(),
```
BurntSushi: Please see #377, which asks for colors on column numbers. What's the use case? Similarly for separators.

src/printer.rs
```diff
@@ -299,11 +290,8 @@ impl<W: WriteColor> Printer<W> {
                 re.replace_all(&buf[start..end], replacer)
             };
             if self.max_columns.map_or(false, |m| line.len() > m) {
-                let _ = self.wtr.set_color(self.colors.matched());
-                let msg = format!(
-                    "[Omitted long line with {} replacements]", count);
-                self.write(msg.as_bytes());
-                let _ = self.wtr.reset();
+                let msg = format!("[Omitted long line with {} replacements]", count);
+                self.write_colored(msg.as_bytes(), |colors| { colors.matched() });
```
BurntSushi: I think the braces in your closure here are superfluous.

src/printer.rs
```diff
@@ -320,10 +308,8 @@ impl<W: WriteColor> Printer<W> {
     fn write_matched_line(&mut self, re: &Regex, buf: &[u8]) {
         if self.max_columns.map_or(false, |m| buf.len() > m) {
             let count = re.find_iter(buf).count();
-            let _ = self.wtr.set_color(self.colors.matched());
             let msg = format!("[Omitted long line with {} matches]", count);
-            self.write(msg.as_bytes());
-            let _ = self.wtr.reset();
+            self.write_colored(msg.as_bytes(), |colors| { colors.matched() });
```
BurntSushi: I think the braces in your closure here are superfluous.

src/printer.rs
```diff
@@ -378,63 +359,62 @@ impl<W: WriteColor> Printer<W> {
         }
     }
 
-    fn write_heading<P: AsRef<Path>>(&mut self, path: P) {
-        let _ = self.wtr.set_color(self.colors.path());
-        self.write_path(path.as_ref());
-        let _ = self.wtr.reset();
+    fn separator(&mut self, sep: &[u8]) {
+        self.write_colored(&sep, |colors| { colors.separator() });
```
BurntSushi: I think the braces in your closure here are superfluous.

src/printer.rs
```diff
@@ -333,9 +319,7 @@ impl<W: WriteColor> Printer<W> {
             let mut last_written = 0;
             for m in re.find_iter(buf) {
                 self.write(&buf[last_written..m.start()]);
-                let _ = self.wtr.set_color(self.colors.matched());
-                self.write(&buf[m.start()..m.end()]);
-                let _ = self.wtr.reset();
+                self.write_colored(&buf[m.start()..m.end()], |colors| { colors.matched() });
```
BurntSushi: I think the braces in your closure here are superfluous.

src/printer.rs
```diff
@@ -378,63 +359,62 @@ impl<W: WriteColor> Printer<W> {
...
+        match self.path_separator {
+            None => self.write_colored(path_str, |colors| { colors.path() }),
+            Some(sep) => {
+                let transformed_path: Vec<_> = path_str.iter().map(|&b| {
+                    if b == b'/' || (cfg!(windows) && b == b'\\') {
+                        sep
+                    } else {
+                        b
+                    }
+                }).collect();
```
BurntSushi: This introduces an additional allocation that wasn't there before, probably as a result of refactoring `path_to_str_native` to return a slice of bytes. Maybe it should return a cow instead?

src/printer.rs
BurntSushi: OK, this happens a lot. I'll stop picking on it.

src/args.rs
kpp: The use case is to improve readability. Color helps humans to orient.

BurntSushi: Right, but if you read #377, then I'm specifically calling into question why humans are reading column numbers, especially when the match itself is colored.

src/printer.rs
kpp: Ur the boss.

src/args.rs
kpp: I don't know =( I wanted to color everything in the output...

src/printer.rs
kpp: https://doc.rust-lang.org/std/primitive.slice.html#method.to_vec

The above `let mut path = path.to_vec();` was that allocation. I replaced for with map. No additional allocations are made by this code.

src/printer.rs
```diff
@@ -204,36 +204,27 @@ impl<W: WriteColor> Printer<W> {
...
     /// Prints the context separator.
     pub fn context_separate(&mut self) {
-        // N.B. We can't use `write` here because of borrowing restrictions.
         if self.context_separator.is_empty() {
             return;
         }
-        self.has_printed = true;
-        let _ = self.wtr.write_all(&self.context_separator);
-        let _ = self.wtr.write_all(&[self.eol]);
+        let separator = self.context_separator.clone();
```
kpp: I thought that possibility to color it overcomes 1 additional allocation?

src/app.rs
kpp: I will update `doc/rg.1.md`. There is only `pandoc=1.16.0.2` in my system, so do the regeneration on your on, please.

src/printer.rs
BurntSushi: D'oh. I missed the `to_vec` in the old code.

src/printer.rs
BurntSushi: I don't think those are our only choices. You should be able to add color without additional allocations. The code just might be a bit more circuitous.

src/args.rs
BurntSushi: I think I'd prefer if you backed out the addition of `column` and `separator` color styles. Whether we add that or not should really be decided in #377.

src/printer.rs
kpp: Will you help me? I am a rust newbie)

src/args.rs
kpp: But... the pic... prettier... =(

src/printer.rs
BurntSushi: The problem is that `self.separator` borrows `self` mutably, so you can't also send in a reference to `self.context_separator`. There are a few ways to solve this, but the simplest is to do what the old code did: duplicate the code and inline the call to `separator` manually.

src/args.rs
BurntSushi: More features means more code that I need to maintain. For example, if we ever added code to auto-detect the best color settings (which seems like something Windows users might appreciate), then the more color settings we have, the harder that code will be to write.

Features need to be decided upon carefully. Once they're added, we can essentially never remove them. Sorry.

kpp:
See how it looks without colors for column and separator:

Image of example

It is easy to mix up column `:22:` with the text itself.

BurntSushi:
I'm not debating whether colors makes things easier to read. ripgrep has colors already, so I obviously see the utility. What I'm trying to ascertain is whether it's worth it to add more colors. I'm trying to do that by asking for use cases for reading column numbers. I don't understand why a human would find a column number to be useful information on a regular basis. Please see the discussion in #377.

kpp:
@BurntSushi +regressions

kpp:
Ah. I got it. The case of reading columns is `-o` flag.

Image of example

There is only one possibility to differ on bar from another bar if they are both on the same line.

kpp:
@BurntSushi is there any unresolved question about this PR?

BurntSushi:
Looks good, thank you!

Issue #362, opened 2017-02-13
Compiled regex exceeds size limit of 10485760 bytes.


GenomicaMicrob:
Hi, I am trying to use a file with more then 8,000 entries (10-20 letter words, one per line. 132K) and get their corresponding lines in a big file (645,151lines, 76M). I use:
`rg -w -f query_file target_file`

I get the error: `Compiled regex exceeds size limit of 10485760 bytes.`
How can I configure it to allow rg to run without the limit?

BurntSushi:
You can't, as of now, without re-compiling and hard-coding a new limit.

I would be interested to hear how well 8,000 entries performs. As of now, it is fed into a single regex as a single alternation. If they are all plain text strings and not patterns, then it might be better to use Aho-Corasick instead (which ripgrep should do, it just doesn't yet).

BurntSushi:
If you were so inclined, the limit is here: https://github.com/BurntSushi/ripgrep/blob/master/grep/src/search.rs#L67

FYI, you can't "remove" the limit (but you can set it arbitrarily high).

GenomicaMicrob:
Thanks, I have run it with a 640-line file (11K) without problems and very fast. The lines of the file are like these:
```
AACY020304192.1.1216
AACY020311045.179.1696
AACY020524983.300.1803
AACY020558197.3281.4799
AACY023705044.289.1515
AACY023835985.1.1221
```

I am fairly new to coding, so excuse me but I need a "for dummies" howto set the limit very high. Basically, what should I do and where?
I have a server with 64 cores, 128 Mb RAM.
sorry

BurntSushi:
The limit probably *should* be exposed as a flag so that it's a knob you can turn easily. That feature is not currently available, so the only way for you to do it is change the source code of ripgrep and recompile it. Briefly:

```
$ git clone git://github.com/BurntSushi/ripgrep
$ cd ripgrep
... try to compile it to make sure your env is setup right
$ cargo build --release
... change the limit in the source code
$ $EDITOR grep/src/search.rs
... compile again
$ cargo build --release
... the ripgrep binary is in target/release
$ ./target/release/rg blah blah blah
```

In order for the above to work, you will need to install Rust. Checkout: https://rustup.rs/

GenomicaMicrob:
OK, solved, thanks.
I ran a file with 11,724 lines in under 1:20 h, not bad compared to several hours with grep.

I changed the "size_limit" from 10 to 1000:

```
[...]

PR #433, opened 2017-04-01
Add dfa-size-limit and regex-size-limit arguments

tiehuis:
See #362.

 - The macro is not really to my taste but I wanted to see if you were happy with it first or had any ideas otherwise. I did try creating a generic function with the required trait bounds but the generic multiplication was the stopping point here for me. Also, if using a `saturating_mul` this is further pronounced since there is currently no trait in std which these are bound to. The `Saturating` trait on the [num](https://github.com/rust-num/num) also only applies for addition to.

- Removed the straight-forward multiplications and replaced with saturating versions. This avoids the potential overflow. Added a test case for this also to ensure it is handled properly. Since these arguments are always some memory bound (as per the suffix) a saturating limit makes the most sense. That being said, it is very (very) unlikely for the current width types.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -129,6 +129,8 @@ fn app<F>(next_line_help: bool, doc: F) -> App<'static, 'static>
         .arg(flag("column"))
         .arg(flag("context-separator")
              .value_name("SEPARATOR").takes_value(true))
+        .arg(flag("dfa-size-limit")
+             .value_name("NUM+SUFFIX?").takes_value(true))
         .arg(flag("debug"))
         .arg(flag("file").short("f")
              .value_name("FILE").takes_value(true)
@@ -162,6 +164,8 @@ fn app<F>(next_line_help: bool, doc: F) -> App<'static, 'static>
         .arg(flag("path-separator").value_name("SEPARATOR").takes_value(true))
         .arg(flag("pretty").short("p"))
         .arg(flag("replace").short("r").value_name("ARG").takes_value(true))
+        .arg(flag("regex-size-limit")
+             .value_name("NUM+SUFFIX?").takes_value(true))
         .arg(flag("case-sensitive").short("s"))
         .arg(flag("smart-case").short("S"))
         .arg(flag("sort-files"))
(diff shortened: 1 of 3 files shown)


src/args.rs
```diff
@@ -312,6 +312,48 @@ impl Args {
...
 }
 
+// This macro is used to generate a member function of `ArgMatches` which
+// parses an argument item of the form `[0-9]*(KMG)?`.
+//
+// This is a macro instead of a generic function due to difficulties in
+// correctly constraining the output $ty to work for a number of integers.
+// Since there are limited numbers of types that currently need this function
+// we perform the type monomorphization ourselves.
+macro_rules! human_readable_size_arg_impl {
```
kpp: You can split this macro into 2 functions: validator and value converter. See [an example of validator](https://github.com/kbknapp/clap-validators/blob/master/src/num/mod.rs)

src/app.rs
```diff
@@ -340,6 +344,13 @@ lazy_static! {
         doc!(h, "debug",
              "Show debug messages.",
              "Show debug messages. Please use this when filing a bug report.");
+        doc!(h, "dfa-size-limit",
+             "The upper size limit of the generated dfa.",
+             "The upper size limit of the generated dfa. The default limit is \
+              10Mb. This should only be changed on very large regex inputs \
```
BurntSushi: I think this should say `10M` instead of `10Mb`. (Even so, the `Mb` suffix means "megabit", which is otherwise wrong. :-))

src/app.rs
```diff
@@ -454,6 +465,11 @@ lazy_static! {
               Note that the replacement by default replaces each match, and \
               NOT the entire line. To replace the entire line, you should \
               match the entire line.");
+        doc!(h, "regex-size-limit",
+             "The upper size limit of the compiled regex.",
+             "The upper size limit of the compiled regex. The default limit \
+              is 10Mb. \n\nThe argument accepts the same size suffixes as \
```
BurntSushi: Same here. `10M` instead of `10Mb`.

src/args.rs
BurntSushi: I'm not sure this macro is necessary, and I'd really like to find a way to get rid of it. Can it be written to specifically return a `u64` and then just convert the result to a `usize`? (That conversion should return an error if the `u64` is greater than [`std::usize::MAX`](https://doc.rust-lang.org/std/usize/constant.MAX.html).)

BurntSushi:
Could you also change the commit message to say, `Fixes #362` so that github properly connects the PR with the issue? Thanks.

tiehuis:
Made fixes as per your suggestions and squashed into a single commit.

I've also added overflow detection instead of the saturation as you argued. Figured that it was better to get it as intended right now rather than leaving it for later.

Added an extra test specifically for 32-bit systems for when `usize < u64`.

BurntSushi:
@tiehuis Thanks again for working on this! I merged this PR in https://github.com/BurntSushi/ripgrep/commit/66efbad871620fe2dbe675438fdc8d9922e72826. I ended up making a few cosmetic changes to make the code more consistent with the style of the rest of the code. Nice work! :-)

kpp:
@BurntSushi stop closing pull requests. There is a button to merge the code with any message you like.

BurntSushi:
@kpp If you have a problem with the way I'm running the project and can suggest a better alternative, then please file an issue.

PR #437, opened 2017-04-04
fix typo in comment

kosta:
See, some people actually read comments ;)


BurntSushi:
Right, thanks! This was a result of my line wrapping going rogue. :-)

kosta:
:)

PR #438, opened 2017-04-05
updates clap and removes home rolled -h/--help distinction

kbknapp:
~~**DO NOT MERGE - until clap v2.23.1 is on crates.io**~~

It's safe to merge now.

- [x] crates.io needs to contain clap v2.23.1
- [x] I need to rebase the `Cargo.lock`

This commit updates clap to v2.23.1

The update contained a bug fix in clap that results in broken code in
ripgrep. ripgrep was unknowingly relying on the bug, but this commit fixes that
issue. The bug centered around not being able to override the
auto-generated help message by supplying a flag with a long of `help`.

Normally, supplying a flag with a long of `help` means whenever the user
passes `--help`, the consuming code (e.g. ripgrep) is responsible for
displaying the help message. However, due to the bug in clap this wasn't
necessary for ripgrep to do unless the user passed `-h`. With the bug
fixed, it meant the user passing `--help` and clap expected ripgrep to
display the help, yet ripgrep expected clap to display the help. This
has been fixed in this commit of ripgrep.

All is well now!

v2.23.0 also brings the abilty to use `Arg::help` or `Arg::long_help`
allowing one to distinguish between `-h` and `--help`. This commit
leaves all doc strings in the `lazy_static!` hashmap however only for
aesthetic reasons.

This means all home rolled handling of `-h`/`--help` has been removed
from ripgrep, yet functionality *and* appearances are 100% the same.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,7 +4,7 @@ version = "0.5.0"
 dependencies = [
  "atty 0.2.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "bytecount 0.1.6 (registry+https://github.com/rust-lang/crates.io-index)",
- "clap 2.22.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "clap 2.23.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "encoding_rs 0.5.0 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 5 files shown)


BurntSushi:
*Excellent!* Thanks so much for doing this. I much appreciate the simpler case analysis. :-)

At some point, I'll get rid of the `USAGES` map and inline the docs, but I like that this PR doesn't do that. :-)

Issue #152, opened 2016-10-05
preserve order of input files in output without sacrificing parallelism


jikamens:
There should be a way to tell rg to preserve the order of input files specified on the command line in the output.

E.g., if I'm searching a bunch of files which are in chronological order, I want the output to be in chronological order.

BurntSushi:
You can do this today, but you have to give up parallelism. If you run `rg -j1 PATTERN FILE FILE ...` then the output should be deterministic and in the order specified. In general, when recursively searching a directory, the order is whatever the file system serves up, but with `-j1` it should at least be deterministic.

jikamens:
Yeah, so I don't _want_ to give up parallelism, obviously.

[GNU parallel](https://www.gnu.org/software/parallel/) let's me be parallel and preserve the order of output with `--keep`. I'm proposing that rg should provide similar functionality. Otherwise `rg -j#` isn't a drop-in replacement for grepping a bunch of files with GNU parallel, which I would really like it to be. ;-)

BurntSushi:
I get what you're saying. Here are my thoughts on the matter:

1. If deterministic ordering + parallelism were possible to do without incurring additional costs, `rg` would do that by default. (Because I'd really like deterministic ordering too.) Perhaps I am wrong and this *is* possible (or perhaps, possible with immeasurable costs), and if so, we should do that.
2. If not (1), then hiding this behind another flag seems like bad UX, but maybe that's the best we can do.
3. Requiring single threaded searches for when you absolutely need deterministic output doesn't seem unreasonable to me.

All of these things combined together means I'm not particularly motivated to work on this, but I grant it's something I'd willing to explore given unbounded resources.

jikamens:
You can implement ordering + parallelism with no additional cost; all rg has to do is read output from the threads in the order that they are launched. This means, of course, that some of the threads will block waiting to be read because their output buffers will fill up, but the performance of such an implementation would be no _worse_ than `-j1`, and in most cases it would be significantly better.

You can improve the performance simply by making the output buffers larger so that they block less.

"You just have to put up with your search running eight times slower on your eight-core machine," is not really a reasonable answer, especially since I _don't_ have to put up with it, since as I pointed out, there are other tools that _will_ allow me to search and preserve ordering in the output.

[...]

PR #450, opened 2017-04-16
walk.rs: queue -> stack

bmalehorn:
Change the pool of "files to search" from queue to a stack. This causes
ripgrep to approximate depth-first search instead of breadth-first
search. This dramatically reduces the size of the pool, since most
directories are much more "wide" than they are "deep".

As a result, ripgrep uses ~45% less peak memory:

Before:

    /usr/bin/time -f '%M' rg PM_RESUME > /dev/null
    ...
    8876

After:

    /usr/bin/time -f '%M' rg PM_RESUME > /dev/null
    16240

Throughput improves a tiny bit:

Before:

    linux_literal (pattern: PM_RESUME)
    ----------------------------------
    rg (ignore)*  0.376 +/- 0.003 (lines: 16)*

After:

    linux_literal (pattern: PM_RESUME)
    ----------------------------------
    rg (ignore)*  0.371 +/- 0.004 (lines: 16)*

Related to #152.

### DIFF
--- a/ignore/src/walk.rs
+++ b/ignore/src/walk.rs
@@ -10,7 +10,7 @@ use std::thread;
 use std::time::Duration;
 use std::vec;
 
-use crossbeam::sync::MsQueue;
+use crossbeam::sync::TreiberStack;
 use walkdir::{self, WalkDir, WalkDirIterator, is_same_file};
 
 use dir::{Ignore, IgnoreBuilder};
@@ -864,7 +864,7 @@ impl WalkParallel {
     ) where F: FnMut() -> Box<FnMut(Result<DirEntry, Error>) -> WalkState + Send + 'static> {
         let mut f = mkf();
         let threads = self.threads();
-        let queue = Arc::new(MsQueue::new());
+        let stack = Arc::new(TreiberStack::new());
         let mut any_work = false;
         // Send the initial set of root paths to the pool of workers.
         // Note that we only send directories. For files, we send to them the
@@ -884,7 +884,7 @@ impl WalkParallel {
                         }
                     }
                 };
-            queue.push(Message::Work(Work {
+            stack.push(Message::Work(Work {
                 dent: dent,
                 ignore: self.ig_root.clone(),
             }));
@@ -902,7 +902,7 @@ impl WalkParallel {
         for _ in 0..threads {
             let worker = Worker {
                 f: mkf(),
-                queue: queue.clone(),
+                stack: stack.clone(),
                 quit_now: quit_now.clone(),
                 is_waiting: false,
                 is_quitting: false,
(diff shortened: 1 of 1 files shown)


BurntSushi:
OK, so I finally had a chance to manually QA this. If I run `rg PM_SUSPEND` in a checkout of the Linux kernel, then the results appear much more consistent from run-to-run in current master than they do with this PR. Is this something you expected to happen?

bmalehorn:
Great, thanks for taking a look!

I've tested it on my end and can also confirm that master is more consistent. I don't really have a great explanation; it's rather tricky to diff two types of undefined behaviors.

I guess this PR is a bit open-ended: how do you defined consistency? Master is more likely to output in the same order from one run to another. But "stack" will usually group directories together. If you only run ripgrep once, which output is better?

Master:

    include/linux/ide.h
    Documentation/dev-tools/sparse.rst
    drivers/ide/ide-pm.c
    include/uapi/linux/apm_bios.h
    arch/x86/kernel/apm_32.c
    Documentation/translations/zh_CN/sparse.txt
    drivers/input/mouse/cyapa.h
    drivers/usb/mtu3/mtu3_hw_regs.h
    drivers/mtd/maps/pcmciamtd.c
    drivers/net/wireless/intersil/hostap/hostap_cs.c

Stack:

    drivers/mtd/maps/pcmciamtd.c
    drivers/ide/ide-pm.c
    drivers/net/wireless/intersil/hostap/hostap_cs.c
    drivers/usb/mtu3/mtu3_hw_regs.h
    drivers/input/mouse/cyapa.h
    Documentation/translations/zh_CN/sparse.txt
    Documentation/dev-tools/sparse.rst
    arch/x86/kernel/apm_32.c
    include/uapi/linux/apm_bios.h
    include/linux/ide.h

BurntSushi:
Right, if you qualify it with "which is better if you only run ripgrep once," then I agree that this PR is better. But I'm not sure that I buy that qualification...

hlieberman:
I did a little bit of a back-of-the-envelope test here on how different these two actually are.

I ran 1,000 greps of PM_SUSPEND of Linux through current HEAD (45e850aff769a9d3f1fc6457bd5e1390021f433c) and through the PR branch here (d4fa0155e7e2fb68de133c0cd36f0be2f92e782f), both with SIMD/AVX optimizations, on a machine with 4 CPU cores.

HEAD showed very little overlap; of the 1000 entries, there are 998 unique md5sums. The PR branch shows no overlap at all; all 1000 entries are unique. As you might expect, there is also no overlap between the two sets.

To get a rough measure of "compactness", I looked at the number of times that top level directories were listed not in a row.  (that is: "a/foo, a/bar, b/foo/, a/bar" would have a score of 3)

Master:
```
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
  13.00   13.00   15.00   15.66   16.25   28.00
```
master

PR:
```
   Min. 1st Qu.  Median    Mean 3rd Qu.    Max. 
   7.00    9.00   11.00   14.21   21.00   35.00 
```
stack

So, the PR seems to produce more concise results on average, though the spread is a little worse.

Food for thought!

BurntSushi:
@hlieberman Thanks for that analysis! I think for now I'm going to close this PR. It's a neat idea, and we might want to revisit it, but I don't feel that there's compelling enough evidence to switch at this point.

PR #452, opened 2017-04-18
Add extended color support

### LINKED ISSUE #261: true colour support
Let us pick [any RGB colour](https://gist.github.com/XVilka/8346728) instead of just eight.

tiehuis:
Fixes #261.

 - Adds support for 256 and rgb color codes. Format is simply comma separated values.
 - Mixes fine with other styles such as bold/nobold. I did write a brief thing about this in the issue but doesn't seem necessary after writing the documentation.
 - Windows falls back to white in the presence of one of these new color codes.
 - This currently uses a fixed buffer for formatting the color codes instead of the `format!` macro. Its more complicated and I'm not entirely sure the benefit but it can always be changed back if the extra allocations are favored over the more complex code.

### DIFF
--- a/termcolor/src/lib.rs
+++ b/termcolor/src/lib.rs
@@ -954,6 +954,60 @@ impl<W: io::Write> Ansi<W> {
                 }
             }
         }
+        macro_rules! write_var_ansi_code {
+            ($pre:expr, $($code:expr),+) => {{
+                // The loop generates at worst a literal of the form
+                // '255,255,255m' which is 12-bytes.
+                // The largest `pre` expression we currently use is 7 bytes.
+                // This gives us the maximum of 19-bytes for our work buffer.
+                let pre_len = $pre.len();
+                assert!(pre_len <= 7);
+                let mut fmt = [0u8; 19];
+                fmt[..pre_len].copy_from_slice($pre);
+                let mut i = pre_len - 1;
+                $(
+                    let c1: u8 = ($code / 100) % 10;
+                    let c2: u8 = ($code / 10) % 10;
+                    let c3: u8 = $code % 10;
+                    let mut printed = false;
+
+                    if c1 != 0 {
+                        printed = true;
+                        i += 1;
+                        fmt[i] = b'0' + c1;
+                    }
+                    if c2 != 0 || printed {
+                        i += 1;
+                        fmt[i] = b'0' + c2;
+                    }
+                    // If we received a zero value we must still print a value.
+                    i += 1;
+                    fmt[i] = b'0' + c3;
+                    i += 1;
+                    fmt[i] = b';';
+                )+
+
+                fmt[i] = b'm';
+                self.write_all(&fmt[0..i+1])
+            }}
+        }
+        macro_rules! write_custom {
+            ($ansi256:expr) => {
+                if fg {
+                    write_var_ansi_code!(b"\x1B[38;5;", $ansi256)
+                } else {
+                    write_var_ansi_code!(b"\x1B[48;5;", $ansi256)
+                }
+            };
+
+            ($r:expr, $g:expr, $b:expr) => {{
+                if fg {
+                    write_var_ansi_code!(b"\x1B[38;2;", $r, $g, $b)
+                } else {
+                    write_var_ansi_code!(b"\x1B[48;2;", $r, $g, $b)
+                }
+            }};
+        }
         if intense {
             match *c {
                 Color::Black => write_intense!("8"),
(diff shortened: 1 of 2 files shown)


termcolor/src/lib.rs
```diff
@@ -954,6 +954,58 @@ impl<W: io::Write> Ansi<W> {
                 }
             }
         }
+        macro_rules! write_var_ansi_code {
```
tiehuis: Some eyes on this would be good if its preferred over `format!`. I've added some test cases here but could have missed something.

tiehuis:
@scottchiefbaker just highlighted a problem with the color parsing (no surprises there). I've made squashed the fix into this commit and made a quick test script which can be used to check if the parsing code matches visually.

https://gist.github.com/tiehuis/15f6a85c1b8fe691008c10b4d325a7c3

scottchiefbaker:
I can confirm that I've tested this branch and it does everything I wanted in this feature. Colors are accurate and simple to specify. This pull-request gets my 👍 for merge. Good work @tiehuis

scottchiefbaker:
@tiehuis I don't know enough (any) rust to speak the quality of this code, but I think the implementation and output are great.

What are your thoughts on this PR?

tiehuis:
I've added my notes about this in the initial post. It's up to BurntSushi whether it gets merged or not and if any changes need to be made.

I know he has a lot of other things to tend to as well so I usually don't want to be too overbearing. He does a great job at managing his various projects so just give him a bit of time and this'll get looked at eventually.

scottchiefbaker:
@BurntSushi any thoughts on this PR? It's working pretty well for me.

leafgarland:
I just wanted to add that the Windows 10 console has supported 256 ansi colours for a while and since Creators Update supports 24 bit RGB.
https://msdn.microsoft.com/en-us/library/windows/desktop/mt638032(v=vs.85).aspx

Other commonly used terminal emulators such as mintty and ConEmu also support 24 bit colours.

Since ripgrep already has a flag for ansi colours, please don't just disable by default on Windows. Windows users can opt in/out depending on what they are using. As an aside, ripgrep is roughly twice as fast on windows with `--color=ansi`.

BurntSushi:
> Since ripgrep already has a flag for ansi colours, please don't just disable by default on Windows.

@leafgarland Can you please elaborate? What does your comment have to do specifically with this PR?

BurntSushi:
@tiehuis Sorry I've been dragging my feet looking at this! Thanks so much for doing it! At first glance, this looks pretty good but I'll want to do a more thorough review before merging.

leafgarland:
@BurntSushi I was commenting because the description mentioned that this PR was defaulting to white on Windows.

tiehuis:
Can't have a long conversation right now but just had a quick look and I think this should handle ansi codes correctly on windows if `--color=ansi` is specified.

The only thing that should be updated potentially is the `should_ansi` function in termcolor which determines if the `AnsiWriter` is used instead of the Windows one. Not sure if this does a thorough enough check for a Windows 10 terminal. I think the description in the initial message would be better described as "Fall back to white on windows where ANSI is not supported (i.e. cmd.exe)".

Haven't tested at all on Windows however, so wanted a minimal change incorporating the functionality first. It should be simple to add support for Windows later if needed with no real trouble.

BurntSushi:
@leafgarland That's only when printing using the console APIs. It looks like this PR doesn't do anything special with the ANSI codes when those are used on Windows.

leafgarland:
Ok, thanks. Sorry, I should have looked at the code!

BurntSushi:
> The only thing that should be updated potentially is the should_ansi function in termcolor which determines if the AnsiWriter is used instead of the Windows one. Not sure if this does a thorough enough check for a Windows 10 terminal

Yeah, this is potentially future work. Def out of scope for this PR. :-)

weiznich:
What is the current state of this pull request? What needs to be done to get it merged?
(I would like to implement a feature based on this in cargo)

BurntSushi:
@weiznich It's waiting on me giving it a more careful review. I'll try to look soon.

scottchiefbaker:
@weiznich I looked at your PR for Rust, and I'm a little confused. What feature does that implement that's related to this? Or would it be a new feature for ripgrep you're waiting to implement?

weiznich:
@alexcrichton moved the coloring in cargo to termcolor with [this commit](https://github.com/rust-lang/cargo/commit/f8fb0a022853ab3f423889d442c695511ece3692). After that it is only possible to use the 8 predefined color codes. This may result in a colors that are barley readable(See my [old PR](https://github.com/rust-lang/cargo/pull/3873) for an example).
This PR introduces two possibilities to define custom colors. Based on this it should be possible to allow configuring the colors through an environment variable as proposed in my PR to cargo.

scottchiefbaker:
@BurntSushi have you had a chance to review this at all?

scottchiefbaker:
@BurntSushi any movement on this? I'd really like to see better colors in ripgrep.

BurntSushi:
Thanks @tiehuis for this! I'm going to close this PR in favor of #766, which is based on your work here. I wanted to just get this merged so I did a few fixups.

PR #455, opened 2017-04-20
Add snapcraft.yaml

ChrisMacNaughton:
[Snapcraft](https://snapcraft.io/) makes Linux packaging very simple in a cross-distro
way. This adds the snapcraft.yaml file to setup a snap of ripgrep

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,11 @@ target
 /ignore/Cargo.lock
 /termcolor/Cargo.lock
 /wincolor/Cargo.lock
+
+# Snapcraft files
+stage
+prime
+parts
+*.snap
+*.pyc
+ripgrep*_source.tar.bz2
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -1,0 +1,15 @@
+name: ripgrep # you probably want to 'snapcraft register <name>'
+version: '0.5.1' # just for humans, typically '1.2+git' or '1.3.2'
+summary: Fast file searcher # 79 char long summary
+description: |
+  ripgrep combines the usability of The Silver Searcher with the raw speed of grep.
+grade: stable # must be 'stable' to release into candidate/stable channels
+confinement: classic # use 'strict' once you have the right plugs and slots
+parts:
+  ripgrep:
+    plugin: rust
+    source: .
+apps:
+  rg:
+    command: env PATH=$SNAP/bin:$PATH rg
+    aliases: [rg]


ChrisMacNaughton:
I have already registered the snap package name ripgrep, will pass ownership of that over and am happy to help with setting up automated builds!

Phrohdoh:
Why snapcraft over flatpack?

I know next to nothing about both but there are multiple offerings available and we should probably weight the pros and cons before selecting one (I believe providing both defeats the purpose).

BurntSushi:
This isn't the place to debate snapcraft vs flatpak.

While I appreciate this PR, I know nothing about snapcraft/flatpak and I don't have any plans to use them in the immediate future, which makes it hard for me to maintain. Is it necessary for `snapcraft.yml` to be in this repo?

(I did make an exception for Brew since so many people are using it, although even that will eventually go away once SIMD works on stable Rust.)

ChrisMacNaughton:
I'd rather not discuss one versus the other as well, although a benefit to snaps is that they are available by default on any Ubuntu >= 16.04.

Regarding inclusion in the repository, it is possible to use snaps without including them in the repository but it is a bit easier to deploy into a testing channel with snaps from CI if it's in tree.

BurntSushi:
All right. I'll merge this for now and we can see how it goes. Thanks!

ChrisMacNaughton:
@BurntSushi Feel free to ping me (or the snapcraft team) if things are needed to help maintaining it :)

BurntSushi:
@ChrisMacNaughton I have a question about this if you don't mind. I haven't updated ripgrep's `snapcraft.yaml` file since I merged this PR. When I went to install ripgrep via snap on my Ubuntu 17.10 machine, I was surprised to see that it install ripgrep 0.7.1, which is the latest release. How does that work?

ChrisMacNaughton:
@BurntSushi looks like somebody is maintaining a build:
```
$ snap search ripgrep
Name       Version  Developer  Notes  Summary
rg-casept  0.5.2    casept     -      ripgrep combines the usability of ag with the raw speed of grep.
rg         0.7.1    tasdomas   -      a command line search tool
```
To me, the best solution would be to actually setup ripgrep to run on the launchpad builders automatically by setting it up at https://build.snapcraft.io

BurntSushi:
@ChrisMacNaughton I guess I'm just *super* confused:

1. The version in ripgrep's `snapcraft.yaml` file is `0.5.1`, which doesn't match either of the versions in your search results.
2. The `snapcraft.yaml` file does not seem to contain any instructions on how to actually build ripgrep. How is it working at all?
3. Earlier in this PR you mentioned that you already registered the name `ripgrep`, but the two ripgrep related packages in snap appear to be `rg-casept` and `rg`.

Maybe the answer is that I need to spend more time in the docs?

ChrisMacNaughton:
1. Anybody can create a snapcraft.yaml, and manage the building and publishing it; ideally upstreams (such as ripgrep, here) will publish their own snapcraft.yaml, and can have https://build.snapcraft.io automatically build on push
2. The magic for snapcraft to know how to build ripgrep is contained[here](https://github.com/BurntSushi/ripgrep/blob/master/snapcraft.yaml#L10), it uses snapcraft's [Rust plugin](https://docs.snapcraft.io/reference/plugins/rust)
3. As a corollary to 1, anybody can register a name and upload anything to it, I currently have ripgrep (would rather hand control of that over to you, frankly) which means that the others are named more interestingly :-)

If you'd like a more real-time discussion, I am on IRC as well as icey

BurntSushi:
@ChrisMacNaughton Awesome, OK! I'm at work, so not on IRC, but I'd be happy to take over the `ripgrep` name and get this setup. I'm not sure what that process looks like though!

Issue #471, opened 2017-05-02
Encoding issue when piping rg output


elirnm:
There seems to be an encoding issue when piping the output of a ripgrep search. The characters `ø æ å ä` (and probably others, those are just the ones I've tested with) are converted to `??` when being piped to anything.

It seems to only be with piping; passing stdout to a file with `>` works fine (at least in cmd.exe where files are created with utf8).

![capture](https://cloud.githubusercontent.com/assets/19817004/25610476/32325c72-2ed8-11e7-8c12-443fb138567f.PNG)

This is what I used to test this.
```
mkdir rgtest
cd rgtest
echo "testøæåä" > strtest.txt
mkdir testø
echo "blahblah" > testø/testå.txt
```

Then search for various things as seen in the picture. (The `--colors` option is unrelated; that's just so you can see the filenames in Powershell).

This is on Windows 7 with ripgrep 0.5.1. I've recreated this in both powershell and cmd.exe, but I don't have a Linux install to test it there.

BurntSushi:
Can you provide the raw bytes of strtest.txt?

elirnm:
```
116
101
115
116
195
184
195
166
195
165
195
164
13
10
```

If that's not what you wanted you'll have to give me more specific instructions.

The file is currently UTF-8 encoded but this issue occurs with other file encodings as well (and when not reading from a file, as seen in the `--files` examples.

elirnm:
Hmm, this might not actually be ripgrep. Piping these characters is succeeding with every native Powershell command, but failing with every call to an executable or any non-built-in program.

![cap](https://cloud.githubusercontent.com/assets/19817004/25614555/4b3a87bc-2ee7-11e7-882e-965428a56dec.PNG)

And while I seem to remember it failing in cmd.exe before, it works fine there now.

Issue #445, opened 2017-04-11
Error with Powershell tab completion script


elirnm:
Attempting to import the _rg.ps1 tab completion script into Powershell throws an exception:

```
At C:\PathTools\_rg.ps1:10 char:41
+                 switch ($_.ToString()) {
+                                         ~
Missing condition in switch statement clause.
    + CategoryInfo          : ParserError: (:) [], ParseException
    + FullyQualifiedErrorId : MissingSwitchConditionExpression
```

There's an empty switch statement there that seems like the cause.

$PSVersionTable:
```
Name                           Value
----                           -----
PSVersion                      5.1.14409.1005
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14409.1005
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

```

BurntSushi:
This is not something I'm capable of diagnosing and fixing. Someone else will need to do it.

elirnm:
It looks like the file is generated by clap? If that's the case then I can create an issue over there.

BurntSushi:
@elirnm Indeed it is! Thanks. :-)

elirnm:
This is fixed in the latest clap.

elirnm:
This seems to work now in 0.5.2.

PR #472, opened 2017-05-04
Add Powershell tips and autocompletion instructions

elirnm:
Updated the readme with tips on piping non-ASCII content (the cause/fix for https://github.com/BurntSushi/ripgrep/issues/471) and on resetting colors after `ctrl+c` in Powershell (because `color` isn't supported), plus instructions for using the Powershell autocompletion file (which is dependent on ripgrep updating to clap 2.23.2, see https://github.com/BurntSushi/ripgrep/issues/445 and https://github.com/kbknapp/clap-rs/issues/931).

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,6 +343,11 @@ or `/etc/bash_completion.d/`.
 
 For **fish**, move `rg.fish` to `$HOME/.config/fish/completions`.
 
+For **PowerShell**, add `. _rg.ps1` to your PowerShell
+[profile](https://technet.microsoft.com/en-us/library/bb613488(v=vs.85).aspx)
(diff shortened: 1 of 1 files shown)


BurntSushi:
@elirnm This looks great, thanks so much for writing it up! Would you mind formatting the additions in the same style as the rest of the README? I think in this case that's making sure the text wraps at 79 columns. And then please squash everything down to one commit. (I can do this for you if you like when I get around to it.)

elirnm:
@BurntSushi I formatted and squashed the changes.

BurntSushi:
@elirnm Thanks so much!

Issue #6, opened 2016-09-16
add support for mercurial


BurntSushi:
Mercurial is widely used enough that we should probably support it. Mercurial will actually be harder to do correctly than git, because an `.hgignore` file can support both regexes and globs. An `.hgignore` file can also specify _subincludes_, which include ignore patterns in a sub-directory (as opposed to `git`, which will read `.gitignore` files in sub-directories automatically).

Thankfully, `ripgrep` translates all globs to regexes, so we should be able to support Mercurial without too much trouble.

See: `hg help hgignore` and `hg help patterns`.

jFransham:
I'll see if I can do some work on this

BurntSushi:
If someone wanted to work on this, I'd be happy to mentor it. I think the code in this portion of ripgrep is isolated well enough that someone could truly focus on just the Mercurial ignore files and not too much else. I would be OK with this being done in multiple PRs. I would also be OK with partial support (e.g., "only `.hgignore` files with globs are supported") so long as there's a clear path to implementing complete support.

The first step would be to add a new `hgignore` submodule to the [`ignore` crate](https://docs.rs/ignore). It should roughly mirror the existing [`gitignore`](https://docs.rs/ignore/0.1.8/ignore/gitignore/index.html) submodule, although `hgignore` itself operates slightly differently than `gitignore`, and that could impact the API.

kalekseev:
@BurntSushi I would love to have the mercurial support in the ripgrep, yesterday I compiled a version that ignore globs from a projects .hgignore https://github.com/kalekseev/ripgrep. For now it's mostly copy-paste of gitignore code. I don't have any rust experience so not sure how far I can get with the task but I'm going to try to add support for hgignore specific features in the next couple of weeks and then create a PR.

BurntSushi:
@kalekseev Mercurial support would be great. Do note that both regexes and globs are supported, and I'd expect there to be plenty of subtle differences between gitignore and hgignore. So I think there needs to be lots of tests. If Mercurial has a test suite we can borrow from, that'd be great. :-)

theor:
apparently this is the whole hgignore test suite : https://www.mercurial-scm.org/repo/hg/file/tip/tests/test-hgignore.t

kalekseev:
@theor thanks that will help

tamird:
@BurntSushi I'd like to pick this up where #477 left off. Do you have any mentoring notes?

kalekseev:
[...]

PR #477, opened 2017-05-09
[WIP] Support for mercurial

kalekseev:
Support for mercurial https://github.com/BurntSushi/ripgrep/issues/6
Very basic support is implemented for now, although it covers my own usecases:
  - patterns from repo's `.hgignore` and from `ignore` control in ui part of `$HOME/.hgrc`
  - `syntax: glob` rules (mostly using git logic)
  - `syntax: regexp` rules (they are directly translated into `Glob::re`)

I'm planning to translate mercurial's tests suit for hgignore files as my next step.

@BurntSushi, I want to start a discussion about support for `subinclude` directives:

    subinclude:path/to/subignorefile reads patterns specifically for paths in the subdirectory

To support this the process of building Ignore object has to depend on the information from .hgignore file near the .hg directory. Therefore we have to guaranty that dirs under mercurial are processed after the parsing of a repo's .hgignore (and all the `subinclude`, `include` files). In the parallel run that looks hard until we lock all the worker's threads (hardly the option).

The other they may be to update retroactively all Ignore objects with the content of the related subignore file.

### DIFF
--- a/globset/src/glob.rs
+++ b/globset/src/glob.rs
@@ -545,7 +545,7 @@ impl<'a> GlobBuilder<'a> {
 
     /// Parses and builds the pattern.
     pub fn build(&self) -> Result<Glob, Error> {
-        if (self.opts.from_regex) {
+        if self.opts.from_regex {
             Ok(Glob {
                 glob: self.glob.to_string(),
                 re: self.glob.to_string(),
--- a/globset/src/glob.rs
+++ b/globset/src/glob.rs
@@ -194,6 +194,8 @@ struct GlobOptions {
     /// Whether to require a literal separator to match a separator in a file
     /// path. e.g., when enabled, `*` won't match `/`.
     literal_separator: bool,
+    /// doc
+    from_regex: bool,
 }
 
 #[derive(Clone, Debug, Default, Eq, PartialEq)]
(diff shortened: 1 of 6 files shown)


globset/src/glob.rs
```diff
@@ -194,6 +194,8 @@ struct GlobOptions {
     /// Whether to require a literal separator to match a separator in a file
     /// path. e.g., when enabled, `*` won't match `/`.
     literal_separator: bool,
+    /// doc
+    from_regex: bool,
```
BurntSushi: I'm not sure this is quite the right way to go about this. In particular, you wind up storing a regex in a field called `glob`, which doesn't seem good to me. Additionally, the `from_regex` option isn't really compatible with the `case_insensitive` or `literal_separator` flags, which also makes the builder strange. I think I'd rather see `Glob` itself be an enum and support a `from_regex` constructor. In that approach, `GlobBuilder` doesn't need to be modified at all.

ignore/src/hgignore.rs
```diff
@@ -0,0 +1,504 @@
...
+    let caps = match RE.captures(data) {
+        None => return None,
+        Some(caps) => caps,
+    };
+    str::from_utf8(&caps[1]).ok().map(|s| PathBuf::from(expand_tilde(s)))
+}
+
+/// Expands ~ in file paths to the value of $HOME.
+fn expand_tilde(path: &str) -> String {
+    let home = match env::var("HOME") {
```
BurntSushi: I think you actually want [`env::home_dir`](https://doc.rust-lang.org/std/env/fn.home_dir.html) instead.

ignore/src/hgignore.rs
```diff
@@ -0,0 +1,504 @@
...
+                        err: err.kind().to_string(),
+                    }
+                }));
+        self.builder.add(parsed);
+        self.globs.push(glob);
+        Ok(self)
+    }
+
+    /// doc
+    pub fn add_glob_line(
```
BurntSushi: There's a lot in common with this method and a similar method in `GitignoreBuilder`. Are the semantics really this identical? And if so, can we split the logic out into a function that both builders can use?

globset/src/glob.rs
kalekseev: > I'd rather see Glob itself be an enum

Could you provide a draft definition for that enum that will suit the project best?
I've tried to replace the Glob structure with the
```
pub enum Pattern {
    Glob { glob: String, re: String, opts: GlobOptions, tokens: Tokens },
    Regex(String),
}
```
but it seems that enums don't have constructors in rust so the private GlobOptions and Tokens types leak.

BurntSushi: Ah, right, yeah you'll have to do something like this:

```rust
pub struct Glob(GlobInner);

enum GlobInner {
    Glob { ... },
    Regex(String),
}
```

The other interesting bit here is that by offering a constructor that accepts a regex, we've essentially tied to the implementation itself to regexes as well. It's mildly unfortunate, but I think it's the right path forward for now.

BurntSushi:
@kalekseev I'm trying to clean up the PR queue, so I think I'm going to close this for now due to inactivity, but please feel free to re-open another PR if you want to continue to attack this!

PR #486, opened 2017-05-24
Add support for QMake files

Manozco:
Here is a PR for supporting QMake (from Qt) files :)


BurntSushi:
Fantastic, thank you!

PR #487, opened 2017-05-24
Fix typo, should be 'mode' instead of 'more'

ericbn:
in man pages.


BurntSushi:
Nice find! Thanks! :-)

PR #488, opened 2017-05-25
Should show filename for one file with vimgrep

ericbn:
With vim configured with:

    set grepprg=rg\ --vimgrep
    set grepformat^=%f:%l:%c:%m

and running the command `:grep 'vimgrep' doc/rg.1`, the output should be:

    doc/rg.1:446:8:.B \-\-vimgrep

but the actual output was:

    446:8:.B \-\-vimgrep

Same issue would happen if results only match one file. Ag behaves as expected.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -532,6 +532,7 @@ impl<'a> ArgMatches<'a> {
             false
         } else {
             self.is_present("with-filename")
+            || self.is_present("vimgrep")
             || paths.len() > 1
             || paths.get(0).map_or(false, |p| p.is_dir())
         }


BurntSushi:
Hmm, I think my original intention was that you'd want to pass `--with-filename` for this, but I'm guessing that the output of `--vimgrep` is completely useless if there's no file name, right?

ericbn:
The grep command for vim should be only configured as `set grepprg=rg\ --vimgrep`, right?

Now, the default grep format in vim is (see `:help grepformat`) `"%f:%l:%m,%f:%l%m,%f  %l%m"`,  meaning `<filename>:<linenumber>:<text>` or the variations that follow. Both `%f` and `%l` are the minimal elements required for a useful output.

Ag introduced the column number on it's `--vimgrep` output, requiring the extra `set grepformat^=%f:%l:%c:%m` configuration (the `^=` prepends to the existing default comma-separated list). Vim supports even more configurable elements (see `:help errorformat`), and having  the column number better positions the cursor when browsing the results.

TL;DR yeah, the `--vimgrep` output is useless without the file names.

BurntSushi:
> The grep command for vim should be only configured as set grepprg=rg\ --vimgrep, right?

I don't know. I don't use ripgrep inside of vim.

But I buy it! Thanks!

PR #489, opened 2017-05-25
Update documentation for --color ansi

ericbn:
In `src/app.rs`, change typo "When ansi used" to "When ansi is used".
Update man pages with missing `ansi` option for `--color`.

### DIFF
--- a/doc/rg.1
+++ b/doc/rg.1
@@ -43,8 +43,8 @@ Only show count of line matches for each file.
 .RE
 .TP
 .B \-\-color \f[I]WHEN\f[]
-Whether to use coloring in match.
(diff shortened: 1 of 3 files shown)


ericbn:
One not-so-closely-related question: why isn't the `fn color(&self) -> bool` in `src/args.rs` testing for the "ansi" value? As fair as I got it, it would return false if preference == "ansi"...

BurntSushi:
Looks like that function is vestigial. If you look below it, there is a `color_choice` function that appears to be used now.

ericbn:
Yeah, it's not being used anymore. Created #490

PR #492, opened 2017-05-26
Update help and man pages

ericbn:
Formatting of rg.1.md. Removed backticks from already indented code.  Added missing italic to some arguments, Replaced -n by --line-number in --pretty for better clarity. Added explicit example of `*.foo` instead of `<glob>` in examples. Added vim information to --vimgrep.

In src/app.rs, also changed help text for pattern and regexp. Actually, "multiple patterns may be given" was not true for the standalone pattern.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -14,10 +14,10 @@ Project home page: https://github.com/BurntSushi/ripgrep
 Use -h for short descriptions and --help for more details.";
 
 const USAGE: &'static str = "
-    rg [OPTIONS] <pattern> [<path> ...]
-    rg [OPTIONS] [-e PATTERN | -f FILE ]... [<path> ...]
-    rg [OPTIONS] --files [<path> ...]
-    rg [OPTIONS] --type-list";
+    rg [options] PATTERN [path ...]
+    rg [options] [-e pattern ...] [-f file ...] [path ...]
+    rg [options] --files [path ...]
+    rg [options] --type-list";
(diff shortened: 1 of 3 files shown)


src/app.rs
```diff
@@ -14,10 +14,10 @@ Project home page: https://github.com/BurntSushi/ripgrep
 Use -h for short descriptions and --help for more details.";
 
 const USAGE: &'static str = "
-    rg [OPTIONS] <pattern> [<path> ...]
-    rg [OPTIONS] [-e PATTERN | -f FILE ]... [<path> ...]
-    rg [OPTIONS] --files [<path> ...]
-    rg [OPTIONS] --type-list";
+    rg [options] PATTERN [path ...]
+    rg [options] (-e PATTERN | -f FILE) [path ...]
```
BurntSushi: I think there should be `...` following `(-e PATTERN | -f FILE)`.

ericbn: Oh I see, because -e and -f can be defined multiple times... As both -e and -f can also be used simultaneously, what about:

    rg [options] [-e pattern ...] [-f file ...] [path ...]

?

BurntSushi: Sure, that's fine. Note that I would like to keep the `rg [options] PATTERN [path ...]` line. :-)

ericbn: Done! Updated the PR commit with the changes.

PR #494, opened 2017-05-29
Add julia as file type

gentbajraj:
This PR adds support for [julia](https://github.com/JuliaLang/julia) files


BurntSushi:
Sorry this took me so long to look at, I think I just plain missed this one. Do you want `julia` as the name or would you prefer `jl`? I don't have a strong preference but I would rather have one and not both.

gentbajraj:
I can't think of a strong rational for it since both versions seem descriptive but I prefer `julia`.

BurntSushi:
All right, let's so with it. If pitchforks visit my door, then I can reluctantly accept having both. (Once this gets released, I won't remove it.)

PR #496, opened 2017-05-29
Add custom Zsh completion file

MaskRay:
clap-rs [1] does not generate good Zsh completion files, thus a custom completion script should be better.
I have refined https://github.com/zsh-users/zsh-completions/pull/510 and added more options to `complete/_rg`.

[1]: https://github.com/kbknapp/clap-rs/issues/868

### DIFF
--- a/README.md
+++ b/README.md
@@ -338,15 +338,17 @@ The syntax supported is
 Shell completion files are included in the release tarball for Bash, Fish, Zsh
 and PowerShell.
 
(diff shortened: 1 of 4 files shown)


BurntSushi:
I'm OK with bringing this into the repo, so long as it's clear that I personally might not be able to keep it up to date.

MaskRay:
Thanks!

PR #498, opened 2017-06-01
Use uppercase for required argument names

ericbn:
This reverts a couple of changes introduced in 4c78ca8

@BurntSushi, not sure if you want to merge this. Just trying to revert some changes I introduced, plus keeping the `PATTERN` argument consistently uppercased, so error messages can look like

```
error: The following required arguments were not provided:
    <PATTERN>
```


BurntSushi:
LGTM. Thanks. :-) Consistency is good!

PR #505, opened 2017-06-06
Add a cshtml ignore type

### LINKED ISSUE #504: [request] More built-in types (globs)
I'd like to have (and possibly implement myself, I can't image this is too difficult):

```
$ rg -t cshtml
$ rg -t javascript
```

I say `javascript` because `cs` has a matching `csharp` but `js` is the only javascript option.

These could of course be added in a shell alias with `--type-add` but then I also need to alias in a `-tcshtml -tjavascript $@`.

Phrohdoh:
Closes #504


BurntSushi:
All righty! Thanks for your patience with helping me understand your request. :-)

PR #510, opened 2017-06-12
Add "msbuild" filetype grouping, and add .inl files to the "cpp" grouping. 

bgianfo:
This change consists of two commits:

- Add a new fietype grouping for "msbuild" related files.
- Add .inl files to be included in the cpp filetype grouping.

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -155,6 +155,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("matlab", &["*.m"]),
     ("mk", &["mkfile"]),
     ("ml", &["*.ml"]),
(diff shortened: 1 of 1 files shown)


elirnm:
`.fsproj` is also used by MSBuild, for F# projects.

BurntSushi:
TIL about `inl` files: https://stackoverflow.com/questions/1208028/significance-of-a-inl-file-in-c

@bgianfo This looks good. Maybe add `fsproj` to `msbuild` as @elirnm suggested?

bgianfo:
@BurntSushi, I pushed an update which adds .fsproj to the msbuild set as well.

Thanks!

BurntSushi:
@bgianfo Thanks! And thanks for amending the previous commit. :-)

PR #512, opened 2017-06-12
add yocto file types

syedelec:
add file types for embedded linux Yocto project : https://www.yoctoproject.org/

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -202,6 +202,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("xml", &["*.xml"]),
     ("yacc", &["*.y"]),
     ("yaml", &["*.yaml", "*.yml"]),
+    ("yocto", &["*.bb", "*.bbappend", "*.bbclass"]),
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks so much!

lilianmoraru:
I would like to point 2 issues:
- Yocto is too generic. These files are specific to 1 tool, called BitBake.
- This is not a complete list. It also has files akin' to headers in C/C++, where the content is inserted directly inside the *.bb* files.

syedelec:
@lilianmoraru, thank you for pointing out these remarks, what do you suggest ?
Actually, I was thinking using `-tbb` but it is too explicit and may create confusion since people may think it only includes `.bb` files.

lilianmoraru:
This would be in the same ballpark as `sh`, `cpp`, `c`, `asm`, etc...

lilianmoraru:
I scrolled through the reference manual and it seems to have these extensions:
```
.bb
.bbappend
.bbclass
.conf
.inc
```
It also can call into Python code, but I guess this could be excluded.

syedelec:
Multiple languages and configurations use `.conf` (e.g `rc.conf`, `syslog.conf` etc...) and `.inc` (e.g in C/C++, PHP and VIM config) as file type. IMHO, they cannot be included in specific yocto/bitbake file types. @BurntSushi, what do you think ?

BurntSushi:
I don't have a lot of domain knowledge here, but it is OK for file types to overlap. For example, `-tc` and `-tcpp` will both search `*.h` files.

PR #514, opened 2017-06-14
fix profile showing command for powershell

santagada:
change the profile showing command to one that matches the microsoft article linked, as its simpler and the old one didn't work at least on windows 10 creators edition.

### DIFF
--- a/README.md
+++ b/README.md
@@ -392,7 +392,7 @@ from the repository root.
 ##### Powershell Profile
 
 To customize powershell on start-up there is a special powershell script that has to be created.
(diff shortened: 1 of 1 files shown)


DoumanAsh:
Hmm... it is rather strange as this command works on my win10
But it might depend on powershell version...

Anyway i didn't know that you could print content of variable like that. It is definitely better!

BurntSushi:
@santagada Does this work on older versions of Windows?

elirnm:
@BurntSushi It works on my Windows 7, which is the oldest OS Powershell supports.

BurntSushi:
@elirnm Thanks for the confirmation!

Issue #506, opened 2017-06-07
Using an unparenthesized pipe with -w produces unexpected results


scottmcm:
I ran `rg -w 'min|max' src/lib*` in the rust source, and got far more results than I expected.  For example:
```
src/liblibc\src\unix\solaris\mod.rs
17:pub type minor_t = ::c_uint;
250:        pub f_namemax: ::c_ulong,
1070:    pub fn mincore(addr: *const ::c_void, len: ::size_t,
```

It seems to be matching `(\bmin)|(max\b)`, whereas I expected `\b(min|max)\b`.

repro'd in 0.4.0 and 0.5.2

BurntSushi:
Hah. Nice. Looks like you are right: https://github.com/BurntSushi/ripgrep/blob/master/src/args.rs#L508

BurntSushi:
This should be a relatively easy bug if anyone wants to take a crack at it. Basically, all that needs to be done is to put a non-capturing group around the inner expression here: https://github.com/BurntSushi/ripgrep/blob/13235b596f93f7f0e7ce76e7967996787147e96e/src/args.rs#L508 --- And put a regression test in `tests/tests.rs` (search for `regression` to see examples of other regression tests).

PR #516, opened 2017-06-15
fix word boundary w/ capture group

emattiza:
fixes BurntSushi/ripgrep#506. Word boundary search as arg had unexpected
behavior. added non-capturing capture group to regex to encapsulate 'or' option search and
prevent expansion and partial boundary finds.

Signed-off-by: Evan.Mattiza <emattiza@gmail.com>


BurntSushi:
Awesome, thanks so much! :-)

PR #520, opened 2017-06-17
Add -x/--line-regexp

okdana:
Let's try this!

This PR adds `-x`/`--line-regexp`, which essentially wraps the pattern in `^`...`$` in order to make it match whole lines only. This is an option defined in the [POSIX spec](http://pubs.opengroup.org/onlinepubs/009695399/utilities/grep.html) for `grep`, and one that i use a lot. I found it rather frustrating that `ag` didn't have it.

I will add additional notes as comments on the diff.

### DIFF
--- a/doc/convert-to-man
+++ b/doc/convert-to-man
@@ -1,4 +1,5 @@
-#!/bin/sh
+#!/bin/sh -e
 
 pandoc -s -t man rg.1.md -o rg.1
-sed -i 's/\.TH.*/.TH "rg" "1"/g' rg.1
+sed -i.bak 's/\.TH.*/.TH "rg" "1"/g' rg.1
+rm -f rg.1.bak # BSD `sed` requires the creation of a back-up file
--- a/src/app.rs
+++ b/src/app.rs
@@ -109,7 +109,8 @@ pub fn app() -> App<'static, 'static> {
         .arg(flag("unrestricted").short("u")
              .multiple(true))
         .arg(flag("invert-match").short("v"))
-        .arg(flag("word-regexp").short("w"))
+        .arg(flag("word-regexp").short("w").overrides_with("line-regexp"))
+        .arg(flag("line-regexp").short("x"))
         // Third, set up less common flags.
         .arg(flag("after-context").short("A")
              .value_name("NUM").takes_value(true)
@@ -348,6 +349,10 @@ lazy_static! {
              "Only show matches surrounded by word boundaries. This is \
               equivalent to putting \\b before and after all of the search \
               patterns.");
+        doc!(h, "line-regexp",
+             "Only show matches surrounded by line boundaries.",
+             "Only show matches surrounded by line boundaries. This is \
+              equivalent to putting ^...$ around all of the search patterns.");
 
         doc!(h, "after-context",
              "Show NUM lines after each match.");
--- a/src/args.rs
+++ b/src/args.rs
@@ -427,7 +427,8 @@ impl<'a> ArgMatches<'a> {
     ///
     /// Note that if -F/--fixed-strings is set, then all patterns will be
     /// escaped. Similarly, if -w/--word-regexp is set, then all patterns
-    /// are surrounded by `\b`.
+    /// are surrounded by `\b`, and if -x/--line-regexp is set, then all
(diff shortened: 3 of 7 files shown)


doc/convert-to-man
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
 
-pandoc -s -t man rg.1.md -o rg.1
-sed -i 's/\.TH.*/.TH "rg" "1"/g' rg.1
+pandoc -s -t man rg.1.md -o rg.1 &&
+sed -i.bak 's/\.TH.*/.TH "rg" "1"/g' rg.1 &&
+rm -f rg.1.bak # BSD `sed` requires the creation of a back-up file
```
okdana: I found that this script didn't run on macOS with just `sed -i`, as the BSD version of `sed` requires the back-up suffix. So i added that in a way that should be compatible with both GNU and BSD `sed`. I also added the `&&` to make the script bail if any part of the process fails.

src/app.rs
```diff
@@ -103,7 +103,8 @@ pub fn app() -> App<'static, 'static> {
         .arg(flag("unrestricted").short("u")
              .multiple(true))
         .arg(flag("invert-match").short("v"))
-        .arg(flag("word-regexp").short("w"))
+        .arg(flag("word-regexp").short("w").overrides_with("line-regexp"))
```
okdana: It should actually be fine to apply both the `word-regexp` and `line-regexp` pattern transformations at once, since `\b` also matches the beginning/end of line. But i figured it'd be cleaner and more future-proof to make the one override the other.

src/args.rs
```diff
@@ -479,10 +480,10 @@ impl<'a> ArgMatches<'a> {
         Ok(self.str_pattern(s))
     }
 
-    /// Converts a &str pattern to a String pattern, including word
+    /// Converts a &str pattern to a String pattern, including line/word
     /// boundaries or escapes if applicable.
     fn str_pattern(&self, pat: &str) -> String {
-        let s = self.word_pattern(self.literal_pattern(pat.to_string()));
+        let s = self.line_pattern(self.word_pattern(self.literal_pattern(pat.to_string())));
```
okdana: This line exceeds 80 cols, and it's becoming quite unwieldy anyway, so i think maybe i should change it. I wasn't sure what would be the best way to do that though.

src/args.rs
```diff
@@ -511,6 +512,16 @@ impl<'a> ArgMatches<'a> {
         }
     }
 
+    /// Returns the given pattern as a line pattern if the -x/--line-regexp
+    /// flag is set. Otherwise, the pattern is returned unchanged.
+    fn line_pattern(&self, pat: String) -> String {
+        if self.is_present("line-regexp") {
+            format!(r"^(?:{})$", pat)
```
okdana: Added non-capturing group to avoid the same problem described in #506.

okdana:
btw @BurntSushi how do you feel about this? Does it fit with your vision for the tool's feature set? If so, is there anything you'd need me to do (besides rebase now) to get it in?

If not, that's cool too obv; i guess i'm just wondering if i should keep it tracking new changes or simply close it

doc/convert-to-man
BurntSushi: > I also added the && to make the script bail if any part of the process fails.

I usually put `set -e` at the top of scripts for that. Could we do that instead? It feels a little nicer than trailing every command with `&&`. :-)

(I'm not up on my shell compatibility, but if this means switching `#!/bin/sh` to `#!/bin/bash`, then I'm fine with that.)

src/args.rs
BurntSushi: Yeah I'd definitely like to stick to a 79 column limit (inclusive). I'd probably just write it like this:

```rust
let litpat = self.literal_pattern(pat.to_string());
let s = self.line_pattern(self.word_pattern(litpat));
```

But if you want to make a new method, that's cool too!

okdana:
Thanks! Updated:

* Rebased
* Accounted for new zsh completion function
* Added `-e` to `convert-to-man` (it does make more sense here, and it works everywhere)
* Changed long line in `args.rs` as suggested

PR #523, opened 2017-06-19
wincolor: Re-fetch the console on all calls

alexcrichton:
The primary motivation for this commit was rust-lang/cargo#4189 where dropping a
`wincolor::Console` would call `CloseHandle` to close the console handle. Cargo
creates a few `Console` instances so it ended up closing stdout a little
earlier as intended!

The `GetStdHandle` function returns handles I believe aren't intended to be
closed (as there's no refcounting). I believe libstd doesn't close these
handles.

This commit also moves to calling `GetStdHandle` on demand which libstd changed
to doing so recently as well, preventing caching of stale handles that change
over time with calls to `SetStdHandle`.

### DIFF
--- a/wincolor/src/win.rs
+++ b/wincolor/src/win.rs
@@ -2,7 +2,7 @@ use std::io;
 use std::mem;
 
 use kernel32;
-use winapi::{DWORD, HANDLE, WORD};
+use winapi::{DWORD, WORD};
 use winapi::winbase::{STD_ERROR_HANDLE, STD_OUTPUT_HANDLE};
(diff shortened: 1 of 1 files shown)


BurntSushi:
All right, I buy it!

I feel like there was a reason why I wasn't re-creating the console, but my guess is that it was an attempt at a premature optimization.

alexcrichton:
Thanks! Mind publishing a new version as well?

BurntSushi:
Ah right, `0.1.4` is on crates.io!

PR #528, opened 2017-06-26
Remove all inline attributes

vegai:
Inline attributes don't seem to have any effect on performance on my own tests. However, I was unable to properly verify this, as the benchsuite program did not quite work on my box.

Feel free to kill this PR if you know I'm not right :)

### DIFF
--- a/src/decoder.rs
+++ b/src/decoder.rs
@@ -178,7 +178,6 @@ impl<R: io::Read, B: AsMut<[u8]>> DecodeReader<R, B> {
     ///
     /// If the internal buffer is too small to read additional bytes, then an
     /// error is returned.
-    #[inline(always)] // massive perf benefit (???)
     fn fill(&mut self) -> io::Result<()> {
         if self.pos < self.buflen {
             if self.buflen >= self.buf.as_mut().len() {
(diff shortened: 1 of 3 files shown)


BurntSushi:
Undoubtedly, some of these probably should be removed. But if you aren't seeing any difference for any of them, then that makes me immediately skeptical. There needs to be a more thorough analysis, because I did put them there for a reason. Namely, at some point in time, I did notice a performance difference.

vegai:
Would it be possible that rustc has improved in a way that makes the explicit inlines obsolete? But I agree that my own benchmarks were primitive and probably did not stress the relevant parts enough.

I'd be happy to run the benchsuite against this version and a regular release version, but I wasn't able to on Arch Linux at least.

BurntSushi:
> Would it be possible

Anything is possible. :-)

Note that some of these inline attributes will only impact certain invocations of ripgrep. For example, `src/decoder.rs` is only used when decoding UTF-16.

> I'd be happy to run the benchsuite against this version and a regular release version, but I wasn't able to on Arch Linux at least.

I use Archlinux. There's an example here: http://blog.burntsushi.net/ripgrep/#benchmark-runner --- If you can't get it to work, then please file an issue and we can figure it out. Please include more details than "it doesn't work."

DoumanAsh:
I could test your PR on windows i suppose.
I'm not really fond on doing manual suggestions to compiler, but sometimes it might be necessary and there is nothing wrong with it.
After all you have inline attributes for a reason.

BurntSushi:
The only way I'd merge this PR is if there is a careful and thorough analysis of each `inline` annotation. That's what I would do. I think it would be better if we tackled it one at a time, so I'm going to close this for now.

vegai:
Yeah, agreed.

Issue #163, opened 2016-10-11
add --iglob flag


hefengrren:
Take HTML for example: `ack --html` matches *.html, *.HTML, *.Html ...

But, `rg -thtml` only matches *.html.

BurntSushi:
Should `-g '*.html'` also be case insensitive? Should every glob be case insensitive?

BurntSushi:
Do other tools that support globbing match case insensitively?

BurntSushi:
Should we really be spending resources trying to match extensions like `hTmL`? That doesn't seem smart.

If there are common case-insensitive equivalents of certain extensions, then maybe they should just be explicitly added.

samlh:
> Should -g '*.html' also be case insensitive? Should every glob be case insensitive?

On windows, definitely.

hefengrren:
ack support the selection (case sensitive or not) for globing: 
> ack2 **-g -i** filepattern | ack2 -x -w searchpattern

BurntSushi:
I would like to see precedent established by other tools before moving
forward. Why is Windows special?

BurntSushi:
For example, what do grep, find and git do?

hefengrren:
Windows is not special, OSX has the same "issue" for compatibility: https://www.quora.com/Why-does-OS-X-choose-to-have-a-case-insensitive-file-system-instead-of-a-case-sensitive-one .

Git has a config `core.ignoreCase`.

BurntSushi:
Folks, I am neither a Windows nor a Mac user. I don't have a whole lot of knowledge of how those platforms work. Please help me understand the problems and what other tools do.

Git's `core.ignoreCase` is a good start. That's something `ripgrep` could look for.

Note that enabling case insensitive globbing could result in a very large performance hit.

samlh:
Sorry for the terse reply earlier - it was just a gut reaction as a Windows user. The general rule on Windows is that all file operations are case insensitive, alas...

I've since tested git on Windows (with core.ignoreCase on), and .gitignore files are case insensitive there. I do think supporting this is important, even with a performance hit, as my fellow teammates add rules expecting that `**/Logs/*` will also match `**/logs/*` (from two different tools).

Otherwise, find.exe from the git for windows distribution behaves as on Linux, unless you change the flag:
```
>"c:\Program Files\Git\usr\bin\find.exe" . -name *.CS

>"c:\Program Files\Git\usr\bin\find.exe" . -iname *.CS
[...]

PR #531, opened 2017-06-29
add --iglob flag

PeterSP:
Working with Chris Stadler, implemented
https://github.com/BurntSushi/ripgrep/issues/163#issuecomment-300012592

This addresses #163

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -336,6 +336,21 @@ sherlock!(glob_negate, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
     assert_eq!(lines, "file.py:Sherlock\n");
 });
 
+sherlock!(iglob, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
+    wd.create("file.HTML", "Sherlock");
+    cmd.arg("--iglob").arg("*.html");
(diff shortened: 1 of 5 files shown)


tests/tests.rs
```diff
@@ -336,6 +336,21 @@ sherlock!(glob_negate, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
...
+sherlock!(iglob, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
+    wd.create("file.HTML", "Sherlock");
+    cmd.arg("--iglob").arg("*.html");
+    let lines: String = wd.stdout(&mut cmd);
+    assert_eq!(lines, "file.HTML:Sherlock\n");
+});
+
+sherlock!(csglob, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
+    wd.create("file1.HTML", "Sherlock");
+    wd.create("file2.html", "Sherlock");
```
PeterSP: This is `file1.HTML` &`file2.html` and not `file.HTML` & `file.html` because the test is intended to create two distinct files and that would not work in the latter case on a case-insensitive filesystem such as that of Mac OS X or Windows.

BurntSushi:
This looks like great work! Thanks @PeterSP and Chris!

Issue #532, opened 2017-06-29
No zsh filename tab-autocompletion


mattsmithdatera:
Info:
```zsh
rg --version
ripgrep 0.5.2

zsh --version
zsh 5.2 (x86_64-apple-darwin15.0.0)
```
Reproduction steps:
1. Use zsh shell
2. Copy _rg autocompletion file to zsh /usr/share/zsh/site-functions location
3. type `rg "test-match" /usr/
4. Start pressing `tab`.
5. Nothing is suggested

Expected behavior (demonstrated with bash shell, where this functions correctly):
1. Use bash shell
2. copy rg bash autocompletion file to /usr/local/etc/bash_completion.d
3. type `rg "test-match" /usr/`
4. Start pressing `tab`
5. Files in /usr/ are suggested for autocompletion

Additional info:
- Zsh autocompletes ripgrep options without issue
- Zsh fails to autocomplete files regardless of presence of zsh autocompletion file in site-functions folder
- ripgrep installed via Homebrew on OSX

BurntSushi:
Someone else will need to handle this.

BurntSushi:
(I'm really not happy about ripgrep providing auto completions like this. I don't have the inclination to debug their problems.)

okdana:
The zsh completion function in the Homebrew package is not the one that's in the `ripgrep` repository. Not sure where theirs comes from, there's no attribution or anything. The one in the repo seems to work fine for me, at least in terms of completing the file name. There are some other (trivial) issues with it, though, that i can put together a PR for if you want. (The biggest one is that the descriptions for `--count` and `--max-count` are the same, which causes zsh to treat them as aliases for each other.)

BurntSushi:
@okdana PRs would be most appreciated. :-)

BurntSushi:
It sounds like this has been fixed, but you need to use the manually curated zsh completion script here: https://github.com/BurntSushi/ripgrep/blob/master/complete/_rg (Which should be the one that is bundled in the release tarball.)

PR #536, opened 2017-07-01
Improve zsh completion function

okdana:
This change is related to #532, though the actual problem there had already been fixed by the earlier addition of the hand-written completion function. In going over that function i found several issues/limitations which i've corrected:

* Added missing options (most notably `-o`)

* Fixed the confusion between `--count` and `--max-count`

* Improved the consistency of the wording — The previous descriptions had mixed usage of capitalisation, punctuation, contractions, &c., some of them were copied from the `--help` output verbatim, and so on

* Added completion for encodings — I didn't want to include a gigantic list of every single individual encoding, so i compressed them all into a few lines by making use of brace expansion. As i noted in the code, it's *super* hard to read this way, but i doubt that list will be changing any time soon, so i think the brevity is worth it (and you could expand them all by just passing them to `printf '%s\n'`, anyway)

* Added completion for colour specs — Might need revisiting if/when #452 lands

* Added partial completion for type specs — I wanted to make it so that it would complete a list of types if you gave it an `include:` directive, but i ran up against the limits of my knowledge of zsh's completion functions trying to figure out how to do that whilst making it clear that you can provide a glob *or* an `include` directive

I think the structure of this file is kind of weird, honestly — how are future maintainers supposed to intuit which options are 'common'? And is the break-down here even accurate? Is `--encoding` *really* more commonly used than `-m` or `-C`? I highly doubt it.

If i had written it i probably would have just put all of the options in alphabetical order. I was afraid it'd be bad etiquette for me to do that here, though, so i didn't.

### DIFF
--- a/complete/_rg
+++ b/complete/_rg
@@ -47,70 +47,73 @@ local -a common_options
 common_options=(
   '(-a --text)'{-a,--text}'[search binary files as if they were text]'
   '(-c, --count)'{-c,--count}'[only show count of matches for each file]'
-  '--color=[whether to use coloring in match]::when:( always never auto )'
+  '--color=[specify when to use colors in output]:when:( never auto always ansi )'
   '(1)*'{-e,--regexp=}'[specify pattern]:pattern'
-  '(-E --encoding)'{-E,--encoding=}'[specify the text encoding of files to search.]:encoding'
-  '(-F --fixed-strings)'{-F,--fixed-strings}'[treat the pattern as a literal string instead of a regular expression]'
+  '(-E --encoding)'{-E,--encoding=}'[specify text encoding of files to search]: :->encoding'
+  '(-F --fixed-strings)'{-F,--fixed-strings}'[treat pattern as literal string instead of regular expression]'
   '*'{-g,--glob=}'[include or exclude files for searching that match the given glob]:glob'
-  '(-h --help)'{-h,--help}'[prints help information]'
-  '(-i -s -S --ignore-case --case-sensitive --smart-case)'{-i,--ignore-case}'[case insensitive search]'
+  '(-h --help)'{-h,--help}'[display help information]'
+  '(-i -s -S --ignore-case --case-sensitive --smart-case)'{-i,--ignore-case}'[search case-insensitively]'
   '(-n -N --line-number --no-line-number)'{-n,--line-number}'[show line numbers]'
   '(-n -N --line-number --no-line-number)'{-N,--no-line-number}'[suppress line numbers]'
-  '(-q --quiet)'{-q,--quiet}'[do not print anything to stdout]'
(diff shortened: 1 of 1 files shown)


BurntSushi:
This is awesome work, thank you! If the bash completion were this good, I might actually use it, so color me green with envy. :-)

> If i had written it i probably would have just put all of the options in alphabetical order. I was afraid it'd be bad etiquette for me to do that here, though, so i didn't.

That would be fine. I'm not sure about why it's split like that either. The `man` page is split that way, but `--encoding` is certainly not considered common, as you say. :-)

lilianmoraru:
These completion files are generated by `clap-rs`.
Is it okay to modify these by hand?
I see 2 problems:
- either next time `clap-rs` generates the completion file, it overwrites everything
- or this completion file is maintained by hand and then it could easily diverge from the flags that are set with `clap-rs`.

If there are problems with the completion, it should probably be reported to `clap-rs`, the maintainer is very responsive.

lilianmoraru:
Ow, sorry, it was written by hand already.
Anyway should probably be reported to `clap-rs` what exactly the library should do better.

okdana:
There are [several out-standing issues](https://github.com/kbknapp/clap-rs/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20zsh%20) related to zsh completion in the `clap` repo, many of which are relevant to `ripgrep`.

The two main issues with the auto-generated function were that it didn't complete operands (`pattern` and `path`) correctly, and it didn't account for options that take arguments (except for `--color`). I think both of those have been reported.

Another (less critical) issue is that it is currently no-where near advanced enough to generate code for complex option-argument completion like what i've added for `--colors` and `--type-add`. It looks like they're discussing how they might eventually do that though.

In the mean time, it *is* bothersome that the completion function can get out-of-synch with the arguments that are actually in the app. Maybe it could be part of the test suite somehow?

I didn't actually try this, but i think running a script like this would be an effective way of ensuring that the options in the `--help` output match the ones in the completion function:

```
while read -r option; do
    ./target/release/rg -q -- "${option}" ./complete/_rg || {
        echo "Completion function missing option: ${option}" >&2
        exit 1
    }
done < <(
    ./target/release/rg --help |
    ./target/release/rg -io -- ' (-[a-z0-9]|--[a-z0-9-]+)\b' |
    tr -d ' ' |
    sort -u
)
```

Not sure how worried anyone is about it, though, since it's not critical functionality. 🤷‍♀️

BurntSushi:
I'm in favor of adding things to CI to help keep this stuff in sync. However, it must not ever have false negatives (or more precisely, any false negatives must be intermittent). False positives are not ideal obviously, but I could live with them (because it's no worse than what we have now, which is nothing).

PR #538, opened 2017-07-03
Add --iglob to zsh completion function

okdana:
In light of #531 getting merged


BurntSushi:
Ah great, thank you! :-)

PR #540, opened 2017-07-06
Add test script to compare `rg --help` output to zsh completion function

okdana:
Per our discussion in #536: Here's something to hopefully prevent the zsh completion function from getting out-of-synch with the actual app.

I used a slightly different strategy from the one i mentioned before, but it's the same basic idea — pull all of the options out of the `--help` output, pull all of the options out of the completion function, compare the two. I tried to address your concerns about false positives/negatives as best i could; i've documented the assumptions i make in the script, and i think it should be OK as long as nobody does anything bizarre.

I've tested this in `dash`, `bash`, and `zsh`. It also works in `ksh93` if you make `local` an alias for `typeset`... but you use `local` in your other scripts so i assume `ksh` compatibility isn't a big concern.

Anyway, it turns out actually that the `-f` option is missing from the completion function, which is fortuitous i guess because it seems to demonstrate that the script behaves as expected! Here's the output:

```
heartswap:ripgrep % ci/test_complete.sh
zsh completion options differ from `--help` options:
--- `rg --help`
+++ complete/_rg
@@ -77,5 +77,4 @@
 -c
 -e
--f
 -g
 -h
```

Unless you have concerns with it i think i'm happy with the script itself. I'm not sure how to go about integrating it with your other tests though. Should i just add it to `appveyor.yml` or does it belong somewhere else?

### DIFF
--- a/ci/test_complete.sh
+++ b/ci/test_complete.sh
@@ -1,0 +1,83 @@
+#!/bin/sh
+
+# Compares options in `rg --help` output to options in zsh completion function
+
+set -e
+
+main() {
+    local rg="target/${TARGET}/release/rg"
+    local _rg='complete/_rg'
+    local ret='0'
+    local helpTemp="$( mktemp )"
+    local compTemp="$( mktemp )"
+    local diff
+
+    [ -e "${rg}" ] || rg="target/${TARGET}/debug/rg"
+
+    if [ ! -e "${rg}" ]; then
+        printf 'File not found: %s\n' "${rg}" >&2
+        ret='1'
+    elif [ ! -e "${_rg}" ]; then
+        printf 'File not found: %s\n' "${_rg}" >&2
+        ret='1'
+    else
+        # 'Parse' options out of the `--help` output. To prevent false positives
+        # we only look at lines where the first non-white-space character is `-`
+        "${rg}" --help |
+        "${rg}" -- '^\s*-' |
+        "${rg}" -io -- '[\t ,](-[a-z0-9]|--[a-z0-9-]+)\b' |
+        tr -d '\t ,' |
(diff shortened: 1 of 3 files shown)


ci/test_complete.sh
```diff
@@ -0,0 +1,81 @@
+#!/bin/sh
+
+# Compares options in `rg --help` output to options in zsh completion function
+
+set -e
+
+main() {
+    local rg="target/${TARGET}/release/rg"
```
BurntSushi: We probably can't do this exact thing. This would at, the very least, depend on ripgrep being built before these tests are run. I think we could probably do it by using a debug build, but right now, a release build is not built in CI. It might be easier to just use `grep`. :P (I think you could set `rg=grep` and it will just work.)

BurntSushi: Oh hmm I see you're using the `--replace` flag. If you just change `rg` to `target/${TARGET}/debug/rg`, then I think that should be OK.

okdana: Yeah, i could easily have replicated most of the find/replace stuff with `grep` and/or `sed`, but i figured that since i needed `rg` for the `--help` output anyway i might as well just use it for everything

okdana:
Updated the `rg` path and added the script to `script.sh` — seems to have worked!

Would you like me to add completion for `-f` as part of this PR, or should i make another one?

BurntSushi:
> Would you like me to add completion for -f as part of this PR, or should i make another one?

Whichever is most convenient for you! I'm find smushing it into this one via a separate commit.

okdana:
Done, seems like everything's passing now!

BurntSushi:
@okdana You're awesome. :-) Thanks!

Issue #524, opened 2017-06-20
RFE: show enabled features in --version output


g2p:
Just to make sure that SSE and AVX are enabled.

See `systemd --version` for what the output could look like.

g2p:
Is it okay if I tweak the output to be more like systemd produces, before this is released?

BurntSushi:
@g2p I think I'm OK with that. Is systemd's output standard or common?

g2p:
It's machine parseable at least; I picked it as the best looking option I'm aware of.

PR #543, opened 2017-07-07
Update `--version` output to show compile-time features

okdana:
#524 sounded fun and easy. I only just barely know what i'm doing in Rust though — is this a reasonable way to go about this? I don't think the method will scale elegantly beyond three or four features, but i figured something fancier could quickly be devised if it came to that in the future.

This is the sort of output you get (i don't have Rust nightly so i added a dummy option to test; it's not present in the actual commit of course):

```
heartswap:ripgrep % cargo build -q --release --features 'foo'
heartswap:ripgrep % target/release/rg --version
ripgrep 0.5.2, with features (+/-): +foo -avx-accel -simd-accel
```

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -50,6 +50,7 @@ pub fn app() -> App<'static, 'static> {
     App::new("ripgrep")
         .author(crate_authors!())
         .version(crate_version!())
+        .long_version(LONG_VERSION.as_str())
         .about(ABOUT)
         .max_term_width(100)
         .setting(AppSettings::UnifiedHelpMessage)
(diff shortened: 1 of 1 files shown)


BurntSushi:
@okdana Yup, this is great! This is probably roughly how I'd do it too. I'm not a huge fan of adding new features (and I actually hope to remove these once Rust's SIMD story improves such that we can enable them dynamically without relying on compilation support), so it should be easy to maintain this list! Thank you. :-)

One nit: could you put `Fixes #524` into the commit message? That way, Github will sync the commit with the issue and close it automatically. :-)

okdana:
Oops, changed it now.

BurntSushi:
Thanks!

PR #545, opened 2017-07-07
Tweak long_version features output

g2p:
This reuses the systemd convention of putting flags on a separate line.
All credit to okdana for the implementation.  Addresses #524.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -199,18 +199,18 @@ lazy_static! {
         let mut features: Vec<&str> = vec![];
 
         if cfg!(feature = "avx-accel") {
-            features.push("+avx-accel");
+            features.push("+AVX");
         } else {
-            features.push("-avx-accel");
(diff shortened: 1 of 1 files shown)


BurntSushi:
@okdana Thoughts?

okdana:
🤔

Well, it does certainly match `systemd`.

The only potential concern i can think of is that making the 'flags' in the list different from the actual feature names means that there's no longer a 1:1 mapping between the list and the crate/build stuff. Not sure if that would ever be an issue (and i think you said you eventually want to make this all unnecessary anyway?), but that was kind of my thinking with the way i wrote it i guess.

I'm cool with it if that doesn't bother you tho

BurntSushi:
Yeah, it's a little iffy, but I think the mapping is still unambiguous, and yeah, hopefully we won't need these some day.

Issue #389, opened 2017-03-01
Different -w behavior from grep/git-grep


crumblingstatue:
```
$ echo '-2' | rg -w '\-2' # rg yields no results
$ echo '-2' | grep -w '\-2' # grep (and git grep) matches
-2
```

Not sure if this is intentional or not, but it did surprise me that I didn't find what I was looking for when searching my codebase with ripgrep, and I had to resort to git-grep.

I used `-w` because I was specifically looking for the value `-2`, and not e.g. `-24`.

Using `ripgrep 0.4.0`.

BurntSushi:
This is highly interesting.

The relevant passage from `man grep` (for GNU grep) is:

```
-w, --word-regexp
       Select only those lines containing matches that form whole words. The
       test is that the matching substring must either be at the beginning of
       the line, or preceded by a non-word constituent character. Similarly,
       it must be either at the end of the line or followed by a non-word
       constituent character. Word-constituent characters are letters, digits,
       and the underscore. This option has no effect if -x is also specified.
```

The key part is "... either be **at the beginning of the line**, or preceded by a non-word constituent character." ripgrep currently implements the `-w` flag by translating the given `pattern` to `\b(?:pattern)\b`, but it looks like grep actually does `(?:^|\b)(?:pattern)(?:$|\b)`. I guess ripgrep should do that as well.

If my hypothesis is correct, then `echo ' -2' | grep -w -e '-2'` should return nothing. This is because neither ` ` nor `-` match `\w`, and therefore, `\b` shouldn't match. Interestingly, it does return a match:

```
$ echo ' -2' | grep -w -e '-2'
 -2
```

While the equivalent ripgrep command does not:

```
$ echo ' -2' | rg -e '(^|\b)-2($|\b)'
```

Interestingly, the same grep command does not either:

```
$ echo ' -2' | egrep -e '(^|\b)-2($|\b)'
```

This has to mean that my interpretation of `-w` is wrong.

Re-reading it, it now seems clearer to me that it isn't actually using word boundary assertions, since it says "preceded by a non-word constituent character" but doesn't say anything about the first letter of the match.

Looking at the source of GNU grep, I spotted this:

```
  /* In the match_words and match_lines cases, we use a different pattern
     for the DFA matcher that will quickly throw out cases that won't work.
     Then if DFA succeeds we do some hairy stuff using the regex matcher
     to decide whether the match should really count. */
  if (match_words || match_lines)
    {
[...]

PR #547, opened 2017-07-08
harmonize ripgrep's -w|--word-regexp with grep's

kevinr:
This fixes https://github.com/BurntSushi/ripgrep/issues/389

grep (and git-grep) allow a match using --word-regexp to begin and end with
non-word characters, by surrounding the match with `(^|[^[:alnum:]_])` and
`([^[:alnum:]_]|$)`.

Before this patch ripgrep, by surrounding the match with `\b`, implicitly
required the match to start with a `\w` character, and would fail to match eg.
the ` - 2` in `1 - 2`, which intuitively should be allowed.

This fixes this behavior by surrounding the match with `(?:^|\W)` and
`(?:\W|$)`, using the Unicode non-word character class `\W` rather than the
ASCII character class `[:alnum:]`

I have tested that this combines correctly with `--only-matching`, that is, leading non-word characters are not printed along with the match when that option is used.

I have *not* tested that this combines correctly with `--replace`; however, use of non-capturing matches should ensure that references are numbered correctly.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -505,7 +505,7 @@ impl<'a> ArgMatches<'a> {
     /// flag is set. Otherwise, the pattern is returned unchanged.
     fn word_pattern(&self, pat: String) -> String {
         if self.is_present("word-regexp") {
-            format!(r"\b(?:{})\b", pat)
+            format!(r"(?:^|\W)(?:{})(?:\W|$)", pat)
         } else {
             pat
         }
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -1738,6 +1738,21 @@ fn regression_389_match_words_bounded_by_non_word_characters() {
     assert_eq!(lines, expected);
 }
 
+// See: https://github.com/BurntSushi/ripgrep/issues/389
+#[test]
+fn regression_389_match_multiple_words_bounded_by_non_word_characters() {
+    let wd = WorkDir::new("regression_389_match_multiple_words_bounded_by_non_word_characters");
+    let path = "file.txt";
+    wd.create(path, "foo-bar-bar-baz\n");
+
+    let mut cmd = wd.command();
+    cmd.arg("-w").arg("-o").arg("bar").arg(path);
+    let lines: String = wd.stdout(&mut cmd);
+
+    let expected = "bar\nbar\n";
+    assert_eq!(lines, expected);
+}
+
 #[test]
 fn type_list() {
     let wd = WorkDir::new("type_list");
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -1723,6 +1723,21 @@ fn regression_506_word_boundaries_not_parenthesized() {
 
 }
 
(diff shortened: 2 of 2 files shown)


kevinr:
Ah. Let me fix the tests and I'll give it another shot.

okdana:
I *think* that in order for this method to work correctly you'd have to change the printer so that when it prints the match it checks to see if `-w` is active, and if the match start/end is not equal to the line start/end it contracts the match by one character on each applicable side.

That would account for the standard behaviour as well as the `-o` behaviour, anyway. Not sure about `--replace`, might be more complicated.

BurntSushi:
Unfortunately, I think @okdana is right. The key difference is that the `\b(pat)\b` formulation uses zero length assertions, so the surrounding `\b` assertions don't actually consume any input in the match. But when you start using `\W`, that might consume at most one Unicode codepoint.

kevinr:
Oh crud. I can reproduce, I wasn't seeing the issue before because of an unrelated issue. To check my understanding, it looks like the code around https://github.com/BurntSushi/ripgrep/blob/master/src/printer.rs#L252 is going to need to either pull the start and end of the target submatch from the match object in the presence of --word-regexp, rather than using the whole match, or increment and decrement the start and end of the whole match by the length of the beginning and ending submatches? I was trying to avoid capturing matches, but perhaps that can't be helped (and perhaps that's why GNU grep uses capturing matches too).

okdana:
If you were going to change the way it's printed, i think you'd have to change it in two different places (`write_match()` and `write_matched_line()`), because non-line-per-match matches (i.e. the default behaviour) have a second search performed on them to figure out where to place the colours. But in each case it's just something like:

```rust
let m_start = m.start() + (m.start() != 0) as usize;
let m_end   = m.end() - (m.end() != buf.len() - 1) as usize;
```

I think probably a bigger issue though is the fact that `\W` not being zero-width means that you can't match two 'words' that are next to each other. For example:

```
given: rg -w bar <<< foo-bar-bar-baz

current method ([...] == match):
  final pattern: \b(?:bar)\b
  actual match:  foo-[bar]-[bar]-baz
  printed match: foo-[bar]-[bar]-baz

proposed method ([...] == match):
  final pattern: (?:^|\W)(?:bar)(?:\W|$)
  actual match:  foo[-bar-]bar-baz
  printed match: foo-[bar]-bar-baz
```

So you'd have to deal with that somehow too. GNU `grep` seems to have some [special handling](http://git.savannah.gnu.org/cgit/grep.git/tree/src/dfasearch.c#n376) for this.

BurntSushi:
@kevinr This PR seems to have gotten stalled, and that's mostly my fault for not staying on top of it. Based on the comments above, this issue appears to be a bit more complex than initially thought. For now, I'd like to consider fixing this issue as part of the libripgrep effort, so I'm going to close this PR. I will sure reference back to it for the tests though! Thanks!

PR #549, opened 2017-07-08
Types extension and Yocto renaming to BitBake

lilianmoraru:
Yocto -> BitBake: because `bitbake` is just one tool of the many tools that the `Yocto` project has and these files are specific to it.
GNUmakefile: the `make` tool is case-sensitive on this one
*.prf: is basically the equivalent of `*.pri`, just that one is included through `include(file)`(where you have `file.pri`) and another is included through `CONFIG += file`(where you have `file.prf` usually in `mkspecs`).


BurntSushi:
Seems good, thanks!

PR #554, opened 2017-07-13
[ignore] Fix matched_path_or_any_parents() for patterns ending in slash

behnam:
In `matched_path_or_any_parents()` implementation, we missed the point
that when we start walking up the tree, we have to set `is_dir` to
`true`, so path `ROOT/a/b/c` matches pattern `/a/`, although the
original path is not a dir.

### DIFF
--- a/ignore/src/gitignore.rs
+++ b/ignore/src/gitignore.rs
@@ -220,15 +220,17 @@ impl Gitignore {
             !path.has_root(),
             "path is expect to be under the root"
         );
-        loop {
-            match self.matched_stripped(path, is_dir) {
-                Match::None => match path.parent() {
-                    Some(parent) => path = parent,
-                    None => return Match::None,
-                },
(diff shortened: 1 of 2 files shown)


behnam:
I didn't test this diff locally (because I did this on a different machine than the Cargo patch) and the bug was revealed in my tests for the cargo patch. Fortunately, it doesn't involve API breaking changes.

Sorry for making more release work for you, @BurntSushi, and thanks in advance. :)

BurntSushi:
Nice find! If you do need to make a breaking API change, then I expect we could get away with a yank. :)

I'll cut a release when I'm at my computer next.

behnam:
Just a friendly reminder to publish `ignore:0.2.2`, as the current one is buggy. :)

BurntSushi:
Ah right, I was on vacation and forgot. `ignore 0.2.2` should now be on crates.io!

Issue #342, opened 2017-01-24
ripgrep output is invisible in PowerShell


nzbart:
When I launch cmd.exe and run PowerShell from within that console window, the output is perfectly legible:

![image](https://cloud.githubusercontent.com/assets/1592130/22260190/287ee1de-e2cd-11e6-8edd-7a36cc123aa5.png)

However, if I launch PowerShell and run rg within the default window, it appears that ripgrep renders the file name in blue rather than magenta, making it invisible:

![image](https://cloud.githubusercontent.com/assets/1592130/22260289/8c94631a-e2cd-11e6-9094-acc822e6c123.png)

When I tried setting my colour scheme using the PowerShell console configuration window (Alt-Space, P from within the console) so that it matched the cmd.exe settings, I noticed that ripgrep was outputting blue text.

Why would ripgrep change the output colour of the file name in some cases, particularly when the background colour is blue, and is it possible to make this algorithm a bit smarter?

BurntSushi:
These are the default styles used by ripgrep in all circumstances: https://github.com/BurntSushi/ripgrep/blob/f5a2d022ecd16dfe67ea829267e9220136f20167/src/args.rs#L706-L711

```rust
        let mut specs = vec![
            "path:fg:magenta".parse().unwrap(),
            "line:fg:green".parse().unwrap(),
            "match:fg:red".parse().unwrap(),
            "match:style:bold".parse().unwrap(),
        ];
```

In other words, ripgrep isn't changing anything on you. I can't really explain what you're observing though. Maybe the blend of colors makes it invisible? Maybe Powershell has a more limited color palette and can't display magenta? I don't know.

You might consider trying to experiment with the color settings, which are configurable on the command line:

```
        --colors <SPEC>...
            This flag specifies color settings for use in the output. This flag may be provided
            multiple times. Settings are applied iteratively. Colors are limited to one of eight
            choices: red, blue, green, cyan, magenta, yellow, white and black. Styles are limited
            to nobold, bold, nointense or intense.
            
            The format of the flag is {type}:{attribute}:{value}. {type} should be one of path,
            line or match. {attribute} can be fg, bg or style. {value} is either a color (for fg
            and bg) or a text style. A special format, {type}:none, will clear all color settings
            for {type}.
            
            For example, the following command will change the match color to magenta and the
            background color for line numbers to yellow:
            
            rg --colors 'match:fg:magenta' --colors 'line:bg:yellow' foo.
```

[...]

PR #557, opened 2017-07-16
Fix invisible file path text in PowerShell

latkin:
Fix for https://github.com/BurntSushi/ripgrep/issues/342

Root cause is that the standard Windows PowerShell launch shortcut sets the console background color to "dark magenta", then re-defines "dark magenta" as RGB(1, 36, 86) aka "powershell noble blue." Thus if a console app writes output with "dark magenta" foreground color, it will be invisible in a default PowerShell console.

Fix is to do a simple, Windows-only check for a "dark magenta" background at startup, and in this case change path output style to "intense magenta," which will render fine.

Result:

image

Potential concerns:
  - Arg parsing now has to know about Windows console. That's not great but it was simplest path forward for me as a rust n00b.
  - This will add "intense" styling to path output even when user has manually overridden the color themselves. Additional `path:style:nointense` will be needed to counteract this change.
    - Seems not terribly burdensome. By definition this will only affect users who are already comfortable overriding color settings.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -692,7 +692,10 @@ impl<'a> ArgMatches<'a> {
     fn color_specs(&self) -> Result<ColorSpecs> {
         // Start with a default set of color specs.
         let mut specs = vec![
+            #[cfg(unix)]
             "path:fg:magenta".parse().unwrap(),
+            #[cfg(windows)]
+            "path:fg:cyan".parse().unwrap(),
             "line:fg:green".parse().unwrap(),
             "match:fg:red".parse().unwrap(),
             "match:style:bold".parse().unwrap(),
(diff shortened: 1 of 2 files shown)


wincolor/src/win.rs
```diff
@@ -97,12 +97,24 @@ impl Console {
     ///
     /// If there was a problem setting attributes on the console, then an error
     /// is returned.
-    pub fn bg(&mut self, intense: Intense, color: Color) -> io::Result<()> {
+    pub fn set_bg(&mut self, intense: Intense, color: Color) -> io::Result<()> {
```
BurntSushi: I'd rather not rename these methods, since that would be a breaking change.

latkin:
@BurntSushi No problem, I understand your concerns. If you are ok with simply changing the Windows defaults, then that's even easier :-)

PR udpated. New behavior:

image

BurntSushi:
Fantastic, thank you!

PR #560, opened 2017-07-18
Refactor completion function

okdana:
Hey again! I've been collecting issues left over from my previous effort and i think i've found as many as i can for now.

The biggest problem, which was carried over from the original, was that option completion was basically broken on zsh 5.1 (the latest version available in Ubuntu Xenial). But i also found some typos in option names, missing actions for options that take arguments, missing exclusion rules, &c.

In addition to fixing bugs i improved the `colorspec` handling slightly, and i added completion for sequences of types when given `--type-add <type>:include:`. And i re-ordered the options to match the `--help` output like i mentioned wanting to do before.

I tested this on zsh 5.1, zsh 5.2, and zsh 5.3 and it seems OK. Really hard to test properly though, so i'm sure i'll find something again....

### DIFF
--- a/ci/test_complete.sh
+++ b/ci/test_complete.sh
@@ -1,6 +1,10 @@
 #!/bin/sh
 
+##
 # Compares options in `rg --help` output to options in zsh completion function
+#
+# @todo If we could rely on zsh being installed we could change all of this to
+# simply source the completion-function file and pull the rg_args array out...
 
 set -e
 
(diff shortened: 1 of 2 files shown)


ci/test_complete.sh
```diff
@@ -1,6 +1,10 @@
 #!/bin/sh
 
+##
 # Compares options in `rg --help` output to options in zsh completion function
+#
+# @todo If we could rely on zsh being installed we could change all of this to
+# simply source the completion-function file and pull the rg_args array out...
```
BurntSushi: It seems like you could probably install zsh in the Travis CI config?

BurntSushi:
*sigh* It looks like Windows CI is failing due to an unrelated test failure. Merging this anyway.

ci/test_complete.sh
okdana: Oh, i wasn't sure it was an option. Knowing that it is, i'll probably do that the next time i need to make changes to the function.

PR #565, opened 2017-07-26
zsh completion... again

okdana:
Sorry for PR spam, found another issue with this. :/

I hadn't noticed that the old completion function didn't support completing values without a space after short options — doing something like `rg -C3 <tab>` instead of `rg -C 3 <tab>` would break the completion.

I fixed that. And, as discussed last time, i updated the test stuff so that it just pulls the optspecs straight out of the function. That should make the 'parsing' a lot more reliable (at the expense of requiring zsh for the test, obv).

Not sure if i put the zsh-installation stuff in the right place, guess we'll see what it says...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,13 @@ env:
   global:
     - PROJECT_NAME=ripgrep
     - RUST_BACKTRACE: full
+
+addons:
+  apt:
+    packages:
+      # Needed for completion-function test
+      - zsh
+
 matrix:
   include:
     # Nightly channel.
--- a/ci/test_complete.sh
+++ b/ci/test_complete.sh
@@ -1,91 +1,94 @@
-#!/bin/sh
+#!/usr/bin/env zsh
 
 ##
 # Compares options in `rg --help` output to options in zsh completion function
-#
-# @todo If we could rely on zsh being installed we could change all of this to
-# simply source the completion-function file and pull the rg_args array out...
 
-set -e
+emulate -R zsh
+setopt extended_glob
+setopt no_function_argzero
+setopt no_unset
+
+get_comp_args() {
+    setopt local_options unset
+
+    # Our completion function recognises a special variable which tells it to
+    # dump the _arguments specs and then just return. But do this in a sub-shell
+    # anyway to avoid any weirdness
+    ( _RG_COMPLETE_LIST_ARGS=1 source $1 )
+    return $?
+}
 
 main() {
-    local rg="target/${TARGET}/release/rg"
-    local _rg='complete/_rg'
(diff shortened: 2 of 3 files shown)


okdana:
Oh, i guess Travis doesn't actually use that arch. I looked for the `addons.apt.packages` file it mentioned but i don't see it. I don't know much about Travis unfortunately :/

Also i guess i should make it so that `script.sh` skips the test if zsh isn't installed? Rather than just exploding?

BurntSushi:
No worries about the PRs! This is kinda how I expect this sort of thing to go. Iterate and improve!

With respect to Travis, I think you can install zsh by putting a special incantation into the `.travis.yml`. Like this: https://docs.travis-ci.com/user/installing-dependencies/#Installing-Packages-on-Standard-Infrastructure

> Also i guess i should make it so that script.sh skips the test if zsh isn't installed? Rather than just exploding?

I think exploding is fine actually. That way, if zsh is somehow not installed, then we aren't tricked into believing the test is running and passing. If we want to test somewhere that zsh isn't feasible to install, then we should specifically ignore the test on that configuration.

ci/install.sh
```diff
@@ -45,11 +45,23 @@ EOF
     fi
 }
 
+install_zsh() {
+    case $TARGET in
+        aarch64-unknown-linux-gnu)
```
BurntSushi: I don't think I understand this. Why are you trying to install zsh only when testing on aarch64? aarch64 is ARM. I suspect ripgrep probably works on ARM, but I don't run CI on it.

okdana: Yeah, i kind of just copy-and-pasted the section above it without thinking too much about it. In hind-sight i see that it special-cased ARM because the C tool chain is arch-specific. But since you mentioned you didn't write these scripts i assume even that is probably just boiler-plate.

BurntSushi: Ah, I [see now](https://github.com/BurntSushi/ripgrep/blob/master/ci/install.sh#L9). Yeah, all boiler plate. The original source is [rust-everywhere](https://github.com/japaric/rust-everywhere).

okdana:
I reverted the `install.sh` stuff and made it explode again if zsh is missing. I added the `zsh` package to `.travis.yml` per your link; that seems to sort it for Linux. Darwin doesn't need anything special because it ships with zsh. Looks like it worked!

okdana:
Oh, it doesn't like `pipe_fail` though. Not really necessary, will remove and rebase.

okdana:
Looks OK now i think

BurntSushi:
Awesome, looks great! Thanks so much!

PR #572, opened 2017-08-08
Remove unused libc dependency

Vurich:
I was just having a quick look at the code and realised that you were depending on `libc` but not using it anywhere, so I removed the dependency. The effect on the benchmarks is still yet to be seen.


BurntSushi:
@Vurich Thanks! This is probably leftover from before the `wincolor`/`termcolor`/`same-file` stuff existed. I don't see how this could impact benchmarks. :-)

Vurich:
I was just being pithy, sorry, I didn't actually think it would do anything either way.

PR #578, opened 2017-08-20
Pass the simd-accel feature to encoding_rs

hsivonen:
If the core functionality uses SIMD, no reason not to use it for the encoding stuff, too.


hsivonen:
It appears that the base revision failed appveyor, too, so that failure doesn't arise from this change.

BurntSushi:
@hsivonen Thanks! (And yeah, the appveyor build needs some attention.)

PR #584, opened 2017-08-25
ignore: document git_global enabled by default

durka:
This function doesn't say it's enabled by default, but it looks like it is, based on my reading of `dir.rs`: https://github.com/BurntSushi/ripgrep/blob/30608f2444a2abc7aae8e7b96cd5cff4f98bfe71/ignore/src/dir.rs#L429-L435


BurntSushi:
Yup, that's right! Thanks!

Issue #200, opened 2016-10-29
ripgrep doesn't stop when its pipe is closed


BurntSushi:
For example, the following two `rg` commands take the same amount of time, but the second one should be much shorter:

```
[andrew@Cheetah subtitles] ls -lh OpenSubtitles2016.raw.en 
-rw-r--r-- 1 andrew users 9.3G Sep 10 11:51 OpenSubtitles2016.raw.en
[andrew@Cheetah subtitles] time rg 'Sherlock Holmes' OpenSubtitles2016.raw.en | wc -l
5107

real    0m1.602s
user    0m1.250s
sys     0m0.350s
[andrew@Cheetah subtitles] time rg 'Sherlock Holmes' OpenSubtitles2016.raw.en | head -n1
You read Sherlock Holmes to deduce that?

real    0m1.626s
user    0m1.247s
sys     0m0.377s
```

BurntSushi:
This one isn't going to be fun to fix. This is what I get for ignoring any errors that occur when printing to `stdout`. (And I really should know better, I handle this correctly in `xsv`.) The issue is that a write to `stdout` in this case will fail with a pipe error, at which point, we should stop searching and quit.

The primary difficulty at present is bubbling up an error from the printing code all the way through the search code.

My question for this in terms of UX is: do we treat all IO errors equally when writing output? Should we do something different if we see a pipe error versus, say, a permission error?

BurntSushi:
I might elect to forgo fixing this until I factor the search code out into a separate crate. (Which will be a while. My guess is at least a month.)

danr:
I upgraded yesterday from rg 0.3.2-1 to 0.4 and now these commands are essentially useless:

    rg --files | head
    rg --files | fzf

This is too bad, since I use rg+fzf in my workflow.

BurntSushi:
@danr There was no regression. This bug was in `0.3.2` as well.

"useless" sounds like a bit of an exaggeration.

danr:
Sorry, I really did not mean to sound harsh. Thank you for your work on `ripgrep`!

kpp:
This is how grep acts:
```
$ strace grep Holmes subtitles/OpenSubtitles2016.raw.en 2>grep.log  | head -n 1

...
read(3, "s\nWell, Watson, what about him?\n"..., 32768) = 32768
write(1, "Roy Holmes.\nThen Simon Harrison "..., 4096) = 4096
[...]

PR #586, opened 2017-08-27
restore the default SIGPIPE behavior as a temporary workaround

oconnor663:
See https://github.com/BurntSushi/ripgrep/issues/200.

This is a Unix-only workaround, until the larger refactoring mentioned in that issue is done. I'm not sure what our current behavior is on Windows, but this should't change it in any case. Are there any cases where rg has side effects, where a sudden termination like this might not be a good idea?


oconnor663:
Thanks for the quick turnaround!

PR #587, opened 2017-08-28
Update encoding_rs to 0.7.0

hsivonen:
This is a performance improvement. The theoretically breaking changes (removal of a cargo feature and removal of `Encoding::for_name()`) don't affect ripgrep.


BurntSushi:
w00t. Thanks!

PR #588, opened 2017-08-29
ignore: add grouped defaults toggle

durka:
Adds a single function to toggle all the enabled-by-default options on `WalkBuilder`. I implemented this for [my crate](https://github.com/haptics-nri/bible-extractor/blob/76769934d69c6cd9857884e91e30eb2f6370973a/src/main.rs#L98) because I wanted a parallel directory iterator but I don't actually need the fancy ignore stuff. Feel free to reject if that isn't an intended use case.

### DIFF
--- a/ignore/src/walk.rs
+++ b/ignore/src/walk.rs
@@ -571,6 +571,29 @@ impl WalkBuilder {
         self
     }
 
+    /// Enables all the standard ignore filters.
+    /// 
+    /// This toggles, as a group, all the filters that are enabled by default:
+    /// 
+    /// - [hidden()](#method.hidden)
+    /// - [parents()](#method.parents)
+    /// - [ignore()](#method.ignore)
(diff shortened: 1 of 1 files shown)


BurntSushi:
@durka Hmmmm. I guess I'm in principle OK with something like this, but I think the name and/or documentation isn't quite clear enough. Can we come up with something better? For example, I think it would be clearer if instead of focusing on "defaults" we focused on "filtering." So I think the docs for this function should explicitly call out that the filters for this iterator are all enabled by default, and this this function can be used to toggle their presence as a group. I think we should also say that filters may be selectively enabled by the caller after using this.

durka:
Maybe `default_filters` or `standard_filters`?

BurntSushi:
I think I like `standard_filters`.

durka:
I changed the name and added more docs.

Issue #478, opened 2017-05-09
improve error messages when -f flag fails


BurntSushi:
For example:

```
$ rg -e foo -f src/search_stream.rs 
Error parsing regex near 'fer.|*/|z{' at character offset 213: Invalid application of repetition operator to: '(?u:f)(?u:o)(?u:o)|(?u:/)*(?u:!)|(?u:T)(?u:h)(?u:e)
```

Aside from the badly formatted regex error message (see #395), the error message should show the file path and probably also the line number at which the error occurred.

nateozem:
let me see what I can do.

andrea-prearo:
@BurntSushi I submitted a [PR](https://github.com/BurntSushi/ripgrep/pull/590) to address this issue. It also tries to avoid the problem with the AST dump (issue https://github.com/BurntSushi/ripgrep/issues/395).

I have one question. Were you thinking about reporting the file path and line number where the error occurred in [regex](https://github.com/rust-lang/regex)?

BurntSushi:
> I have one question. Were you thinking about reporting the file path and line number where the error occurred in regex?

Probably not. I could maybe see a case for pushing the line number reporting into the regex crate, but certainly not the file path.

andrea-prearo:
Got it. Thanks!

petr-tik:
looks like you can close this issue due to rust-lang/regex#396

I am lurking around the repo as a happy user of ripgrep - thanks for the awesome project

BurntSushi:
@petr-tik Thanks for the kind words, but this is definitely not fixed yet. For example:

```
$ cat /tmp/patterns
foo
bar(wat
quux

$ rg -f /tmp/patterns
regex parse error:
    foo|bar(wat|quux
           ^
error: unclosed group
```

There are a few problems here:

1. The file path of the pattern file isn't shown in the error message. This is important if you use `-f` multiple times.
2. There is no line number corresponding to the pattern that failed to parse.
3. The error shows *all* of the patterns in the file given joined with a `|` instead of just the pattern that is invalid.

petr-tik:
Ok, I must have misunderstood the PRs. Only wanted to help you declutter your issues, in case it had been forgotten.

J-Kappes:
Which files in the repo could be relevant to this?

c0d3l0v3r-HeHe:
Hi @BurntSushi , just checking whether this issue is still open.
[...]

Issue #395, opened 2017-03-07
Providing an invalid regex produces a misformatted error message


bdesham:
I tried to search for the regex `\s*{`, which is invalid—the `{` needed to be backslash-escaped. The error that ripgrep produced was kind of bizarre though:

<img width="1030" alt="Screenshot of my terminal, showing the weird error message from ripgrep" src="https://cloud.githubusercontent.com/assets/354230/23663706/5f83e890-0321-11e7-9b8b-11a63b1e9a35.png">

I’m using ripgrep 0.4.0 with the built-in macOS Terminal application. The text encoding is set to UTF-8.

Here is a hex dump of ripgrep’s error message:

```
$ rg "\s*{" 2>&1 | hexdump -C
00000000  45 72 72 6f 72 20 70 61  72 73 69 6e 67 20 72 65  |Error parsing re|
00000010  67 65 78 20 6e 65 61 72  20 27 5c 73 2a 7b 27 20  |gex near '\s*{' |
00000020  61 74 20 63 68 61 72 61  63 74 65 72 20 6f 66 66  |at character off|
00000030  73 65 74 20 33 3a 20 49  6e 76 61 6c 69 64 20 61  |set 3: Invalid a|
00000040  70 70 6c 69 63 61 74 69  6f 6e 20 6f 66 20 72 65  |pplication of re|
00000050  70 65 74 69 74 69 6f 6e  20 6f 70 65 72 61 74 6f  |petition operato|
00000060  72 20 74 6f 3a 20 27 28  3f 75 3a 5b 09 2d 0d 20  |r to: '(?u:[.-. |
00000070  2d 20 c2 85 2d c2 85 c2  a0 2d c2 a0 e1 9a 80 2d  |- ..-....-.....-|
00000080  e1 9a 80 e2 80 80 2d e2  80 8a e2 80 a8 2d e2 80  |......-......-..|
00000090  a9 e2 80 af 2d e2 80 af  e2 81 9f 2d e2 81 9f e3  |....-......-....|
000000a0  80 80 2d e3 80 80 5d 29  2a 27 2e 0a              |..-...])*'..|
```

Other queries that end with the bad `{` produce fine-looking error messages:

```
$ rg "\s{"                   
Error parsing regex near '\s{' at character offset 3: Missing maximum in counted
repetition operator.
$ rg "a*{"
Error parsing regex near 'a*{' at character offset 2: Invalid application of
repetition operator to: '(?u:a)*'.
```

[...]

PR #590, opened 2017-08-30
Implement custom error for grep errors

andrea-prearo:
This PR tries to address the issues with error messages returned by [regex](https://github.com/rust-lang/regex):
- Issue https://github.com/BurntSushi/ripgrep/issues/478
- Issue https://github.com/BurntSushi/ripgrep/issues/395

It does that by implementing a custom error to wrap such errors and truncating them before the portion that contains all the details about the AST (` at character ...`).

For instance, the output for the example provided in Issue https://github.com/BurntSushi/ripgrep/issues/478 is:
```sh
$ rg -e foo -f src/search_stream.rs 
Regex syntax error caused by: Error parsing regex near 'fer.|*/|z{'
```

Here are other few examples of error messages:
```sh
$ rg "s{"
Regex syntax error caused by: Error parsing regex near '\s{'
```
```sh
$ rg "a*{"
Regex syntax error caused by: Error parsing regex near 'a*{'
```

### DIFF
--- a/grep/src/lib.rs
+++ b/grep/src/lib.rs
@@ -30,7 +30,7 @@ pub type Result<T> = result::Result<T, Error>;
 #[derive(Debug)]
 pub enum Error {
     /// An error from parsing or compiling a regex.
-    Regex(regex::Error),
+    Regex(GrepError),
     /// This error occurs when an illegal literal was found in the regex
     /// pattern. For example, if the line terminator is `\n` and the regex
     /// pattern is `\w+\n\w+`, then the presence of `\n` will cause this error.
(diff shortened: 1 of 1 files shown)


BurntSushi:
Hmmm... Thanks for trying to fix this! Unfortunately, I'm not sure this is quite the right path to take here. By implementing a custom error, I meant using the regex parser directly and then implementing a different error message than the one produced by `regex-syntax` directly: https://github.com/rust-lang/regex/blob/32eb9643b45bebd2ce5169e75daed6e8579e74b2/regex-syntax/src/lib.rs#L1549 --- Basically, do the same case analysis, but don't try to print any sub-expressions in the error message.

There are a couple problems with this PR as is. Firstly, the error message truncation ends up cutting off important pieces of the message. For example, `rg 's{'` used to be

```
$ rg 's{'
Error parsing regex near 's{' at character offset 2: Missing maximum in counted
repetition operator.
```

but now it's

```
$ rg 's{'                                                                                                                                                                                     
Regex syntax error caused by: Error parsing regex near 's{'
```

Personally, I think the unpleasant corner case isn't worth dropping the extra information for all regex parse errors.

The other issue I have with this PR as-is is that it's reporting less information to the caller by translating the error to a much less structured form. I'd expect to see this fixed by retaining a structured error message and then fixing the output.

At some point, I'd like to fix this for real by using a better regex parser/pretty printer.

andrea-prearo:
It makes perfect sense. I am going to close this PR.
If nobody is working on tweaking the message produced by `regex-syntax`, I may look into that.
Thanks!

BurntSushi:
@andrea-prearo Yeah I think a PR to the regex crate that removes the sub-expressions from the messages would be great!

andrea-prearo:
@BurntSushi Sounds good. I'll look into that.

petr-tik:
looks like you can close this issue due to https://github.com/rust-lang/regex/pull/396

I am lurking around the repo as a happy user of ripgrep - thanks for the awesome project

Issue #320, opened 2017-01-14
Output only matched patterns (data extraction)


Hrxn:
Hi,

just been trying ripgrep for the first time, but I couldn't find an easy and straightforward way to just output the matching pattern.

For example, trying to search in a file like this:
`rg "(\{.*\}\})" example.txt`

The regex matching works as intended, correct result is highlighted in bold red text.

But when I tried to redirect ("pipe") the output to another program I realized that the actual output is still the unchanged input.

Am I missing something here? I couldn't find an option just for this output, seemed pretty straightforward to me, but neither `rg --help` or the front page here seem to mention something..

BurntSushi:
It's a missing feature: #34. It's on my list of things to do soon, but it's part of a larger refactoring effort, so it might take a little longer.

There is a work around today though. You can use the `-r/--replace` flag:

    rg "(\{.*\}\})" example.txt -r '$1'

I think that should do it. (On mobile so i haven't checked.)

Hrxn:
> rg "(\{.*\}\})" example.txt -r '$1'

Not really. Replacing works as intended apparently, in this case:
* match and take capture group 1
* replace with reference 1, i.e. capture group 1 again

But the thing is, the actual output is still unchanged, because the program always returns the whole line where the match was found.

To help illustrate things better, I made an example file: https://fnpaste.com/YZvd

But I found a way around this: Changing the rx syntax to match the whole line, and then replace the whole line with capture group 1.

For example:
`rg ".*40 (\{.*\}\}).*" example.txt -r "$1"`

This gives me the desired part. But there is a big caveat: I use the `40 ` as an anchor to make the group match the correct part, but this is specific for this example only, I don't actually know the value.

I tried `".*(\{.*\}\}).*"` in my text editor and it actually works there, whereas the same pattern only matches the last `{...}}` part of the target inside the line with rg.

I'm not really familiar with the intricacies of regular expressions in Rust, so this needs a bit of tweaking, I guess, although it should definitely be possible.

BurntSushi:
You're right. I forgot to add that you need to make your regex match the entire line in order for `-r/--replace` to work like `-o/--only-matching`.

I think this gives the desired output for you though:

```
$ rg '^.*?(\{.*\}\}).*?$' example.txt -r '$1'
```

The trick is to make the `.*` lazy by using `.*?`. When it's lazy, it will try to match the least possible. Otherwise, it tries to match the most possible.

Hrxn:
[...]

PR #593, opened 2017-09-01
Allow -o, --only-matching work together with -r, --replace

beyondcompute:
Hello!

Thank you for your amazing work on much-needed and super-fast fully-featured regex search utility which is `ripgrep`.

However I’ve been missing a bit a feature similar to _ack’s_ helpful `--output` which allows printing only certain capture groups. That feature is discussed in https://github.com/BurntSushi/ripgrep/issues/320 and the discussion offers a solution. However it requires matching an entire line, which still might seem a bit less straightforward for some users.

Now, there is probably a reason why `-o` is made to conflict with `-r` so please excuse me if this PR is really out-of-place. However after implementing it, upon superficial gaze, things seem to be working fine:
<details>
<summary>
Examples outputs
</summary>

#### ripgrep
```bash
$ ./target/debug/rg 'of (\w+)' -Nor '$1' ./tests/hay.rs 
this
detective
luck
straw
cigar
```

#### ack
```bash
$ ack 'of (\w+)' --output '$1' ./tests/hay.rs 
this
detective
luck
straw
cigar
```
</details>

####

So I hope this has a chance of being considered. Thank you again!

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -162,7 +162,7 @@ pub fn app() -> App<'static, 'static> {
         .arg(flag("no-ignore-parent"))
         .arg(flag("no-ignore-vcs"))
         .arg(flag("null").short("0"))
-        .arg(flag("only-matching").short("o").conflicts_with("replace"))
+        .arg(flag("only-matching").short("o"))
         .arg(flag("path-separator").value_name("SEPARATOR").takes_value(true))
         .arg(flag("pretty").short("p"))
         .arg(flag("replace").short("r")
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -286,7 +286,12 @@ impl<W: WriteColor> Printer<W> {
             let line = {
                 let replacer = CountingReplacer::new(
                     self.replace.as_ref().unwrap(), &mut count);
-                re.replace_all(&buf[start..end], replacer)
+                if self.only_matching {
+                    re.replace_all(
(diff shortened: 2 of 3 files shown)


BurntSushi:
I can't quite remember why we set the flags to conflict. There might be some subtle corner cases. I'd have to go back to that PR and read.

With that said, I'm not in principle against this, and I feel like this combination of flags makes sense.

ayosec:
> I can't quite remember why we set the flags to conflict.

I guess that the conflict was set because the [original implementation](https://github.com/beyondcompute/ripgrep/commit/90a11dec5e5765af560d343e879869a60363ef54) didn't add the [`if self.only_matching`](https://github.com/beyondcompute/ripgrep/blob/90a11dec5e5765af560d343e879869a60363ef54/src/printer.rs#L310-L313) condition in the [`if self.replace.is_some()`](https://github.com/beyondcompute/ripgrep/blob/90a11dec5e5765af560d343e879869a60363ef54/src/printer.rs#L291-L309) block.

The combination of `-r` and `-o` seems pretty natural for me. Actually, I tried to write `rg -o 'Enqueued (\S+)' -r '$1'` a few minutes ago, and then I saw that there was impossible :-( .

BurntSushi:
I rebased this on top of master and merged it in https://github.com/BurntSushi/ripgrep/commit/f887bc1f8673fea9969adae6e70b68e3761ba171. Thanks!

PR #603, opened 2017-09-08
Avoid expensive check with --files

### LINKED ISSUE #600: rg --files triggers MsMpEng.exe on Windows
Version 0.6.0.

Not sure what might cause that, but MsMpEng.exe hogs the CPU and rg.exe makes only very slow progress.

When I use an alternate implementation to list the files in node.js, MsMpEng.exe remains silent or at least doesn't jam the CPU.

chrmarti:
Fixes #600

Stacktraces show that this check is the culprit. Listing all files in my test folder with 235k files on Windows now takes about 4.2 seconds. It didn't finish before I canceled before that when MsMpEng.exe kicked in.

The stacktrace:
image

/cc @BurntSushi @roblourens

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -108,6 +108,7 @@ fn run_parallel(args: Arc<Args>) -> Result<u64> {
             let dent = match get_or_log_dir_entry(
                 result,
                 args.stdout_handle(),
+                args.files(),
                 args.no_messages(),
             ) {
                 None => return Continue,
@@ -154,6 +155,7 @@ fn run_one_thread(args: Arc<Args>) -> Result<u64> {
         let dent = match get_or_log_dir_entry(
             result,
             args.stdout_handle(),
+            args.files(),
             args.no_messages(),
         ) {
             None => continue,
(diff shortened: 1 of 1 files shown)


roblourens:
Excellent detective work @chrmarti! A little background @BurntSushi, with this change we will be able to use ripgrep to power quick open in VS Code as well.

chrmarti:
The Mac CI build failed with an empty log.

BurntSushi:
@chrmarti Nice detective work! I think my key issue here is that this only fixes the problem when using the `--files` flag, but based on your observations, it sounds like searching the files will result in a tremendous bottleneck.

So I'm not quite sure what to do. It seems like we should try to fix the root issue, but maybe it would be a good idea to just take this in for now?

roblourens:
I think Defender is doing the "right thing" here, and in text search, by scanning files when they are opened. It does cache results when possible, so this won't happen every time, but it will happen often. That was basically the result of our discussion with the Defender team. So I think the best thing we can do for now is to remove an unnecessary file open when using `--files`.

BurntSushi:
@roblourens Hmm all right, that's good enough for me!

PR #604, opened 2017-09-11
Ripgrep installs from Linuxbrew just fine and works as expected

thedrow:
Mentioned it in the README file :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,7 +154,8 @@ but you'll need to have the
 [Microsoft VC++ 2015 redistributable](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
 installed.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -154,7 +154,7 @@ but you'll need to have the
 [Microsoft VC++ 2015 redistributable](https://www.microsoft.com/en-us/download/details.aspx?id=48145)
 installed.
 
-If you're a **Mac OS X Homebrew** user, then you can install ripgrep either
+If you're a **Mac OS X Homebrew** or a **Linuxbrew`` user, then you can install ripgrep either
```
BurntSushi: I think you have a Markdown error here. Linuxbrew should have `**` at the end.

Also, could you please make sure this is wrapped to 79 columns inclusive? Thanks.

Issue #609, opened 2017-09-22
Add support for terraform files


tnguyen14:
Can terraform files `.tf` be added as a supported type? Essentially `--type-add 'tf:*.tf -ttf`

BurntSushi:
Sure! Please submit a PR. :-)

tnguyen14:
Created #610.

BurntSushi:
Done in #610 610

PR #610, opened 2017-09-23
add support for terraform file type

tnguyen14:
This is to address #609.

I've not worked with Rust before, so please kindly let me know if there are things I'm missing.


BurntSushi:
Looks perfect to me! :-)

PR #620, opened 2017-10-01
Fix printing context after an early return from a search

### LINKED ISSUE #402: --max-count has precedence over --after-context and --context
I think `--after-context` (`-A`) and `--context` (`-C`) should have the same precedence as `--before-context` (`-B`), when being combined with `--max-count` (`-m`), but they do not:

```bash
$ cat > rgtest.txt <<EOF
> a
> b
> c
> d
> e
> EOF
$ rg -B2 c rgtest.txt
1-a
2-b
3:c
$ rg -C2 c rgtest.txt
1-a
2-b
3:c
4-d
5-e
$ rg -A2 c rgtest.txt
3:c
4-d
5-e
$ # and now with --max-count:
$ rg -B2 c -m1 rgtest.txt
1-a
2-b
3:c
$ rg -C2 c -m1 rgtest.txt
1-a
2-b
3:c
$ rg -A2 c -m1 rgtest.txt
3:c
```

From my expectations, `-B` works correctly, `-C` works like `-B` which is unexpected, and `-A` has no effect at all.

My use case is to show a colleague one instance of many repeated matches, but also with some context. If `-A` combined with `-m` would yield the same result as with `-B`, then my use case would have been fulfilled.

I am using ripgrep-0.3.2-1.fc24.x86_64, packaged for Fedora through copr (by carlgeorge-ripgrep).

Thanks for a great program! (edit: removed trailing whitespace in code block)

Detegr:
Print the context if there's a context left to be printed after returning early from a search (because of `--max-count`).

Fixes #402.

### DIFF
--- a/src/search_stream.rs
+++ b/src/search_stream.rs
@@ -299,6 +299,15 @@ impl<'a, R: io::Read, W: WriteColor> Searcher<'a, R, W> {
                 }
             }
         }
+        if self.after_context_remaining > 0 {
+            if self.last_printed == self.inp.lastnl {
+                try!(self.fill());
+            }
+            let upto = self.inp.lastnl;
+            if upto > 0 {
+                self.print_after_context(upto);
+            }
+        }
         if self.match_count > 0 {
             if self.opts.count {
                 self.printer.path_count(self.path, self.match_count);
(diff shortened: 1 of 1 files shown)


Detegr:
Added a couple of test cases as well that now work the same way as `grep` does.

PR #622, opened 2017-10-03
Add message to README to inform about `cargo install ripgrep` not being `strip`ed

### LINKED ISSUE #595: ripgrep binary doesn't appear to be `strip`ed on Unix like expected
rustc 1.20.0 (f3d6973f4 2017-08-27)
cargo 0.21.0 (5b4b8b2ae 2017-08-12)
ripgrep 0.6.0 -AVX -SIMD

Despite the discussion in #413 and the resultant commit 30ca3ecca65d06ebc1d0794f3968397709aff9a8, I'm still seeing rather large binaries after a fresh `cargo install ripgrep` on my CentOS 7 box (~27 MiB)

Running `strip ~/.cargo/bin/rg` gives me a binary that's a magnitude of size smaller, like expressed in #413.

Did I misunderstand the effect of that commit, or is this a bug?

dvergeylen:
Notify user cargo install ripgrep contains debug symbols and informs how to stripe them.

Fix #595

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,8 +210,11 @@ $ # (Or using the attribute name, which is also `ripgrep`.)
 ```
 
 If you're a **Rust programmer**, `ripgrep` can be installed with `cargo`.
-* Note that the minimum supported version of Rust for ripgrep is **1.17**, although ripgrep may work with older versions.
-* Note that the binary will include debug symbols. Use `strip $(which rg)` to remove them (and save Mbs).
+* Note that the minimum supported version of Rust for ripgrep is **1.17**,
+  although ripgrep may work with older versions.
+* Note that the binary may be bigger than expected because it contains debug
+  symbols. This is intentional. To remove debug symbols and therefore reduce
+  the file size, run `strip` on the binary.
 
 ```
 $ cargo install ripgrep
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -209,9 +209,9 @@ $ nix-env --install ripgrep
 $ # (Or using the attribute name, which is also `ripgrep`.)
 ```
 
-If you're a **Rust programmer**, `ripgrep` can be installed with `cargo`. Note
-that the minimum supported version of Rust for ripgrep is **1.17**, although
-ripgrep may work with older versions.
+If you're a **Rust programmer**, `ripgrep` can be installed with `cargo`.
+* Note that the minimum supported version of Rust for ripgrep is **1.17**, although ripgrep may work with older versions.
```
BurntSushi: Please format the README in a style that is consistent with the rest of the text. In this case, please wrap text to 79 columns inclusive.

README.md
```diff
@@ -209,9 +209,9 @@ $ nix-env --install ripgrep
...
+* Note that the binary will include debug symbols. Use `strip $(which rg)` to remove them (and save Mbs).
```
BurntSushi: Please reword this to:

> Note that the binary may be bigger than expected because it contains debug symbols. This is intentional. To remove debug symbols and therefore reduce the file size, run `strip` on the binary.

PR #625, opened 2017-10-07
Color --replace text using the match type

### LINKED ISSUE #521: Color replacements when using --replace
ripgrep colors matches in red (by default). That's very useful.

When using the `--replace` option, I would find it very useful if the replacements were colored. Currently, nothing is colored (except file paths and line numbers of course).

Here's a screenshot showing what I mean:

screenshot from 2017-06-17 14-21-23

And here's a text sketch:

```
$ cat test
exponential notation: 123e4
another example: 42e-10
$ rg '(\d+)e' --replace '${1}E' test
1:exponential notation: <color>123E</color>4
2:another example: <color>42E</color>-10
```

What do you think?

(By the way, thank you very much for ripgrep. It's excellent.)

jamessan:
Closes #521

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -3,29 +3,58 @@ use std::fmt;
 use std::path::Path;
 use std::str::FromStr;
 
-use regex::bytes::{Captures, Regex, Replacer};
+use regex::bytes::{Captures, Match, Regex, Replacer};
 use termcolor::{Color, ColorSpec, ParseColorError, WriteColor};
 
 use pathutil::strip_prefix;
 use ignore::types::FileTypeDef;
+
+/// Track the start and end of replacements to allow coloring them on output.
+#[derive(Debug)]
+struct Offset {
+    start: usize,
+    end: usize,
+}
+
+impl Offset {
+    fn new(start: usize, end: usize) -> Offset {
+        Offset { start: start, end: end }
+    }
+}
+
+impl<'m, 'r> From<&'m Match<'r>> for Offset {
+    fn from(m: &'m Match<'r>) -> Self {
+        Offset{ start: m.start(), end: m.end() }
+    }
+}
 
 /// CountingReplacer implements the Replacer interface for Regex,
 /// and counts how often replacement is being performed.
 struct CountingReplacer<'r> {
     replace: &'r [u8],
     count: &'r mut usize,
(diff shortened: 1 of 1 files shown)


src/printer.rs
```diff
@@ -3,29 +3,57 @@ use std::fmt;
...
     replace: &'r [u8],
     count: &'r mut usize,
+    offsets: &'r mut Vec<Offset>,
 }
 
 impl<'r> CountingReplacer<'r> {
-    fn new(replace: &'r [u8], count: &'r mut usize) -> CountingReplacer<'r> {
-        CountingReplacer { replace: replace, count: count }
+    fn new(
+        replace: &'r [u8], count: &'r mut usize,
```
BurntSushi: Could you put each parameter on its own line please?

src/printer.rs
```diff
@@ -3,29 +3,57 @@ use std::fmt;
...
 }
 
 impl<'r> Replacer for CountingReplacer<'r> {
     fn replace_append(&mut self, caps: &Captures, dst: &mut Vec<u8>) {
         *self.count += 1;
+        let start = dst.len();
         caps.expand(self.replace, dst);
+        let end = dst.len();
+        if start != end {
+            self.offsets.push(Offset::new(start, dst.len()));
```
BurntSushi: I feel like it would be clearer if this were written as `Offset::new(start, end)`.

src/printer.rs
```diff
@@ -295,44 +324,40 @@ impl<W: WriteColor> Printer<W> {
                 self.write_eol();
                 return;
             }
-            self.write(&line);
-            if line.last() != Some(&self.eol) {
-                self.write_eol();
-            }
+            self.write_matched_line(offsets, &line[..], false);
```
BurntSushi: I think `&line` should be sufficient here? I guess you might need `&*line` because `line` is a cow I think.

src/printer.rs
jamessan: Indeed. :) I originally didn't have the empty replacement check and forgot to change this to `end` when I added it.  Thanks.

src/printer.rs
jamessan: Yeah, `&*line` is needed.

jamessan:
> This looks pretty good to me!

Thanks!  My first contribution to a rust project. :)

PR #642, opened 2017-10-16
Skip regression 210 test on APFS

### LINKED ISSUE #559: Regression test #210 fails on macOS High Sierra
I'm not positive as to why, but I have a feeling it may be related to [normalization-insensitivity in APFS](https://developer.apple.com/library/content/documentation/FileManagement/Conceptual/APFS_Guide/FAQ/FAQ.html#//apple_ref/doc/uid/TP40016999-CH6-DontLinkElementID_3). If I understand the APFS docs and the test case correctly, this may actually be an issue in `std::ffi`, and if that's the case then I'll report this up the chain.

```
Stack trace:

failures:

---- regression_210 stdout ----
	thread 'regression_210' panicked at '~/src/ripgrep/target/debug/deps/ripgrep-tests/regression_210/93/foo�bar: Error { repr: Os { code: 92, message: "Illegal byte sequence" } }', tests/workdir.rs:285
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
   1: std::panicking::default_hook::{{closure}}
   2: std::panicking::default_hook
   3: std::panicking::rust_panic_with_hook
   4: std::panicking::begin_panic
   5: std::panicking::begin_panic_fmt
   6: integration::workdir::nice_err
   7: integration::workdir::WorkDir::create_bytes
   8: integration::workdir::WorkDir::create
   9: integration::regression_210
  10: <F as test::FnBox<T>>::call_box
  11: __rust_maybe_catch_panic
  12: std::panicking::try::do_call
  13: __rust_maybe_catch_panic
  14: <F as alloc::boxed::FnBox<A>>::call_box
  15: std::sys::imp::thread::Thread::new::thread_start
  16: _pthread_body
  17: _pthread_start


failures:
    regression_210

test result: FAILED. 136 passed; 1 failed; 0 ignored; 0 measured
```

sebnow:
APFS does not support creating filenames with invalid UTF-8 byte codes,
thus this test doesn't make sense. Skip it on file systems where this
shouldn't be possible.

Fixes #559

### DIFF
--- a/tests/workdir.rs
+++ b/tests/workdir.rs
@@ -50,7 +50,8 @@ impl WorkDir {
     /// Try to create a new file with the given name and contents in this
     /// directory.
     pub fn try_create<P: AsRef<Path>>(&self, name: P, contents: &str) -> io::Result<()> {
-        self.try_create_bytes(name, contents.as_bytes())
+        let path = self.dir.join(name);
+        self.try_create_bytes(path, contents.as_bytes())
     }
 
     /// Create a new file with the given name and size.
--- a/tests/workdir.rs
+++ b/tests/workdir.rs
@@ -41,12 +41,14 @@ impl WorkDir {
         }
     }
 
-    /// Create a new file with the given name and contents in this directory, or panic on error.
+    /// Create a new file with the given name and contents in this directory,
+    /// or panic on error.
     pub fn create<P: AsRef<Path>>(&self, name: P, contents: &str) {
         self.create_bytes(name, contents.as_bytes());
     }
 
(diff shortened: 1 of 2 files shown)


tests/workdir.rs
```diff
@@ -41,24 +41,34 @@ impl WorkDir {
         }
     }
 
-    /// Create a new file with the given name and contents in this directory.
+    /// Create a new file with the given name and contents in this directory, or panic on error.
```
BurntSushi: Please wrap this (and other comments) to 79 columns inclusive. Thanks!

Issue #441, opened 2017-04-08
^$ matches a phantom line at end of file


BurntSushi:
This is simple to demonstrate. Note the behavior difference between GNU grep and ripgrep:

```
$ cat scratch 
a
b

c


d
$ rg '^$' scratch
3:
5:
6:
8:
$ egrep -n '^$' scratch
3:
5:
6:
```

The `^$` pattern is a useful idiom for detecting empty lines in a file. ripgrep's behavior in particular outputs a line numbered 8, even though there are only 7 lines in the file:

```
$ wc -l scratch
7 scratch
```

More generally, ripgrep should never emit a line with a number greater than the number of lines in the file, even if the file ends with a trailing `\n`.

See also https://github.com/BurntSushi/ripgrep/issues/416 and https://github.com/rust-lang/regex/issues/355#issuecomment-292679709 for the root cause

BurntSushi:
cc @BatmanAoD

BatmanAoD:
Thanks for filing this. I agree that it's reasonable to keep the `regex` crate behavior.

(Also, I know you asked for example use-cases for this, and unfortunately, although I know I've used something like `ack '^$'` in the past, I can't actually remember the specifics. I did find some StackOverflow questions with numerous upvotes about how to search for empty lines, but I didn't see anyone discussing specifically why they wanted to do that. This may not be a common enough use-case to be a high-priority bug.)

BurntSushi:
@BatmanAoD Right. When I asked that, I hadn't thought of "match empty lines." Once I spent a couple of minutes thinking, I remembered it. That's a strong enough use case as far as I'm concerned. :-)

BurntSushi:
(I wouldn't expect someone to care about matching empty lines necessarily, but I could see someone being interested in counting them.)

blankname:
I use matching empty lines in conjunction with the --invert-match option to filter out empty lines.

BatmanAoD:
@blankname That actually shouldn't be a problem! You'll "erroneously" filter out a phantom line--which should have no effect.

@BurntSushi Counting seems like an interesting application indeed, but as soon as I thought about that I thought "isn't that just a sort of crumby, broken `loc`?"

BurntSushi:
I've never heard of, or used, `loc`. :-)

BatmanAoD:
[...]

PR #651, opened 2017-10-23
Solving extra blank line error #441

deg4uss3r:
Hi guys taking a stab at #441 

There are 3 tests that are failing:
* search_stream::tests::after_context_invert_one1
* search_stream::tests::before_context_invert_one1
* search_stream::tests::invert_match_line_numbers

This has to do with the way the tests are passed, since a command line test produces the correct assertion: 
```
> ../target/debug/rg 'Sherlock' -v -C 1 ./sherlock 
1-For the Doctor Watsons of this world, as opposed to the Sherlock
2:Holmeses, success in the province of detective work must always
3-be, to a very large extent, the result of luck. Sherlock Holmes
4:can extract a clew from a wisp of straw or a flake of cigar ash;
5:but Doctor Watson has to have it taken out for him and dusted,
6:and exhibited clearly, with a label attached.
```

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -2,6 +2,7 @@ use std::error;
 use std::fmt;
 use std::path::Path;
 use std::str::FromStr;
+use std::io::BufRead;
 
 use regex::bytes::{Captures, Match, Regex, Replacer};
 use termcolor::{Color, ColorSpec, ParseColorError, WriteColor};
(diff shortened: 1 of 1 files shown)


src/printer.rs
```diff
@@ -305,6 +306,9 @@ impl<W: WriteColor> Printer<W> {
             self.write_path_sep(b':');
         }
         if let Some(line_number) = line_number {
+            if line_number > buf.lines().count() as u64 {
+                return;
+            }
```
BurntSushi: Sorry, but this just won't work on multiple levels. Firstly, `buf` doesn't necessarily contain the entire contents of the file being searched (this *is* true for the memory map searcher only). Secondly, `buf.lines().count()` is recounting all of the lines in the buffer **every time a match is written**, which would be a devastating performance regression.

BurntSushi: Third, if the end user disabled line numbers, then this check won't run at all.

PR #652, opened 2017-10-24
Properly match the rules "**/" and "!**/"

### LINKED ISSUE #649: incorrect gitignore behaviour
Steps

1. Clone nim project `git clone https://github.com/nim-lang/Nim.git`
2. In nim folder run `rg uint64`.  It gives no results which is incorrect.  Silver searcher gives correct results
3. Used `git check-ignore lib\system.nim` to verify git doesn't ignore the file
4.  `rg --debug` output [here](https://gist.github.com/i64/130d3c63c98fcb3304cc649d23838e2b).  Lib folder is being ignored

Using 0.7.1 release from [here](https://github.com/BurntSushi/ripgrep/releases/download/0.7.1/ripgrep-0.7.1-x86_64-pc-windows-msvc.zip)

lambda:
When processing a rule that ends in a slash, we strip it off and set the
`is_only_dir` flag.  We then apply the rule that paths that aren't
absolute should be given an implicit `**/` prefix, while avoiding
adding that prefix if it already exists.

However, this means that we miss the case in which we had already
stripped off the trailing slash and set `is_only_dir`.  Correct this
by also explicitly checking for that case.

Fixes #649

### DIFF
--- a/ignore/src/gitignore.rs
+++ b/ignore/src/gitignore.rs
@@ -455,7 +455,7 @@ impl GitignoreBuilder {
         // anywhere, so use a **/ prefix.
         if !is_absolute {
             // ... but only if we don't already have a **/ prefix.
-            if !glob.actual.starts_with("**/") {
+            if !(glob.actual.starts_with("**/") || (glob.actual == "**" && glob.is_only_dir)) {
                 glob.actual = format!("**/{}", glob.actual);
             }
         }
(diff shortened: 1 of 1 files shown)


lambda:
I think the build failure here on macOS is spurious. The same test passes on my own machine (running 10.12.6, while the build machine is 10.11.6), and the test that failed is one that is supposed to panic but failed with a `SIGBUS` instead.

lambda:
Huh, failed twice the exact same way, which I can't reproduce here with the exact same nightly compiler, though a different kernel. That's strange.

BurntSushi:
That's... strange. I just kicked it again.

Looks like a great fix, thank you. :-)

BurntSushi:
Damn, defeated again. I'm not sure what's going on. I'm at work so I can't really dig into this at the moment, but I would at least like to try to get to the bottom of this before merging. (I can't imagine how this PR is causing this error though!)

lambda:
Yeah, it's really strange, I also don't see how this change could have caused this error, but it makes sense to get it worked out before merging since you do need to keep the CI working happily.

I might try spinning up a 10.11 VM this evening and see if I can repro the issue.

BurntSushi:
The same CI problem is happening with other PRs, so it's clearly not provoked by this specific PR. Therefore, I shall merge! Thanks again @lambda!

lambda:
Wonder if the CI problem was rust-lang/rust#45866?

PR #655, opened 2017-10-25
add notebook file extensions to types

mpacer:
This PR adds Jupyter notebook file extensions (`.ipynb`/`.jpynb`) to the type list. 

Some context: Notebooks are widely used in the scientific computing world. Under-the-hood are json files that are a bit unwieldy (understatement) to look at when printed to the command line. 

I used the `nb` key for notebook, but am happy to change it if that is preferred.

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -148,6 +148,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("json", &["*.json", "composer.lock"]),
     ("jsonl", &["*.jsonl"]),
     ("julia", &["*.jl"]),
+    ("jupyter", &["*.ipynb", "*.jpynb"]),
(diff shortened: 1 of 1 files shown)


mpacer:
@BurntSushi it looks like the test failure may be related to resource availability and not the code I added. Is that an accurate assessment or do I need to do something else to make this pass the tests?

BurntSushi:
I believe Mathematica uses `nb` as an extension for their notebook files. How should we handle that?

(And yes, I believe the CI failures are unrelated to this PR.)

BurntSushi:
ping @mpacer

carno:
How about `s/nb/jupyter/`?

mpacer:
oh! sorry totally missed this while I've been at a conference, I'm fine with using ipynb or jupyter, whichever is preferred.

PR #667, opened 2017-11-08
Add Protocol Buffer file type

jesse-c:
Simply adds a file type filter for [Protocol Buffers](https://developers.google.com/protocol-buffers/).


BurntSushi:
Thanks!

PR #668, opened 2017-11-09
Support both [^...] and [!...] for globset class negation

### LINKED ISSUE #663: Glob: support for [^...] notation
It looks like RipGrep is already supporting the `[^...]` syntax for glob patterns which will match on a character-range with negation. Can the `[!...]` syntax also be supported for the same purpose?

From http://man7.org/linux/man-pages/man7/glob.7.html
 
> An expression "[!...]" matches a single character, namely any
> character that is not matched by the expression obtained by removing
> the first '!' from it.  (Thus, "[!]a-]" matches any single character
> except ']', 'a' and '-'.)

okdana:
Fixes #663.

I hope this is the idiomatic way to do this. Seems to work at least:

```
% ls
abc  bbc  cbc  dbc

# Old behaviour
% \rg -u --files -g '[!ab]bc'
cbc
dbc
% \rg -u --files -g '[^ab]bc'
abc
bbc

# New behaviour
% rg-dev -u --files -g '[!ab]bc'
cbc
dbc
% rg-dev -u --files -g '[^ab]bc'
cbc
dbc
```

For the tests i just duplicated a bunch of the ones that were there for `[!...]` 🤷‍♀️

### DIFF
--- a/globset/src/glob.rs
+++ b/globset/src/glob.rs
@@ -840,12 +840,15 @@ impl<'a> Parser<'a> {
                 Ok(())
             }
         }
-        let mut negated = false;
         let mut ranges = vec![];
-        if self.chars.peek() == Some(&'!') {
-            assert!(self.bump() == Some('!'));
-            negated = true;
-        }
+        let negated = match self.chars.peek() {
+            Some(&'!') | Some(&'^') => {
+                let bump = self.bump();
+                assert!(bump == Some('!') || bump == Some('^'));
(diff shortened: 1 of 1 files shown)


okdana:
Oops. Forgot to bump the `globset` version used by `ignore`. Pushing fix now....

Does the `ignore` version need changed too?

okdana:
Oh, sure. Done, rebased

PR #657, opened 2017-10-29
Update to memmap 0.6

danburkert:
`memmap` 0.6.0 introduces major API changes in anticipation of a 1.0
release. See https://github.com/danburkert/memmap-rs/releases/tag/0.6.0
for more information. CC danburkert/memmap-rs#33.

PR #678, opened 2017-11-13
bump memmap to 0.6

ignatenkobrain:


ignatenkobrain:
This needs a bit more work because of API change. WIll update PR later today.

ignatenkobrain:
And actually this is duplicate of https://github.com/BurntSushi/ripgrep/pull/657

PR #688, opened 2017-11-19
fix a bunch of clippy warnings

matthiaskrgr:


src/main.rs
```diff
@@ -137,10 +137,8 @@ fn run_parallel(args: &Arc<Args>) -> Result<u64> {
             Continue
         })
     });
-    if !args.paths().is_empty() && paths_searched.load(Ordering::SeqCst) == 0 {
-        if !args.no_messages() {
-            eprint_nothing_searched();
-        }
+    if !args.paths().is_empty() && paths_searched.load(Ordering::SeqCst) == 0 && !args.no_messages() {
```
BurntSushi: Please drop this entire commit. I prefer the existing style.

matthiaskrgr: done :)

matthiaskrgr:
I added another commit that adds clippy as optional dependency.
```` cargo build --features clippy```` will compile with static analysis :)

File: Cargo.toml [hunk 2]
```diff
@@ -33,6 +33,7 @@ path = "tests/tests.rs"
 atty = "0.2.2"
 bytecount = "0.1.4"
 clap = "2.26"
+clippy = {version = "*", optional = true}
```
BurntSushi: Sorry, but please remove this and any corresponding `cfg`s. I don't want to add a very unstable linter to ripgrep's dependencies.

matthiaskrgr: This is not a default dependency, clippy is *only* pulled in if we explicitly build with "--features clippy"

BurntSushi: Yes, I understand. I don't want it.

matthiaskrgr: Alright, I removed it.

PR #694, opened 2017-11-27
Omit context separators when using a contextless option like -c or -l

### LINKED ISSUE #693: Don't display context separators when only listing files
```
❯ command rg --context=3 -l "context-separator"
complete/_rg
--
doc/rg.1.md
--
src/app.rs
--
src/args.rs
```

When listing files, even if `--context` is an option, I don't think we need the context separator. I understand that those options don't make much sense in this contrived example. However, I have an alias set up for `rg` that adds a bunch of options that I want as defaults, including `--context`. I (sometimes) use `-l` to list out files that I can pass to `vim`, e.g. `vim $(rg "context-separator" -l)`.

I tried to set `--context=0` when using `-l` – which would be an ok tradeoff for me – but it looks like multiple `--context` flags are not allowed.

`rg … | sed '/--/d'` is another potential solution, but it would break if `--context-separator` specified a different separator.

okdana:
Fixes #693.

When `rg` has been placed in a 'contextless mode' by the use of `--count`, `--files-with-matches`, or `--files-without-match`, it should omit the context separator even if given a context option like `-A` or `-C`. This matches how (e.g.) `grep` behaves.

Before:
```
% \rg -cC1 context.separator
complete/_rg:1
--
doc/rg.1.md:1
--
src/app.rs:3
--
src/printer.rs:9
--
src/args.rs:7
```

After:
```
% target/release/rg -cC1 context.separator
complete/_rg:1
doc/rg.1.md:1
src/app.rs:3
src/args.rs:7
src/printer.rs:9
```

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -153,14 +153,16 @@ impl Args {
 
     /// Retrieve the configured file separator.
     pub fn file_separator(&self) -> Option<Vec<u8>> {
-        let use_heading_sep =
-            self.heading
-            && !self.count
-            && !self.files_with_matches
-            && !self.files_without_matches;
+        let contextless =
+            self.count
+            || self.files_with_matches
+            || self.files_without_matches;
+        let use_heading_sep = self.heading && !contextless;
+
         if use_heading_sep {
             Some(b"".to_vec())
-        } else if self.before_context > 0 || self.after_context > 0 {
+        } else if !contextless
+            && (self.before_context > 0 || self.after_context > 0) {
             Some(self.context_separator.clone())
         } else {
             None
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -1799,6 +1799,25 @@ fn regression_568_leading_hyphen_option_arguments() {
     assert_eq!(lines, "foo -n -baz\n");
 }
 
+// See: https://github.com/BurntSushi/ripgrep/issues/693
+#[test]
+fn regression_693_context_option_in_contextless_mode() {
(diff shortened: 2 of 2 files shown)


src/args.rs
```diff
@@ -153,14 +153,15 @@ impl Args {
...
+        let contextless =
+            self.count
+            || self.files_with_matches
+            || self.files_without_matches;
+        let use_heading_sep = self.heading && !contextless;
+
         if use_heading_sep {
             Some(b"".to_vec())
-        } else if self.before_context > 0 || self.after_context > 0 {
+        } else if !contextless && self.before_context + self.after_context > 0 {
```
okdana: I changed the `>0` test slightly just so the condition would fit on one line 🤷‍♀️

BurntSushi: Haha nice, we've all been there. But... I actually think this changes behavior. I think if someone did `rg -A 9223372036854775808 -B 9223372036854775808 foo`, then the sum would wrap to exactly `0` and therefore not show any context at all. Probably best to just use the same condition and wrap the conditional (or lift it out of the `if` into its own intermediate name). Or perhaps even better, put the logic into a method or something. Up to you!

okdana: Oh, i didn't think about that. I've switched it back. Let me know if the style is wrong

BurntSushi:
Bah, looks like darwin is backlogged: https://www.traviscistatus.com/ --- Looks good enough for me.

PR #706, opened 2017-12-04
Support ignore files with custom names

ptzz:
This allows for application specific ignorefile names, e.g. using
`.fdignore` for `fd`.

- Moved the hardcoded `.rgignore` from the ignore crate to ripgrep

- `.ignore` is always respected and has the highest precedence

### DIFF
--- a/ignore/src/dir.rs
+++ b/ignore/src/dir.rs
@@ -14,7 +14,7 @@
 // well.
 
 use std::collections::HashMap;
-use std::ffi::OsString;
+use std::ffi::{OsString, OsStr};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, RwLock};
 
@@ -112,7 +112,8 @@ struct IgnoreInner {
     /// Explicit global ignore matchers specified by the caller.
     explicit_ignores: Arc<Vec<Gitignore>>,
     /// Ignore files used in addition to `.ignore`
-    ignorefiles: Vec<String>,
+    custom_ignore_filenames: Vec<OsString>,
+    custom_ignore_matcher: Gitignore,
     /// The matcher for .ignore files.
     ignore_matcher: Gitignore,
     /// A global gitignore matcher, usually from $XDG_CONFIG_HOME/git/ignore.
@@ -213,11 +214,21 @@ impl Ignore {
     /// Like add_child, but takes a full path and returns an IgnoreInner.
     fn add_child_path(&self, dir: &Path) -> (IgnoreInner, Option<Error>) {
         let mut errs = PartialErrorBuilder::default();
+        let custom_ig_matcher =
+            if !self.0.opts.ignore {
+                Gitignore::empty()
+            } else {
+                let custom_ignores : Vec<String> = 
+                    self.0.custom_ignore_filenames.iter().map(|x| x.clone().into_string().unwrap()).collect();
+                let (m, err) = create_gitignore(&dir, &custom_ignores);
+                errs.maybe_push(err);
+                m
+            };
         let ig_matcher =
             if !self.0.opts.ignore {
                 Gitignore::empty()
             } else {
-                let (m, err) = create_gitignore(&dir, &self.0.ignorefiles);
+                let (m, err) = create_gitignore(&dir, &[".ignore"]);
                 errs.maybe_push(err);
                 m
             };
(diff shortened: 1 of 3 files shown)


ptzz:
See discussion in https://github.com/sharkdp/fd/issues/156, specifically https://github.com/sharkdp/fd/issues/156#issuecomment-347371863

ignore/src/dir.rs
```diff
@@ -513,6 +525,14 @@ impl IgnoreBuilder {
         self
     }
 
+    /// Enables reading of ignore files in addition to `.ignore`
+    ///
+    /// Earlier names have lower precedence
+    pub fn ignorefile(&mut self, filename: String) -> &mut IgnoreBuilder {
```
ptzz: `ignorefile()` vs existing `add_ignore()` (for global ignorefiles) might be a bit confusing. Happy to change the name into something better. Or maybe change `add_ignore` to `add_global_ignore`?

ptzz:
Should probably add tests that uses multiple ignore files and validates precedence.

ptzz:
@BurntSushi Thanks for the comments!

> I think the way to solve your problem is to add an additional layer of ignore files that has higher precedence than .ignore but uses a custom (probably application specific) name.

Indeed I didn't read this carefully. I assumed `.ignore` would have the highest precedence -- but it's now the last element in `ignorefiles`, i.e. highest precedence. I understand that just correcting the order won't solve the hierarchy issues that you talk about though.

I will try to find time to fix this properly.

ptzz:
@BurntSushi I *think* it should be working as expected now. Also addressed your comments on naming and method signatures.

ignore/src/dir.rs
```diff
@@ -109,8 +109,11 @@ struct IgnoreInner {
     /// The absolute base path of this matcher. Populated only if parent
     /// directories are added.
     absolute_base: Option<Arc<PathBuf>>,
-    /// Explicit ignore matchers specified by the caller.
+    /// Explicit global ignore matchers specified by the caller.
     explicit_ignores: Arc<Vec<Gitignore>>,
+    /// Ignore files used in addition to `.ignore`
+    custom_ignore_filenames: Vec<OsString>,
+    custom_ignore_matcher: Gitignore,
```
BurntSushi: This should have a doc string.

ignore/src/dir.rs
```diff
@@ -210,14 +213,21 @@ impl Ignore {
 
     /// Like add_child, but takes a full path and returns an IgnoreInner.
     fn add_child_path(&self, dir: &Path) -> (IgnoreInner, Option<Error>) {
-        static IG_NAMES: &'static [&'static str] = &[".rgignore", ".ignore"];
-
         let mut errs = PartialErrorBuilder::default();
+        let custom_ig_matcher =
+            if !self.0.opts.ignore {
```
BurntSushi: I suspect we don't want this. The caller needs to supply an application specific name, I assume, in order for a matcher to be created at all. So by passing a name, they are implicitly enabling it.

ignore/src/dir.rs
```diff
@@ -109,8 +109,11 @@ struct IgnoreInner {
     /// The absolute base path of this matcher. Populated only if parent
     /// directories are added.
     absolute_base: Option<Arc<PathBuf>>,
-    /// Explicit ignore matchers specified by the caller.
+    /// Explicit global ignore matchers specified by the caller.
     explicit_ignores: Arc<Vec<Gitignore>>,
+    /// Ignore files used in addition to `.ignore`
+    custom_ignore_filenames: Vec<OsString>,
```
BurntSushi: This should be in a `Arc<Vec<OsString>>`. Otherwise you're going to wind up copying it everywhere.

ptzz:
@BurntSushi Tests have been added for 1+2 above. As for precedence tests, I added one that verifies that a custom ignore file have higher precedence than `.ignore`, and one that verifies that earlier custom ignore files have lower precedence than later.

As for 3, there are already tests that uses `.rgignore`, maybe that's sufficient? It's not hardcoded in the ignore crate anymore, but passed as a parameter from `ripgrep`.

ptzz:
@BurntSushi Do you have concerns about merging this and/or want more time for review? Let me know if there is anything I can address.

ptzz:
Thanks, it was a great learning experience doing a bit of Rust!

PR #714, opened 2017-12-15
Add an `only-capture` flag for printing only the text in the specified capture group

markprzepiora:
This is the first Rust code I've ever written so I apologize in advance if it's horrible.

This option implies `--only-matching`, and will result in only the given capture group (specified by its number) being printed.

I often find myself wanting this functionality in `rg`/`ag`/`ack`... I can imitate *some* use-cases with look-behinds/look-aheads, but not all. (Look-aheads and look-behinds need to be fixed-length, in particular.)

### DIFF
--- a/src/printer.rs
+++ b/src/printer.rs
@@ -89,6 +89,8 @@ pub struct Printer<W> {
     null: bool,
     /// Print only the matched (non-empty) parts of a matching line
     only_matching: bool,
+    /// If only_matching is true, print only the given capture number.
+    only_capture: usize,
     /// A string to use as a replacement of each match in a matching line.
     replace: Option<Vec<u8>>,
     /// Whether to prefix each match with the corresponding file name.
@@ -115,6 +117,7 @@ impl<W: WriteColor> Printer<W> {
             line_per_match: false,
             null: false,
             only_matching: false,
+            only_capture: 0,
             replace: None,
             with_filename: false,
             colors: ColorSpecs::default(),
@@ -179,6 +182,12 @@ impl<W: WriteColor> Printer<W> {
     /// Print only the matched (non-empty) parts of a matching line
     pub fn only_matching(mut self, yes: bool) -> Printer<W> {
         self.only_matching = yes;
+        self
+    }
+
+    /// If only_matching is true, print only the given capture group.
+    pub fn only_capture(mut self, number: usize) -> Printer<W> {
+        self.only_capture = number;
         self
     }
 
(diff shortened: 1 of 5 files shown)


okdana:
Are you familiar with the `--replace` option? It can do this (and more). I'm on mobile but it's something like:

```
rg --replace '$1' '.*foo(bar)baz.*'
```

markprzepiora:
Oof, how embarrassing! How did I never know about this?

Sorry to waste your time with this PR. :) At least I learned a little bit of Rust in the process.

src/printer.rs
```diff
@@ -89,6 +89,8 @@ pub struct Printer<W> {
     null: bool,
     /// Print only the matched (non-empty) parts of a matching line
     only_matching: bool,
+    /// If only_matching is true, print only the given capture number.
+    only_capture: usize,
```
BurntSushi: For reference, this should probably be an `Option<usize>`, where `Some(0)` is equivalent to `--only-matching`.

src/printer.rs
```diff
@@ -278,7 +287,9 @@ impl<W: WriteColor> Printer<W> {
             return self.write_match(
                 re, path, buf, start, end, line_number, mat.0, mat.1);
         }
-        for m in re.find_iter(&buf[start..end]) {
+        for caps in re.captures_iter(&buf[start..end]) {
+            let m = caps.get(self.only_capture).map_or(caps.get(0).unwrap(), |m| m);
```
BurntSushi: This is interesting behavior. If the capture index doesn't exist in the match, then the result is to use the entire match? That doesn't feel quite right to me. It kind of feels like it should use an empty string. And in that case, the line emitted would be empty I guess, which feels a little weird but is consistent with `--only-matching` and `--replace` I believe.

tests/tests.rs
```diff
@@ -1828,3 +1828,21 @@ fn type_list() {
     // This can change over time, so just make sure we print something.
     assert!(!lines.is_empty());
 }
+
+#[test]
+fn only_capture() {
```
BurntSushi: I suspect we need more tests than this. In particular, the case of testing the use of a capture index that sometimes doesn't exist, depending on the match.

BurntSushi:
@okdana Awesome, I think I agree. I also like the the idea of having a short flag for this too. `-O` doesn't appear to be used in grep, so I think it'd be OK to use that.

@markprzepiora If you would like to respond to the feedback here and get this PR updated, then I think this can be merged!

markprzepiora:
Thanks for the feedback, everyone! :)

I have a busy day at work for the next two days but I will grok the changes you requested and try to implement them Thursday evening.

beyondcompute:
Isn’t the same thing addressed in #593? ☺️

BurntSushi:
@beyondcompute Indeed! `rg 'fn (\w+)' -or '$1'` does the expected thing. In light of that, I think I'm going to close this out. Documentation improvements to increase awareness of combining `--only-matching` and `--replace` would be great though. Thanks @markprzepiora for the PR!

PR #715, opened 2017-12-17
Add armhf build to Travis CI

lilianmoraru:
The purpose of this is so that `ripgrep` will create the "release" binary for the most popular ARM target - but I did not manage to test if the binary is correctly pushed to the "releases" page...

### DIFF
--- a/ci/before_deploy.sh
+++ b/ci/before_deploy.sh
@@ -6,8 +6,12 @@ set -ex
 
 # Generate artifacts for release
 mk_artifacts() {
-    RUSTFLAGS="-C target-feature=+ssse3" \
-      cargo build --target $TARGET --release --features simd-accel
+    if is_ssse3_target; then
+        RUSTFLAGS="-C target-feature=+ssse3" \
+        cargo build --target $TARGET --release --features simd-accel
+    else
+        cargo build --target $TARGET --release
+    fi
 }
 
 mk_tarball() {
@@ -15,11 +19,12 @@ mk_tarball() {
     local td=$(mktempd)
     local out_dir=$(pwd)
     local name="${PROJECT_NAME}-${TRAVIS_TAG}-${TARGET}"
-    mkdir "$td/$name"
+    local gcc_prefix="$(gcc_prefix)"
+    mkdir "${td:?}/${name}"
     mkdir "$td/$name/complete"
 
     cp target/$TARGET/release/rg "$td/$name/rg"
-    strip "$td/$name/rg"
+    ${gcc_prefix}strip "$td/$name/rg"
     cp {doc/rg.1,README.md,UNLICENSE,COPYING,LICENSE-MIT} "$td/$name/"
     cp \
       target/$TARGET/release/build/ripgrep-*/out/{rg.bash-completion,rg.fish,_rg.ps1} \
--- a/ci/script.sh
+++ b/ci/script.sh
@@ -4,6 +4,9 @@ set -ex
 
 . $(dirname $0)/utils.sh
 
+# "." - dot is for the current directory(ripgrep itself)
+components=( . grep globset ignore termcolor )
+
 # NOTE Workaround for rust-lang/rust#31907 - disable doc tests when cross compiling
 # This has been fixed in the nightly channel but it would take a while to reach the other channels
 disable_cross_doctests() {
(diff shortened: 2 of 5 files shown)


lilianmoraru:
While trying to make sure that ARM tests are not run in the CI, I observed that `cargo clean` was being run only on `ripgrep` and not on all the subprojects(in `run_test_suite()`) - while analyzing if this is ok, I observed that every subproject gets its own `target` directory which means that these don't share dependencies.

Using cargo workspaces here should improve the build speed.

ci/utils.sh
```diff
@@ -54,3 +54,13 @@ architecture() {
             ;;
     esac
 }
+
+# 1 - if TRUE
```
okdana: Having a shell function return `1` on success/true and `0` on failure/false is pretty unorthodox. Also there's a lot of unnecessary forking here. It doesn't really matter, it's not like it's performance-critical or anything, but (in the spirit of some of your other changes) it would probably be more idiomatic to do something like this:

```sh
case "${TARGET}" in
  i686-unknown-netbsd) return 1 ;;
  i686*|x86_64*)       return 0 ;;
esac
return 1
```

Then you can use it like:

```sh
if is_ssse3_target; then
  # do SSSE3 stuff
else
  # do not-SSSE3 stuff
fi
```

ci/before_deploy.sh
```diff
@@ -6,20 +6,26 @@ set -ex
 
 # Generate artifacts for release
 mk_artifacts() {
-    RUSTFLAGS="-C target-feature=+ssse3" \
-      cargo build --target $TARGET --release --features simd-accel
+    is_ssse3_target
+    if [ $? -eq 1 ]; then
```
okdana: Related to my other comment: I don't believe this construction will work reliably anyway — if `is_ssse3_target` returns `1` here, it will cause the script to bail because of `set -e`. The effect of `set -e` needs to be suppressed by putting it in the `if` condition (or using `||` or `&&`)

ci/utils.sh
lilianmoraru: > Having a shell function return 1 on success/true and 0 on failure/false is pretty unorthodox.

That's what I thought too but then decided to make it more traditional programming languages-like... :)
Good point though and I will change it.

Btw, I used `$?` because I tried to do:
```sh
is_ssse3_target() { return 0; }
test is_ssse3_target && echo "TRUE" || echo "FALSE"
```
And this would not work, but I see that this works:
```sh
if is_ssse3_target; then echo "TRUE"; else echo "FALSE"; fi
```
The alternative was to not use early-return and just `echo true/false`.

ci/script.sh
```diff
@@ -15,28 +18,23 @@ disable_cross_doctests() {
...
+        cargo "${1:?}" --target $TARGET --verbose --manifest-path "${component}/Cargo.toml"
+    done
 }
 
 main() {
     # disable_cross_doctests
-    run_test_suite
+    run_cargo clean
+    run_cargo build
+    if [ "$(architecture)" = "amd64" ] || [ "$(architecture)" = "i386" ]; then
```
BurntSushi: Does this mean we aren't running tests on arm? Why?

ci/utils.sh
```diff
@@ -54,3 +54,13 @@ architecture() {
             ;;
     esac
 }
+
+# 1 - false
+# 0 - true
```
BurntSushi: I think you can just remove this comment. As @okdana said, `1` for `false` is standard in shell scripts.

BurntSushi:
Also, thanks for the tip about workspaces. I will do it soon.

ci/script.sh
lilianmoraru: In order to run the tests on ARM, you would need to setup `qemu` inside the Travis CI.
This is doable(I've seen it done somewhere) but I don't currently have the knowledge(nor would I like to hold this PR for that) to easily add this feature.

It would definitely add overhead - you would probably need to limit it to run only when preparing a release or something like this.

lilianmoraru:
> Also, thanks for the tip about workspaces. I will do it soon.

The thing is almost a one-liner(works on `1.17`):
```
[workspace]
members = [ "grep", "globset", "ignore", "termcolor", "wincolor" ]
```

But you need to remove `[profile.release]` from `ignore`.

For the full "Win", if all the packages would have the same versions of the dependencies than you would have full re-use.

At some point the CI should also be switched to use `cargo build --all`.

lilianmoraru:
Does not seem complicated: https://github.com/PJK/libcbor/blob/master/.travis-qemu.sh

lilianmoraru:
I pushed another change:
Installing specifically `gcc-4.8-arm-linux-gnueabihf` because if Travis moves to `Ubuntu 16.04` at some point, `gcc-arm-linux-gnueabihf` would become `GCC 5.3`, which would be incompatible with RPI's `GCC 4.9`(glibc API/ABI).

ci/script.sh
BurntSushi: Oh I see, I understand. Thanks for explaining that to me!

lilianmoraru:
@BurntSushi Speaking of overhead. What's your opinion on removing the `cargo clean` step?
I have not yet encountered problems with incremental compilation unless you are building a C/C++ application through `build.rs`(which `ripgrep` doesn't).
That would allow to ask Travis to cache the `target` directory and have considerably better build speeds.

BurntSushi:
@lilianmoraru It sounds fine to me. I doubt there is any strong reason for it to be there now. It's probably there because I copied it from somewhere else, and it was probably there because I was debugging something and just didn't remove it. :)

PR #718, opened 2017-12-18
Improve smartness of --smart-case

### LINKED ISSUE #717: --smart-case detection of upper-case characters is unintuitive
The `--smart-case` option doesn't behave the way i expect in terms of determining whether the pattern contains upper-case characters. Specifically, a search like the following will fail:

```
rg -S 'foo\w' <<< FOOBAR
```

This is because it expands the class `\w` and discovers that it contains the range of upper-case characters `A-Z`.

`ag` uses a different metric: It simply checks the input pattern byte by byte to see if there's an upper-case character in the ASCII range. This works out in the `foo\w` case, but it will fail in the similarly obvious case of `foo\S` (which `rg` happens to get right) — not to mention multi-byte stuff like `É` and `ß`.

There are also more complicated cases where they both fail in different ways, like this:

```
rg -S '\p{Ll}' <<< FOOBAR
```

`rg` returns a match here, because `\p{Ll}` expands to only lower-case characters. But i think this should be treated as an 'explicit' range, like `[A-Z]` would be.

At the very least, though, i feel like something as basic as `foo\w` should probably work as expected with smart case.

Is there any way to determine through the AST the actual input that produced a class? Or would that be easy to add? I'm not sure i'm at the level yet where i should be messing with the `regex` crate, but that seems like it'd be a convenient way to sort it.

The easiest 'good enough' alternative that comes to mind would be to do a semi-naïve scan of the pattern that's just smart enough to ignore stuff like short-hand classes, but that feels kind of dumb (especially since i get the impression that you don't really care for this feature in the first place...).

okdana:
Fixes #717 (partially)

As suggested, i changed the smart-case feature in the `grep` crate to look for upper-case characters in the pattern itself rather than the AST. I also added some tests for it, and a note in the manual.

I didn't update the `grep` version number — you don't want me to do that, right?

### DIFF
--- a/grep/src/search.rs
+++ b/grep/src/search.rs
@@ -102,9 +102,9 @@ impl GrepBuilder {
 
     /// Whether to enable smart case search or not (disabled by default).
     ///
-    /// Smart case uses case insensitive search if the regex is contains all
-    /// lowercase literal characters. Otherwise, a case sensitive search is
-    /// used instead.
+    /// Smart case uses case insensitive search if the pattern contains only
+    /// lowercase characters (ignoring any characters which immediately follow
+    /// a '\'). Otherwise, a case sensitive search is used instead.
     ///
     /// Enabling the case_insensitive flag overrides this.
     pub fn case_smart(mut self, yes: bool) -> GrepBuilder {
@@ -197,8 +197,6 @@ impl GrepBuilder {
     }
 
     /// Determines whether the case insensitive flag should be enabled or not.
-    ///
-    /// An error is returned if the regex could not be parsed.
     fn is_case_insensitive(&self) -> Result<bool> {
         if self.opts.case_insensitive {
             return Ok(true);
@@ -206,12 +204,7 @@ impl GrepBuilder {
         if !self.opts.case_smart {
             return Ok(false);
         }
-        let expr =
-            try!(syntax::ExprBuilder::new()
-                 .allow_bytes(true)
-                 .unicode(true)
-                 .parse(&self.pattern));
-        Ok(!has_uppercase_literal(&expr))
+        Ok(!has_uppercase_literal(&self.pattern))
     }
 }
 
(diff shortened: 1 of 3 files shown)


grep/src/search.rs
```diff
@@ -317,38 +310,26 @@ impl<'b, 's> Iterator for Iter<'b, 's> {
...
+/// negate the effect of the smart-case option.
+///
+/// Ideally we would be able to check the AST in order to correctly handle
+/// things like '\p{Ll}' and '\p{Lu}' (which should be treated as explicitly
+/// cased), but we don't currently have that option. For now, our 'good enough'
+/// solution is to simply perform a semi-naïve scan of the input pattern and
+/// ignore all characters following a '\'. The ExprBuilder will handle any
+/// actual errors, and this at least lets us support the most common cases,
+/// like 'foo\w' and 'foo\S', in an intuitive manner.
+fn has_uppercase_literal(pattern: &String) -> bool {
```
BurntSushi: This should accept a `&str`. I believe the callers won't need to be changed because of auto deref.

(Basically, you almost never want a `&String` or a `&Vec<T>`, and instead should prefer `&str` or `&[T]`. It obviously doesn't _really_ matter because this is an internal helper function, but I do want to stay idiomatic.)

grep/src/search.rs
```diff
@@ -317,38 +310,26 @@ impl<'b, 's> Iterator for Iter<'b, 's> {
...
+        } else if c.is_uppercase() {
+            return true;
         }
-        Group { ref e, .. } => has_uppercase_literal(e),
-        Repeat { ref e, .. } => has_uppercase_literal(e),
-        Concat(ref es) => es.iter().any(has_uppercase_literal),
-        Alternate(ref es) => es.iter().any(has_uppercase_literal),
-        _ => false,
     }
+    return false;
```
BurntSushi: This should just be `false` (no semi-colon) instead of `return false;`. :-)

grep/src/search.rs
```diff
@@ -395,4 +376,20 @@ mod tests {
...
+        assert_eq!(has_uppercase_literal(&"foO".to_string()), true);
+        assert_eq!(has_uppercase_literal(&"foo\\\\".to_string()), false);
+        assert_eq!(has_uppercase_literal(&"foo\\w".to_string()), false);
+        assert_eq!(has_uppercase_literal(&"foo\\S".to_string()), false);
+        assert_eq!(has_uppercase_literal(&"foo\\p{Ll}".to_string()), true);
+        assert_eq!(has_uppercase_literal(&"foo[a-z]".to_string()), false);
+        assert_eq!(has_uppercase_literal(&"foo[A-Z]".to_string()), true);
+        assert_eq!(has_uppercase_literal(&"foo[\\S\\t]".to_string()), false);
+        assert_eq!(has_uppercase_literal(&"foo\\\\S".to_string()), true);
+    }
```
BurntSushi: Awesome, thank you for these tests!

okdana:
Oops. Fixed i think

PR #726, opened 2018-01-01
Replace all usages of try! macro with the ? operator

balajisivaraman:
So as I was learning and going through the codebase, I came across `try!` a lot. Thinking it was a new macro/syntax I had to learn, I searched the Internet for it only to [learn](https://doc.rust-lang.org/std/macro.try.html) that it was an earlier alternative to the newer `?` operator, which should be preferred nowadays.

As a result of this, I have changed all usages of the `try!` macro to the newer `?` operator in the codebase.

My reasoning is as follows:
- The latest edition (2nd) of the Rust Programming language book teaches newcomers about the `?` as the shortcut for [propagating errors](https://doc.rust-lang.org/book/second-edition/ch09-02-recoverable-errors-with-result.html#a-shortcut-for-propagating-errors-).
- I also find that `?` is easier to parse and read than `try!` and is much less of an overhead on my brain. (Note: Subject to personal taste.)

I did not want to create a new issue for this since I assume we only want to do that for actual issues with RipGrep and not changes we want to make to the codebase. So I went ahead and created a PR instead. 

Please feel free to close this PR if you all feel otherwise and are OK with letting `try!` remain as such. Thanks!

### DIFF
--- a/globset/README.md
+++ b/globset/README.md
@@ -36,7 +36,7 @@ This example shows how to match a single glob against a single file path.
 ```rust
 use globset::Glob;
 
-let glob = try!(Glob::new("*.rs")).compile_matcher();
+let glob = Glob::new("*.rs")?.compile_matcher();
 
 assert!(glob.is_match("foo.rs"));
 assert!(glob.is_match("foo/bar.rs"));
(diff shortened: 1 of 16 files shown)


BurntSushi:
Awesome, thanks! I did indeed want to do this change at some point. For some history, ripgrep was written before `?` was stabilized (or somewhere around then). Even after `?` was stabilized, I wanted to continue supporting older versions of Rust. But enough time has passed, and ripgrep's minimum version is well above Rust 1.13 now anyway.

There are dependencies of ripgrep (like the regex crate) that still aim to support Rust 1.12, and therefore can't use `?` yet.

PR #728, opened 2018-01-01
Documentation updates (path operand, &c.)

### LINKED ISSUE #725: glob ignored for items in path
I found the following surprising:
```
$ echo bar > bar
$ rg -g 'foo' --files *
bar
$ rg -g 'foo' --files bar
bar
```

I was expecting the above rg commands to return nothing.

okdana:
Fixes #725 (i guess?)

* I capitalised the place-holders `options` and `path` to make them consistent with `PATTERN` &al.

* I added documentation of the positional arguments to the man page

* I added a note to both the man page and the usage help to clarify that an explicit `PATH` overrides glob/ignore rules. This could possibly be clearer i guess — a glob/ignore rule is only overridden if the path as given on the command line contains a segment that actually matches it — but it didn't seem worth writing a whole paragraph about

* I'm not sure when this became an issue, but when i regenerated the man page all of the long option names were broken — Pandoc had converted `--foo` into `–foo` (with an en dash). I updated `convert-to-man` to disable 'smart typography', which seems to have fixed it for Pandoc 2.0.6; hopefully it doesn't cause issues for contributors on older versions


BurntSushi:
RE ci failures: My bet is on the capitalization of `path`.

okdana:
lol, oops. Should have done a more thorough test than just `rg --help` i guess.

Pushed a fix, it seems better now.

PR #732, opened 2018-01-06
Update crossbeam to 0.3.2

stjepang:
Crossbeam 0.3.0 uses the old auto trait syntax and 0.3.2 switched to the new one. Ripgrep is a part of Rust's test suite and, since it depends on crossbeam 0.3.0, it is currently blocking the following PR, which removes the old auto trait syntax in favor of the new one: https://github.com/rust-lang/rust/pull/46480#issuecomment-355735400

Diff between versions 0.3.0 and 0.3.2: https://github.com/crossbeam-rs/crossbeam/compare/v0.3.0...v0.3.2

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -55,7 +55,7 @@ dependencies = [
 
 [[package]]
 name = "crossbeam"
-version = "0.3.0"
+version = "0.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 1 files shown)


BurntSushi:
@stjepang Cool! Do you pin to a specific commit on ripgrep, or should I prioritize a new release?

stjepang:
Rust's test suite pins to a specific commit, so we're covered there.

But I still suggest releasing a new version because crossbeam 0.2.10 will not compile with nightly rustc as soon as https://github.com/rust-lang/rust/pull/46480 gets merged. Even worse, I'm afraid it won't compile with stable rustc in the future either (from version 1.25 onwards).

BurntSushi:
So that means newer versions of Rust won't be able to compile older versions of ripgrep? That doesn't sound good to me.

BurntSushi:
@stjepang Anyway, thanks so much for the heads up! I'm glad this won't catch me by surprise. :) I will see if I can get the full story at rust-lang/rust.

stjepang:
Yes, the problem is that the following will cause a compilation error, even if you disable the `nightly` feature. The reason is that this syntactic construction errors during the parsing stage, even before the compiler gets to handling attributes.

```
#[cfg(feature = "nightly")]
unsafe impl ZerosValid for .. {}
```

PR #733, opened 2018-01-06
[WIP] Hgignore support

kalekseev:
Hi @BurntSushi, last time you said that you would accept hg support with minimal feature set support https://github.com/BurntSushi/ripgrep/issues/6/ so during the holidays I refactored my original pull request https://github.com/BurntSushi/ripgrep/pull/477.

Now globset is unchanged, everything live in hgignore module therefore current code shouldn't be affected by merging in hgignore support.

The places that I think should be fixed before merge are marked by FIXME, please review the code and tell me what do you think, is this the right direction to move?


ignore/src/hgignore.rs
```diff
@@ -0,0 +1,513 @@
...
+        let original = line.to_string();
+
+        let mut literal_separator = false;
+        let has_slash = line.chars().any(|c| c == '/');
+        let mut is_absolute = false;
+        if line.starts_with("\\#") {
+            line = &line[1..];
+            is_absolute = line.chars().nth(0) == Some('/');
+        } else {
+            // FIXME: this is not the case for hgignore
```
kalekseev: Can anyone describe what leading slash mean in regular glob or especially in hgignore? I can't find a case that would match glob with leading slash in hgignore, maybe just skip them?

BurntSushi:
@kalekseev I just wanted to quickly chime in and say thanks so much for this PR! It might take me a bit to review though. :)

ignore/src/hgignore.rs
BurntSushi: I think `man gitignore` says it best:

> A leading slash matches the beginning of the pathname. For example, "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".

It other words, it's a form of an anchor.

BurntSushi: (I don't know whether it's a thing in `hgignore` or not.)

kalekseev: > I don't know whether it's a thing in hgignore or not.

```
$ cat > .hgignore
 syntax: glob
 /foo
 bar

$ touch foo
$ touch bar
$ hg status
 ? .hgignore
 ? foo

$ mkdir xxx
$ touch xxx/foo
$ hg status
 ? .hgignore
 ? foo
 ? xxx/foo
```

I've tested it and haven't found a case where leading slash would match anything, I believe git uses leading slash because standard glob doesn't support leading slash at all. That's why I assumed that we can skip such rules.

Globs in .hgignore are not rooted if you want rooted glob you should use `glob:`, from the `hg help patterns`

```
To use an extended glob, start a name with "glob:". Globs are rooted at
    the current directory; a glob such as "*.c" will only match files in the
    current directory ending with ".c".
```

kalekseev:
Hi @BurntSushi, I will have a couple of holidays at the beginning of May and I want to finish this pr, could you briefly review it or just comment what do you think about these two places:
 
1) Here I need the path https://github.com/BurntSushi/ripgrep/pull/733/files#diff-90bb8ac326bac2bf846c8c831a2abba9R180 as in Candidate 
 https://github.com/BurntSushi/ripgrep/pull/733/files#diff-258f99315c5f915cf0fdea294e12a8efR480 should I copy methods from globset::pathutils into ignore::pathutils to bulid that path?

2) We have to skip lines with invalid regex. I use regexset and I think maybe we can build regex in `add_regex_line` method to be sure that it's valid or this is too bad for performance? 
https://github.com/BurntSushi/ripgrep/pull/733/files#diff-90bb8ac326bac2bf846c8c831a2abba9R369

kalekseev:
If anyone subscribed to that pr then please test current version with your mercurial setup.
```
$ git clone git@github.com:kalekseev/ripgrep.git
$ cd ripgrep
$ git checkout hgignore
$ cargo build --release
$  cp target/release/rg ~/.local/bin/rg # change to your PATH dir
```
Things that are not implemented (skipped by ripgrep):
 - Extended mercurial patterns: path, rootfilesin, listfile, listfile0, include, subinclude
 - python specific regexes (that are not valid rust regexes)

BurntSushi:
@kalekseev I hate to close this out, and I'm very appreciative of your hard work on this feature, but after thinking about this for a long time, I no longer believe its viable for ripgrep to support Mercurial. I left a longer explanation here: https://github.com/BurntSushi/ripgrep/issues/6#issuecomment-406812551

PR #734, opened 2018-01-07
trivial: update instructions for Fedora

ignatenkobrain:


README.md
```diff
@@ -191,18 +191,24 @@ If you're a **Gentoo** user, you can install `ripgrep` from the [official repo](
 $ emerge sys-apps/ripgrep
 ```
 
+If you're a **Fedora 28+** user, you can install `ripgrep` from [copr](https://copr.fedorainfracloud.org/coprs/carlwgeorge/ripgrep/):
+
+```
+$ sudo dnf install ripgrep
```
BurntSushi: How come we don't need the `sudo dnf copr enable carlwgeorge/ripgrep` line here?

ignatenkobrain: because it is in official repos for quite some time (but only F28+) already

ignatenkobrain:
```
➜  ~ sudo docker run --rm registry.fedoraproject.org/fedora:rawhide sh -c 'dnf -y install ripgrep -q && rg --version' 
ripgrep 0.7.1
-AVX -SIMD
```

ignatenkobrain: I'm planning to backport it to Fedora 27 as well, but yet don't have time for it..

BurntSushi: Oh I see. I think I got confused since both instructions talk about a COPR, which I just assumed was similar to something like adding a new PPA in Ubuntu. (Unfortunately, I am hopelessly unfamiliar with both Ubuntu and Fedora.) Anyway, thank you!

PR #741, opened 2018-01-10
Add `--passthru` option

### LINKED ISSUE #740: Add support for `--passthr{u,ough}` flag
After reading [this comment](https://news.ycombinator.com/item?id=16097948) on HN, I thought it'd be neat to add support for the `--passthrough` / `--passthru` flag in ripgrep as well.

Thoughts?

okdana:
Fixes #740 

I ran into one issue that i didn't consider, which is that `--replace` won't work properly with this option (since it treats zero-width matches the same as any others). That's kind of a shame i think (if for no other reason than it would make the unit tests more accurate), but i suppose people can use `sed` if they really need that functionality.

I made it override with `--count` since it seemed relatively safe and intuitive to me. Don't feel strongly about it though.

Also, in hind sight maybe it's not worth mentioning that it's equivalent to adding a `^` pattern, since it's just an implementation detail that isn't really useful to anyone. Let me know if you agree and i'll remove it.

Some of the changes here might conflict with #728, but only slightly if so.


src/app.rs
```diff
@@ -166,6 +166,9 @@ pub fn app() -> App<'static, 'static> {
         .arg(flag("no-ignore-vcs"))
         .arg(flag("null").short("0"))
         .arg(flag("only-matching").short("o"))
+        .arg(flag("passthru").alias("passthrough")
+            .conflicts_with_all(&["only-matching", "replace"])
+            .overrides_with("count"))
```
BurntSushi: Hmm I kind of feel like `--passthru` should conflict with `--count`, no? They are somewhat mutually incompatible flags. I could be convinced though! Maybe explain this a bit more?

I agree with the other conflicts. Thanks for thinking about that!

doc/rg.1.md
```diff
@@ -278,6 +278,11 @@ Project home page: https://github.com/BurntSushi/ripgrep
 : Print only the matched (non-empty) parts of a matching line, with each such
   part on a separate output line.
 
+--passthru, --passthrough
+: Show both matching and non-matching lines. This is equivalent to adding ^ to
+  the list of search patterns. This option overrides --count and cannot be used
+  with --only-matching or --replace.
```
BurntSushi: Yeah I agree that we shouldn't mention `^`. I could see the implementation changing for example to support `--replace` by implementing this flag differently. (e.g., By not modifying the regex.)

src/app.rs
BurntSushi: Hmmmm... I am conflicted. But you're right, we already have this problem today with other flags. I'm fine with the behavior in this PR, and adding `!self.is_present("count")` sounds good.

okdana:
Updated

PR #742, opened 2018-01-11
termcolor: Write `Ansi::reset()` using `\x1b[0m` instead of `\x1b[m`

kennytm:
AppVeyor CI supports coloring the console through ANSI escape codes, however its support is incomplete (appveyor/ci#1824). One particular issue is that `\x1b[m` is not recognized as the same as `\x1b[0m`. 

This caused colored output using this library extremely ugly on AppVeyor as the color is never reset ([example](https://ci.appveyor.com/project/rust-lang/rust/build/1.0.6004/job/a23aftwu88yh35c4), start from line 675).

While technically an AppVeyor bug, I think changing the code here would improve compatibility anyway. (That issue is also untouched for 2 months so I doubt if AppVeyor is going to fix it soon.)

### DIFF
--- a/termcolor/src/lib.rs
+++ b/termcolor/src/lib.rs
@@ -973,7 +973,7 @@ impl<W: io::Write> WriteColor for Ansi<W> {
     }
 
     fn reset(&mut self) -> io::Result<()> {
-        self.write_str("\x1B[m")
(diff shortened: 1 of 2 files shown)


BurntSushi:
Thanks!

PR #744, opened 2018-01-12
Add Ubuntu installation guide to README

cstorres:
I added documentation for Ubuntu users who want to install `ripgrep` using the snap tool

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,7 +228,7 @@ $ # (Or using the attribute name, which is also `ripgrep`.)
 If you're an **Ubuntu** user, `ripgrep` can be installed from the `snap` store.
 * Note that if you are using `16.04 LTS` or later, snap is already installed.
 * For older versions you can install snap using
-[this guide](https://docs.snapcraft.io/core/install-ubuntu)
+[this guide](https://docs.snapcraft.io/core/install-ubuntu).
 
 ```
 sudo snap install rg
--- a/README.md
+++ b/README.md
@@ -225,6 +225,12 @@ $ nix-env --install ripgrep
 $ # (Or using the attribute name, which is also `ripgrep`.)
 ```
 
+If you're an **Ubuntu** user and you are using `16.04 LTS` or later, you can install `ripgrep` from the `snap` store:
+
+```
+sudo snap install rg
+```
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -225,6 +225,12 @@ $ nix-env --install ripgrep
 $ # (Or using the attribute name, which is also `ripgrep`.)
 ```
 
+If you're an **Ubuntu** user and you are using `16.04 LTS` or later, you can install `ripgrep` from the `snap` store:
```
BurntSushi: Could you please wrap this to 79 columns inclusive like the rest of the README?

Also, do Ubuntu users need to install `snap` themselves, or is it on 16.04 by default?

Finally, could you remove the reference to Ubuntu above? (Since we know show how to install it.)

cstorres: Yes of course. I wrapped the lines to 79 and add some clarifications for the Ubuntu versions. `snap` is installed by default in Ubuntu 16.04 and later versions.

What I didn't understand is what do you mean with this:
> Finally, could you remove the reference to Ubuntu above? (Since we know show how to install it.)

Thanks for the feedback! :)

BurntSushi: Just search for mentions of `Ubuntu`. The README says that Ubuntu isn't mentioned below, but this PR adds instructions for Ubuntu, so that section is now outdated. :-)

README.md
```diff
@@ -225,6 +225,15 @@ $ nix-env --install ripgrep
 $ # (Or using the attribute name, which is also `ripgrep`.)
 ```
 
+If you're an **Ubuntu** user, `ripgrep` can be installed from the `snap` store.
+* Note that if you are using `16.04 LTS` or later, snap is already installed.
+* For older versions you can install snap using
+[this guide](https://docs.snapcraft.io/core/install-ubuntu)
```
BurntSushi: Could you please add a period to the end of this line? Thanks.

BurntSushi:
@cstorres Thanks! And thanks for your patience with my nits. :)

cstorres:
I should thank you for letting me contribute with your project :) your nits
were perfectly reasonable, so I can't complain :)

PR #748, opened 2018-01-13
yarn.lock is not YAML

### LINKED ISSUE #747: Is yarn.lock really yaml
hey there
under filetypes `DEFAULT_TYPES` `yarn.lock` is listed as `yaml`. I don't think that is correct.

okdana:
Fixes #747

Why *not* invent your own data-serialisation format?

🤷‍♀️


BurntSushi:
Thanks for doing this so quickly! :-)

PR #751, opened 2018-01-14
[WIP] Add support for compressed file formats (closes #539)

### LINKED ISSUE #539: add decompression support on Unix systems
@wavexx [Suggests farming out decompression to established stream decompressors](https://github.com/BurntSushi/ripgrep/issues/225#issuecomment-312873930). I am somewhat attracted to this path since it will cover a lot of use cases, and at least on the face of things, seems relatively easy to implement. If this turns out to be a much larger project than I anticipated, then the trade off may no longer make sense. The intention is to support decompression on all supported platforms, so we don't want to invest a bunch of work in a Unix-only option.

I'm not sure when I personally will have time to work on this, but as long as it's easy to maintain, I'm not intrinsically opposed to farming this out to another process. Note that the core search code can work on *any* arbitrary thing that implements [`io::Read`](https://doc.rust-lang.org/std/io/trait.Read.html), which means the glue code required for this should be pretty limited. There's already an example of this at play for encoding support, where any supported encoding is first transcoded to UTF-8 before being searched. The scope of that addition was pretty much limited to [implementing `io::Read`](https://github.com/BurntSushi/ripgrep/blob/master/src/decoder.rs) and doesn't require touching any of the core search code.

If someone else wanted to work on this, I'd be happy to mentor it. I haven't thought about it too much, but I think these are probably the high level steps one needs to take:

* Come up with a way of detecting when a certain type of decompressor is needed. It's probably sufficient to use the file extension as a first pass, although one should consider special cases like `.tar.gz` (probably by skipping them at first, but maybe eventually trying to read `tar` archives as well).
* Implement glue code that takes as input a command that executes stream decompression on a specific file path and outputs an implementation of `io::Read` that is suitable to pass to the `io::Read` searcher.
* Think about what a path forward looks like for replacing out-of-process decompression with in-process decompression. For example, if and when Rust gets a native gzip decompression library that is suitable for use (maybe it already is), then how could it be used to replace use of, say, the `gzip` command seamlessly? This will be important for eventually migrating to cross platform decompression support.
* Give some thought to Windows support. It doesn't actually seem like this is inherently Windows-incompatible, but may favor Unix systems by virtue of the tooling typically available in a Unix environment.
* Initially ignore any and all decompression on `stdin`. Callers can invoke their own decompressor in this case easily enough that we shouldn't do anything smart. (A future path might add a flag to force decompression of a specific format.)

balajisivaraman:
Creating this PR for review purposes. I still haven't added any tests for the newer code, which is one of the main things still pending here.

Notes:
- Had to add `globset` as dependency for `ripgrep` module to easily check for supported file formats. The in-built `path.extension()` method did not allow me to verify for `*.tar.gz` files. 😞 
- We could've used [`output`](https://doc.rust-lang.org/std/process/struct.Command.html#method.output) from Command instead of [`spawn`](https://doc.rust-lang.org/std/process/struct.Command.html#method.spawn), which pipes stdout and stderr by default and gives us access to the process' exit status. However, it would've meant loading the entire file into memory in the form of a `Vec<u8>`. I've therefore gone with the `spawn` approach here, coupled with peeking at the `stderr` to verify whether the command succeeded or failed. It works nicely enough in practice in my testing.
- Also, I noticed that the `spawn` method ends up returning an `Err` immediately if the command is not found. All other errors are obtained from the process' stderr. That's why I've handled it such that if `spawn` returns an `Error`, then `rg` will output the appropriate debug message.
- I couldn't think of any other CLI flag that conflicts with `-z/--search-zip`. One thing to think of is how should `rg` handle when the `-a` argument is provided? Right now, if `-a` and `-z` are both provided, `rg` will search zip files as per the new code, but will also end up treating it as a binary file and try searching it. Are we OK with this?
- Not really fond of passing the `no-messages` flag to the `decompressor`, but there's no better option if we want all error handling to remain within the module itself. In this way, the `worker` module remains largely unaffected if we choose to change how we do decompression at a later point in time.

Any other thoughts and suggestions are welcome.


complete/_rg
```diff
@@ -87,6 +87,7 @@ _rg() {
     '(-w -x --line-regexp --word-regexp)'{-w,--word-regexp}'[only show matches surrounded by word boundaries]'
     '(-e -f --file --files --regexp --type-list)1: :_rg_pattern'
     '(--type-list)*:file:_files'
+    '*'{-z,--search-zip}'search through zip files'
```
okdana: The exclusions and the syntax itself are wrong here. At a minimum it should be something like this:

```
'(-z --search-zip)'{-z,--search-zip}'[search through zip files]'
```

We might also want to add other exclusions, but i'd like to do another pass on this file within the next few weeks, so it can wait until then to get fancier.

I would also suggest changing the description wording to something like *search contents of compressed files* — 'through' is a bit ambiguous, and we don't actually search *zip* files at all.

doc/rg.1
```diff
@@ -184,6 +184,15 @@ Only show matches surrounded by line boundaries.
...
 .RS
 .RE
+.TP
+.B \-z, \-\-search\-zip
+Search through compressed files.
+Currently gz, bz2, xz and lzma formats are supported.
+.RS
+.PP
+Note that rg expects to find the decompression binaries for the
+respective formats on your system\[aq]s PATH for use with this flag.
```
okdana: 'On your `PATH`' feels strange to me — '*in* your `PATH`'?

src/app.rs
```diff
@@ -583,6 +584,11 @@ lazy_static! {
               only clears the default type definitions that are found inside \
               of ripgrep.\n\nNote that this MUST be passed to every \
               invocation of ripgrep. Type settings are NOT persisted.");
+        doc!(h, "search-zip",
+             "Search through compressed files.",
+             "Search through compressed files. Currently gz, bz2, xz, and \
+              lzma files are supported. This options expects the uncompression \
```
okdana: Same comment about 'on your `PATH`', but also i think 'uncompression' should be 'decompression' here

okdana:
There is a problem with the completion function, and i added a few bike-shed wording remarks. I won't comment on the Rust stuff except to say that it does at least make sense to me.

I tested the patch on macOS (just briefly) and it works! This is neat as hell, ty!

complete/_rg
balajisivaraman: I'll fix the syntax as per your suggestion. This file is always throwing me off. Sorry about that.

I thought of using `-C/--search-compressed` instead of `-z/--search-zip`. The only reason I went with the latter is that it would be familiar to folks coming from using `ag`. Like you said, `search-zip` doesn't really make a lot of sense here. Should we just go with the former option?

okdana: The completion-function syntax takes a while to get used to, yeah. Every few weeks i learn that something i used to do in these files is wrong, lol.

I like `-z` for the short option. It's what `ag` and `sift` both use, and the connection with 'Ziv' is immediately intuitive. As far as the long option, idk. `--search-zip` feels slightly inaccurate, but you're right about `ag` — it's one of the most common recursive file searchers, and definitely the most well known one that supports searching compressed files, so it's not unreasonable to borrow the terminology. `--search-compressed` makes sense too tho, w/e 🤷‍♀️

src/decompressor.rs
```diff
@@ -0,0 +1,117 @@
...
+    /// Create a new decompress command
+    fn new(cmd: &'static str, args: Vec<&'static str>) -> DecompressionCommand {
+        DecompressionCommand {
+            cmd, args
+        }
+    }
+}
+
+lazy_static! {
+    static ref DECOMPRESSION_COMMANDS: HashMap<&'static str, DecompressionCommand> = {
```
tamird: why isn't this an enum?

better yet, can't this stuff be done without spawning subprocesses? i imagine there are rust libraries for most of these.

okdana: The rationale for this implementation is explained in #225 and #539

There are still very few pure-Rust (de)compression libraries AFAIK

src/decompressor.rs
```diff
@@ -0,0 +1,116 @@
+use std::ffi::OsStr;
+use std::io::Read;
+use std::path::Path;
+use std::collections::HashMap;
+use std::process::Command;
+use std::process::Stdio;
+use std::process::ChildStdout;
+use globset::{Glob, GlobSet, GlobSetBuilder};
```
BurntSushi: Imports should be consistent with the rest of the code. They should be grouped in three blocks, where each block is separated by whitespace. The first block are `std` imports. The second block are external crate imports. The third block are internal crate imports. Each block should be sorted lexicographically.

src/decompressor.rs
```diff
@@ -0,0 +1,116 @@
...
+
+/// A decompression command, contains the command to be spawned as well as any
+/// necessary CLI args
+struct DecompressionCommand {
+    cmd: &'static str,
+    args: Vec<&'static str>
```
BurntSushi: Since we are building these values in a `lazy_static!`, we don't actually need the `Vec`. We can use an `&'static [&'static str]` instead.

src/decompressor.rs
```diff
@@ -0,0 +1,116 @@
...
+
+/// Returns a handle to the stdout of the spawned decompression process for
+/// `path`, which can be directly searched in the worker.
+///
+/// If there is any error in spawning the decompression command, then return
+/// `None`, after outputting any necessary debug or error messages.
+pub fn get_reader(path: &Path, no_messages: bool) -> Option<ChildStdout> {
+    if is_tar_archive(path) {
+        debug!("Tar archives are currently unsupported: {}", path.display());
+        None
```
BurntSushi: In general, I like to keep control flow straight-forward and avoid gratuitous nesting. Early returns are very useful for this. For example, this code could be written as:

```rust
if is_tar_archive(path) {
    debug!(...);
    return None;
}

let extension = ...;
```

That way, you don't need to indent the entire rest of the function.

src/decompressor.rs
```diff
@@ -0,0 +1,116 @@
...
+                if error.is_empty() {
+                    process.stdout
+                } else {
+                    if !no_messages {
+                        eprintln!("Error occurred while trying to decompress path: \
+                                   {}: {}", path.display(),
+                                  String::from_utf8_lossy(error.as_slice()));
+                    }
+                    None
+                }
```
BurntSushi: OK, so the issue here is that this will deadlock in non-trivial cases. You might not have hit it if you only tested on small files. Basically the issue here is that if you don't read from stdout, then the process you spawned (if it's implemented sanely) will at some point block on writing to stdout because it is connected to a pipe. If the calling process never reads from stdout, then the spawned process will eventually block. If the calling process is in turn blocked on something else from the spawned process---which in this case, we are, because we're trying to read from `stderr` until the spawned process stops writing to it---then you wind up with a deadlock.

If you only test this on small files then you won't notice this because the spawned process will likely be using buffered output. And since it's connected to a pipe, the buffer is usually biggish (~4K?) as opposed to being line buffered on a tty. As a result, the spawned process will finish doing its thing before ever flushing its buffer, which gives the appearance of code that works. :-)

The "fix" to this is to couple the `child` with reading stdout and wait to read stderr until the process terminates. This basically means that you can't return `ChildStdout` directly. You need to return something else that wraps up this logic.

BurntSushi: As an additional upside, the reader that you return can return normal error values, which are handled by the worker. This means you don't need to thread `no_messages` through the code.

tests/tests.rs
```diff
@@ -1609,6 +1609,94 @@ clean!(suggest_fixed_strings_for_invalid_regex, "foo(", ".",
...
+    let mut cmd = wd.command();
+    cmd.arg("-z").arg("Sherlock").arg("sherlock.gz");
+
+    wd.assert_non_empty_stderr(&mut cmd);
+
+    let output = cmd.output().unwrap();
+    let err = String::from_utf8_lossy(&output.stderr);
+    assert_eq!(err.contains("not in gzip format"), true);
+}
+
```
BurntSushi: Fantastic tests, thank you. :-)

BurntSushi:
I merged this PR in #767. Thanks again @balajisivaraman!

src/decompressor.rs
balajisivaraman: Ah, I knew it. 😞  When I made this change and looked at the code, I knew there had to be something I was missing.

Thank you for the detailed write up on the issues with my code. It makes a lot of sense to me and helps me understand it better. It also gives me incentive to do some more reading up on pipes and child processes, which I will. 👍

src/decompressor.rs
balajisivaraman: Ah, that makes a lot of sense. I did do this in the `worker.rs` file but missed it here. 👍

balajisivaraman:
@BurntSushi, Thank you for merging this. It is one of my first significant contributions of any kind to a OSS project. I learned a lot from working on it as well as going through the changes you made on top of it. Cheers. :-)

BurntSushi:
@balajisivaraman It's a helluva first contribution! It will be a highlight feature in the next release. :)

PR #752, opened 2018-01-16
Add GN as new filetype

Leandros:
GN is Googles new build system, originally developed to build chromium. They use the *.gn and *.gni file endings.


BurntSushi:
Thanks!

PR #753, opened 2018-01-16
Add *.hxx to the cpp filetype

Leandros:
.hxx is commonly used as a header extension for C++ files, since .cxx was already added, I was free to add the header extension to match the implementation file.


BurntSushi:
Thanks!

PR #768, opened 2018-01-30
update bytecount to improve perf with current rustc

llogiq:
This incorporates the simplification @Veedrac made, because rustc apparently changed codegen. It does *not* include any additional dependencies.


BurntSushi:
@llogiq Thanks! I think you probably want to run `cargo update -p bytecount` as well in order to bump the lockfile.

llogiq:
Thanks, amended.

Issue #553, opened 2017-07-13
Don't raise an error when a choose-one flag is specified repeatedly


sarahgerweck:
There is currently no way to set custom defaults for Ripgrep. The recommended practice is to set an alias instead. However, this is somewhat thwarted by the fact that Ripgrep is too aggressive in validating that you never repeat yourself. This also has other consequences.

E.g., if you prefer smart case matching, you can make an alias like this:

```
alias rg='rg -S'
```

This _mostly_ works, but it means that I now get an error if I ever execute any commands that include `-S` in them. Ripgrep give the error below:
```
error: The argument '--smart-case' was provided more than once, but cannot be used multiple times
```

This is undesirable behavior both when writing scripts and when trying to use aliases as defaults. Both scripts and aliases face the same problem: building up a complex command line by nesting contexts. This rule means that this process needs to be globally aware of all the outer contexts.

I think it is much more common and more user friendly that commands like this use the last one specified for flags that are mutually exclusive, and don't raise errors if you specify the same flag twice. Otherwise, trying to build nontrivial command lines from variables or aliases is somewhat a nightmare.

Fortunately, it at least allows `rg -S -i`. I think this is a good example of why raising an error for `rg -S -S` is undesired behavior. The `rg -S -i` case is *more* overspecified than the `rg -S -S` case. Ultimately, the current rule doesn't actually protect you from any undesired behavior, but it does prevent useful use cases.

BurntSushi:
Right, agreed. My feeling is that this should get lumped into #196 (where I also left a [comment explaining some things just now](https://github.com/BurntSushi/ripgrep/issues/196#issuecomment-314943495)), although I can appreciate that these are really two orthogonal issues. It just so happens that persistent config is blocked on it. But we can make your situation better without needing to finish persistent config, and this issue is thorny enough that I think giving it its own ticket is a good idea.

As a possible work-around until this is fixed properly, if you prefix your command with `\` (e.g., `\rg -S pattern`), then your shell should ignore any aliases you have set up and invoke the command directly.

> This is undesirable behavior both when writing scripts and when trying to use aliases as defaults. Both scripts and aliases face the same problem: building up a complex command line by nesting contexts. This rule means that this process needs to be globally aware of all the outer contexts.

[...]

PR #772, opened 2018-02-04
add support for persistent config

BurntSushi:
This PR fixes both #196 and #553. The commit messages have most of the details (including some goodies like a 2x decrease in release build time), but the TL;DR is that ripgrep now recognizes a `RIPGREP_CONFIG_PATH` environment variable. When set and non-empty, ripgrep will read shell arguments from it (one per line) and prepend them to whatever explicit CLI arguments have been given. A new flag, `--no-config`, prevents any sort of configuration from being loaded from a file.

This does not add any support for auto-loading config files from pre-determined locations like one's `$XDG_CONFIG_HOME` directory. [It was decided](https://github.com/BurntSushi/ripgrep/issues/196#issuecomment-362850321) to punt on that for now, and one wonders how far we can get with just the env var approach!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,14 +68,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "env_logger"
-version = "0.4.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-dependencies = [
- "log 0.3.9 (registry+https://github.com/rust-lang/crates.io-index)",
-]
-
-[[package]]
 name = "fnv"
 version = "1.0.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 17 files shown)


okdana:
To fix the completion-function check you can add this to `_rg` (around line 55):

```
"--no-config[don't read configuration files]"
```

BurntSushi:
@okdana Yup! Figured it out. :-) I'm quite happy that CI caught that! I completely forgot we even had the test.

hoodie:
this is a neat feature, but it falls down if you install from crates.io:
```
ripgrep 0.8.0 (rev )
```

BurntSushi:
@hoodie Thanks, I created #789. It is low priority though.

BatmanAoD:
@BurntSushi Just wanted to say that 0.8 is a great update; I just migrated to Windows 10 so it's great to have native colors in `cmd`, and the persistent config is very nice. Thanks for the awesome tool!

PR #775, opened 2018-02-05
fixes a PATTERN->pattern typo in argument conflicts

kbknapp:
Found this while testing kbknapp/clap-rs#868


BurntSushi:
Lovely! Thanks for finding this. :-)

okdana:
Oops

BurntSushi:
@okdana This one was actually my fault, not yours! I switched the identifier name to `pattern` (was `PATTERN`) but kept the argument value name as `PATTERN` (which is what is shown in the `--help` output) and forgot to update a place that was using `PATTERN`. :-)

PR #783, opened 2018-02-09
Ignore (one) trailing line break per --file parameter

mqudsi:
Currently ripgrep interprets a trailing line break at the end of a
pattern file specified via `--file` or similar as a global "match-all"
pattern. Given that many editors are configured to automatically add a
trailing new line at the end of a file on save, this is not necessarily
a good thing.

This patch removes a single new line from the end of each pattern file
contents by simply dropping an empty rule from the pattern vector if it
is the last rule in the vector after loading rules from a file. Multiple
line breaks are not ignored, and no overhead is added by omitting this
check from within the for loop (not that it would really matter since
this is just init time). Additionally, line breaks not at the end of the
file are similarly kept.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -511,6 +511,9 @@ impl<'a> ArgMatches<'a> {
                     for line in io::BufReader::new(f).lines() {
                         pats.push(self.str_pattern(&line?));
                     }
+                    if Some(&self.str_pattern("")) == pats.last() {
+                        pats.pop();
+                    }
                 }
             }
         }
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -1879,6 +1879,22 @@ fn files() {
             || lines == path("dir/file\nfile\n"));
 }
 
+// See: https://github.com/BurntSushi/ripgrep/pull/783
+#[test]
+fn ignore_trailing_line() {
+    let wd = WorkDir::new("ignore_trailing_line");
+    wd.create("foo", "foo");
+    wd.create("bar", "bar");
+    wd.create("patterns.txt", "foo\n\n");
+
+    let mut cmd = wd.command();
+    cmd.arg("--file").arg("patterns.txt")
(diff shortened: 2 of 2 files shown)


src/args.rs
```diff
@@ -511,6 +511,9 @@ impl<'a> ArgMatches<'a> {
                     for line in io::BufReader::new(f).lines() {
                         pats.push(self.str_pattern(&line?));
                     }
+                    if Some(&self.str_pattern("")) == pats.last() {
```
mqudsi: This can be replaced with `while Some(....)` to make it drop all trailing line breaks.

mqudsi:
@BurntSushi I did, but to be honest, I probably didn't do it the way you would have. I tried to just copy the syntax from a nearby test.

BurntSushi:
Hmm... OK, so it occurred to me that these are exactly the kinds of cases where it is good to be consistent with other tools. grep supports this flag, and it actually looks like ripgrep's current behavior is consistent with grep's behavior. To me, this is a subtle enough of a change that it makes more sense to be consistent with existing tools instead of trying to be smarter.

Apologies for asking you to write a test and then rejecting the PR, but the test does look good! Thank you!

mqudsi:
No problem. Perhaps I'm missing something however - is there every a time where you'd want a match-anything rule (loaded from a file, at that)?

BurntSushi:
@mqudsi Possibly. At least for GNU grep, it might be useful as a trick to show every live and color only those things which match. ripgrep does have a `--passthru` flag (recently added) that lets you explicitly do this. It's also consistent with the fact that the empty file will match every pattern.

With that said, even if there weren't a compelling use case (and honestly, the above isn't particularly compelling to me), then matching existing behavior with well established tools in subtle cases like this is important for its own sake.

mqudsi:
Cheers!

PR #787, opened 2018-02-11
Add ?Sized bound to WriterColor for &'a mut T

Marwes:
Just noticed this while writing a ?Sized bound when using termcolor.


BurntSushi:
Thanks!

PR #794, opened 2018-02-13
ignore: add support for VHDL type

unsignedint:
VHDL is a very popular language for hardware design for FPGA etc. I think this would be useful for others :-)


BurntSushi:
Thanks!

PR #799, opened 2018-02-13
[WIP] Add --stats flag and implement basic stat tracking (closes #411)

### LINKED ISSUE #411: Way to report total count of matches?
Is there a way to print the total count of matches? I can pipe the output to a line counter, but then I don't get the actual matches printed.

balajisivaraman:
Creating this PR for review purposes and as a WIP thread for adding the `--stats` flag to ripgrep.

- Implement tracking for basic stats (no. of matches, files with matches, files searched, and time taken)
- Ensure `--stats` is ignored when `--files`, `--files-with-matches` or `--files-without-match` are passed
- Implement tracking of bytes searched for `--no-mmap` searches using the `counting_reader` trick. This will require changing the worker's `run` method to return an Optional `bytes_searched` attribute along with the existing `match_count`. We also need to ensure this happens only when `--stats` is passed.

Other thoughts:

- Have used the default `Duration` implementation from `std` as did not want to pull in any other dependencies such as [`time`](https://crates.io/crates/time) or [`chrono`](https://crates.io/crates/chrono), and this served the purpose very well.
- Have defaulted the time to always display in seconds since this is what `ag` does as well. `rg` also prints only 3 numbers after the decimal point, as opposed to `ag`'s 6.

  The only reasoning behind this is that Rust only allows us to control the digits printed after the decimal point, not overall. If we run a very long search, we don't want to print something like `2751.648931 seconds`. So have chosen the nicer option of 3 digits after the decimal point.

  I think `ag` prints 6 digits overall, with the number of digits printed after the decimal point depending on the overall width of the string. (This is the behaviour I've observed; have not verified this with the code.)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -557,6 +557,7 @@ pub fn all_args_and_flags() -> Vec<RGArg> {
     flag_search_zip(&mut args);
     flag_smart_case(&mut args);
     flag_sort_files(&mut args);
+    flag_stats(&mut args);
     flag_text(&mut args);
     flag_threads(&mut args);
     flag_type(&mut args);
@@ -1444,6 +1445,23 @@ This flag can be disabled with --no-sort-files.
     let arg = RGArg::switch("no-sort-files")
         .hidden()
         .overrides("sort-files");
+    args.push(arg);
+}
+
+fn flag_stats(args: &mut Vec<RGArg>) {
+    const SHORT: &str = "Print statistics about this ripgrep search.";
+    const LONG: &str = long!("\
+Print statistics about this ripgrep search. When this flag is present,
+ripgrep will print the following stats at the end of the search: total
+no. of matches, no. of files with matches, no. of files searched, and
+the time taken for the entire search to complete.
+
+Note that this flag has no effect if --files, --files-with-matches or
+--files-without-match is passed.");
+
+    let arg = RGArg::switch("stats")
+        .help(SHORT).long_help(LONG);
+
     args.push(arg);
 }
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,6 +27,7 @@ use std::sync::Arc;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::mpsc;
 use std::thread;
+use std::time::{Duration, Instant};
 
 use args::Args;
 use worker::Work;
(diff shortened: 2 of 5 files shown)


balajisivaraman:
I'll close this PR as the current implementation is simply wrong. When I looked at the code for this originally, I overlooked the fact that the searchers return matching line count and not the individual match count, which is what we want for the `--stats` flag. My apologies for that.

I'll work on the proper implementation, which I expect to be a bit more involving than this one. In the meantime, I don't want this clogging up the PR space here.

src/app.rs
```diff
@@ -1447,6 +1448,23 @@ This flag can be disabled with --no-sort-files.
...
+fn flag_stats(args: &mut Vec<RGArg>) {
+    const SHORT: &str = "Print statistics about this ripgrep search.";
+    const LONG: &str = long!("\
+Print statistics about this ripgrep search. When this flag is present,
+ripgrep will print the following stats at the end of the search: total
+no. of matches, no. of files with matches, no. of files searched, and
+the time taken for the entire search to complete.
+
+Note that this flag has no effect if --files, --files-with-matches or
+--files-without-match is passed.");
```
BurntSushi: Can you put the `");` on its own line please?

src/app.rs
```diff
@@ -1447,6 +1448,23 @@ This flag can be disabled with --no-sort-files.
...
+
+    let arg = RGArg::switch("stats")
+        .help(SHORT).long_help(LONG);
+
```
BurntSushi: Remove this blank line please.

src/main.rs
```diff
@@ -373,6 +395,15 @@ fn eprint_nothing_searched() {
                Try running again with --debug.");
 }
 
+fn print_stats(match_count: u64, paths_searched: u64, paths_matched: u64, time_elapsed: Duration) {
```
BurntSushi: All lines of source code in ripgrep should be wrapped to 79 columns inclusive (excepting URLs or other things that are inconvenient to wrap). Please see examples of other function declarations for formatting tips.

balajisivaraman:
@BurntSushi, I wouldn't mind getting this PR merged at all. :-)

The only reason why I closed this was I'm going to begin work on the full implementation of it with bytes searched and individual count, so I thought we wouldn't want this anyway.

But seeing as that is more involving change with multiple moving parts, it might take some time. So it makes sense to go ahead with this partial implementation in the meanwhile.

I'll update the PR to reflect your comments. Thanks! :-)

BurntSushi:
@balajisivaraman This was great work, thanks so much! I merged this in https://github.com/BurntSushi/ripgrep/commit/00520b30f5f38e543e17b1a4cc5e8417bc488ea4 with no real changes other than fixing up style (lines should be limited to 79 columns inclusive) and resolving conflicts. I also tweaked the wording of output slightly to indicate that total matches actually corresponds to the total number of matched lines. (So we leave the door open to emit another statistic that reports the total matches.)

balajisivaraman:
@BurntSushi, Thank you for merging this. And my apologies for not updating this PR. I was waiting for the other PR (`match_line_count`) to get merged so I could update it properly and completely forgot about it. Thanks again. 😄

Issue #359, opened 2017-02-12
json output option to allow scripting


timotheecour:
Could we have a json output option?
use case: making it easier to integrate in other tools [1]
eg: 

```
rg '\bWORLD\b' $args --output=json
{"path":"foo/bar.txt","line":3,"offset":258,"column":4,"length":5,"match":"WORLD" ,"context":"hello WORLD !" }
 ```
offset corresponds to byte offset in file (useful for efficient access for the other tool)
length is in bytes (in case of non-ascii)

The alternative (manual parsing) is error prone, eg:
* if filename is `foo/bar:13:foo.txt` (contains `:`)
*  if the escape codes colorizing the matched word are hard to find because the context contains some escape codes itself (eg it can actually happens if it's a log file)
* relying on --column doesn't work for multiple matches in 1 line (ok just saw --vimgrep but doens't show colors, so we can't get the extent of each match)

[1] NOTE: one of these other tools is multiline search built on top of ripgrep

BurntSushi:
I think what you're asking for is "structured output," where JSON might be your preference. In that case, I think this is a dupe of #244. To be clear, I do *not* want to be in the business of producing a bunch of different output formats because there is so much complexity involved. (Every option that influences output has to be considered against the new format.)

> NOTE: one of these other tools is multiline search built on top of ripgrep

While I'm in favor of adding some type of structured output to ripgrep, I would rather see these types of tools built with [libripgrep](https://github.com/BurntSushi/ripgrep/milestone/1) than with ripgrep itself.

timotheecour:
since these structured output (json or not, doesn't matter) will be meant  for machine consumption, it would make sense to keep options to a bare minimum for this mode: ignore color, and have a single formatting option (doesn't matter so long all information is there to get to the match).

One option maybe: whether or not to show the corresponding matching line of text from the file.

BurntSushi:
Then this should be closed as a duplicate.

There are *many* options, and their effects all need to be individually reasoned through.

> is there any doc on libripgrep, maybe a code snippet?

Please see the link in my previous comment. It's not done yet. Libripgrep refers to a collection of libraries. The endgame is to make ripgrep itself very small with most logic in reusable libraries. A lot of it is already done.

> can I link against it say from C-like languages?

Someone (probably not me) will need to put in the work to build a C interface, and then ffi bindings for their favorite language.

Issue #795, opened 2018-02-13
--line-number-width doesn't work well with --no-heading


timotheecour:
#### What version of ripgrep are you using?

ripgrep 0.8.0 (rev )
-SIMD -AVX

#### What operating system are you using ripgrep on?

osx

### what i'm observing:

rg --no-heading --line-number-width=10 packages

```
source/dubregistry/dbcontroller.d:       289:           m_packages.update(["_id": packId], ["$set": ["stats": stats]]);
source/dubregistry/dbcontroller.d:       345:                   auto res = m_packages.aggregate(["$group": group], ["$project": project]);
source/dubregistry/dbcontroller.d:       371:           foreach( bp; m_packages.find() ){
source/dubregistry/dbcontroller.d:       380:                           m_packages.update(["_id": p._id], ["$set": ["versions": newversions]]);
source/dubregistry/web.d:        46:            DbPackage[] m_packages;
source/dubregistry/web.d:        66:                    Package[] packages;
source/dubregistry/web.d:        77:            DbPackage[] packages;
```

### what i'd expect

that the line numbers would align WRT left margin (ie that they'd all lign up), not that they'd align WRT rightmost character of path (which has variable position due to different path lengths)

BurntSushi:
Sorry, but this will never work. The whole point of `--line-number-width` is to specify a fixed padding. There is no intelligent "alignment" going on. We can either let this bug stand or mark `--no-heading` as a conflicting flag with `--line-number-width`.

erikrw:
Perhaps I'm overlooking something, but could the padded line number be
printed before the file path on each line?

BurntSushi:
@erikrw It could, but we won't. That's not the standard grep output format.

BurntSushi:
I'm regretting ever adding this flag. I wish I considered this output format, because if I had, I would have never let this flag be a thing.

This is absolutely not something that ripgrep will ever fix because we are not going to get into the business of "aligning" output. I thought we could get away with something simple, but I didn't think carefully about the interaction with `--no-heading`. So this flag is either going to get removed in the next release or the flags get marked as conflicting (resulting in a ripgrep error).

timotheecour:
i can leave without this but what would be controversial about the following:

(requires monospace font, the | are aligned)

```
source/dubregistry/dbcontroller.d:371: |       foreach( bp; m_packages.find() ){
[...]

Issue #462, opened 2017-04-27
machine interface


albfan:
Hi, I'm the maintainer of https://github.com/albfan/ag.vim

Someone ask me to integrate with rg, and after trying it is sensibly faster than ag so want to ask if it's possible to add an option to expose results in a parseable way ready for machines.

By now I'm parsing filename, column, line and contents from `--group` option so this is not a stopper, but would be great to offer results that way to ease maintenance and complex features like hightlight code

Maybe `--porceain` could be a great name for option. Anything will work

kpp:
How about existing `--vimgrep`?

tiehuis:
#244 is related. May be worthwhile to see if what is outlined there differs from what your requirements.

albfan:
I use mostly search with context:

    $ rg foo -C4

Those contexts lines are really useful, so I really work on AgGroup (not the commands related with quicklist). Parsing results with context slows the interaction and is error prone

@tiehuis That would help to highlight without reprocess regex (with the escape characters hell)

Deterministic results like:

```
file: foo
   - matches: [
        {
            line: 1,
            col: 10,
            match: foo,
            before: 3,
            after: 4,
            result: "asdfasdfasdf
asdfsadfasdf
foo
asfasdfasdf
"         
 },
...
]
```

etc etc woud be great as there's lot of info rg knows but do not expose (reflected in highlights for example)

tiehuis:
Just to add, here (#359) was a similar issue regarding structured output (which looks very much the same as what you are suggesting) and a response from BurntSushi.

albfan:
@tiehuis Thanks that's exactly what I need.

I was afraid to ask for a library, but seems there's a work in progress. That's perfect for another idea I have for rg (use on gnome builder as a global search https://github.com/albfan/gnome-builder/issues/14)

Then we need to contribute on rg library. That makes sense, thanks!

Issue #360, opened 2017-02-12
multiline search for simple cases


timotheecour:
even though ripgrep is line oriented and full multiline search is sadly out of question (after reading https://github.com/BurntSushi/ripgrep/issues/176), supporting simple cases would be very useful, eg:

## full description
* `--multiline=[unordered|ordered] --window_size=N [--pattern $regex_i ...]`

will return a list of matches `file:line` such that:

with --unordered:
* file:line contains one of the K requested patterns
* the K-1 other patterns can be found in file at lines [line...line+N)

with --ordered:
* file:line contains the 1st of the K requested patterns
* the K-1 other patterns can be found in file at lines [line...line+N) in the order in which they are given on command line

`--window_size=N` represents the context to search for the K matches. When unspecified, N defaults to infinity (ie all patterns must occur in `file` from `line` to `end of file`

When only a single pattern is used, it has same behavior as standard `rg`

## Examples

`foo.d`:
```d
// first line
auto getFoo(string[] bar){
  // some comment
  return bar.sort.uniq;
}
```

```bash
$ rg --multiline=ordered --window_size=3 --pattern `string[]` --pattern `sort\b`
foo.d:2:auto getFoo(string[] bar){
foo.d:4%return bar.sort.uniq;

$ rg --multiline=ordered --window_size=2 --pattern `string[]` --pattern `sort\b`
# no match, `foo.d:4` is beyond `foo.d:[2...2+window_size)

$ rg --multiline=ordered --window_size=3 --pattern `sort\b` --pattern `string[]`
# no match, the first pattern matches at `foo.d:4` and we can't find the 2nd pattern in `foo.d:[4...4+3)`

$ rg --multiline=unordered --window_size=3 --pattern `sort\b` --pattern `string[]`
# now there's a match since we provided `unordered`
foo.d:2:auto getFoo(string[] bar){
foo.d:4%return bar.sort.uniq;
```

BurntSushi:
I don't understand the feature request. I don't know what `window_size` means and I don't know what `--multiline` does and I don't know the difference between `unordered` and `ordered`. It sounds like you have a lot of context in your head that you need to put into writing as a specification.

Tip: consider reusing the `-A`/`-B`/`-C` flags.

BurntSushi:
[...]

PR #802, opened 2018-02-14
fix issue #359 --machine-readable

timotheecour:
@BurntSushi 

addresses
* https://github.com/BurntSushi/ripgrep/issues/359 (json output option to allow scripting #359) 
* https://github.com/BurntSushi/ripgrep/issues/244 (Output format with parseable match indices? #244)
* https://github.com/BurntSushi/ripgrep/issues/462 (machine interface #462)

not using json but something simple:

outputs in format: `file:line:col:byte_offset:match_size:match:`

all the info is there for another tool to do efficient further processing as discussed in https://github.com/BurntSushi/ripgrep/issues/359 or https://github.com/BurntSushi/ripgrep/issues/795#issuecomment-365453598

## example
```
./target/debug/rg --color=never --only-matching --machine-readable 'machine(\-.*|\w)'
complete/_rg:47:12:2858:62:machine-readable)'{-v,--machine-readable}'[machine readable]':

src/printer.rs:68:5:1847:9:machine_:
src/printer.rs:114:13:3762:9:machine_:
src/printer.rs:175:12:5567:9:machine_:
src/printer.rs:176:14:5634:9:machine_:
src/printer.rs:319:17:1728:9:machine_:

src/args.rs:55:5:1225:9:machine_:
src/args.rs:189:14:6155:9:machine_:
src/args.rs:189:36:6177:9:machine_:
src/args.rs:383:13:4832:9:machine_:
src/args.rs:383:48:4867:20:machine-readable"),:

src/app.rs:535:10:4018:9:machine_:
src/app.rs:1031:9:5399:9:machine_:
src/app.rs:1036:30:5620:19:machine-readable"):

grep/src/data/sherlock.txt:9356:5:5141:9:machiner:
grep/src/data/sherlock.txt:12009:36:2316:9:machiner:
```

## design choices
* byte_offset is needed for O(1) access in a file (eg using seek or using a byte buffer)
* line+column is needed for line oriented tools and human convenience
* match_size is needed to know extent of match
* match is needed for tools to do further filtering on matched pattern to allow skipping reading the corresponding file when possible
* file:line:col (and this exact syntax) is a good order for the 1st 3 things to output because ripgrep already outputs in this format, and lots of tools recognize this to jump to a location (eg sublimetext)

## design questions
* not yet sure how to present context (both within line and across neighboring lines)
A tool can just use the match and compute the context by reading the matched file at corresponding location; but maybe we can also output context as follows:

for context within line: all info is present here since we have column + match size:
src/app.rs:1031:9:5399:9:123:here is the entire line:
in format:

`file:line:col:byte_offset:match_size:line_total_size:line:`
where `line_total_size` is useful in case we want to truncate at a max line length

* for context across lines we could do:

src/app.rs:1031:9:5399:9:123:here is previous line\n:here is the entire line\nhere is the next line:
in format:
`file:line_of_context_start:byte_offset_of_context_start:byte_offset_from_context_start:match_size:line_total_size:multi_lines:`

## use cases enabled by this (via external tooling)
* https://github.com/BurntSushi/ripgrep/issues/360 (multiline search for simple cases #360)
* https://github.com/BurntSushi/ripgrep/issues/346 (Feature idea: "context matches" #346)
* sorting without current performance limitation of --sort-files which disables parallelism
* output alignment (see https://github.com/BurntSushi/ripgrep/issues/795#issuecomment-365453598)
* https://github.com/BurntSushi/ripgrep/issues/357 (filter out duplicate results #357)

## TODO
- [ ] update man pages

### DIFF
--- a/complete/_rg
+++ b/complete/_rg
@@ -44,6 +44,7 @@ _rg() {
     '(-i -s -S --case-sensitive --ignore-case --smart-case)'{-i,--ignore-case}'[search case-insensitively]'
     '--ignore-file=[specify additional ignore file]:file:_files'
     '(-v --invert-match)'{-v,--invert-match}'[invert matching]'
+    '(-v --machine-readable)'{-v,--machine-readable}'[machine readable]'
     '(-n -N --line-number --no-line-number)'{-n,--line-number}'[show line numbers]'
     '(-N --no-line-number)--line-number-width=[specify width of displayed line number]:number of columns'
     '(-w -x --line-regexp --word-regexp)'{-x,--line-regexp}'[only show matches surrounded by line boundaries]'
--- a/src/app.rs
+++ b/src/app.rs
@@ -532,6 +532,7 @@ pub fn all_args_and_flags() -> Vec<RGArg> {
     flag_ignore_case(&mut args);
     flag_ignore_file(&mut args);
     flag_invert_match(&mut args);
+    flag_machine_readable(&mut args);
     flag_line_number(&mut args);
     flag_line_number_width(&mut args);
     flag_line_regexp(&mut args);
@@ -1023,6 +1024,16 @@ fn flag_invert_match(args: &mut Vec<RGArg>) {
 Invert matching. Show lines that do not match the given patterns.
 ");
     let arg = RGArg::switch("invert-match").short("v")
+        .help(SHORT).long_help(LONG);
+    args.push(arg);
+}
+
+fn flag_machine_readable(args: &mut Vec<RGArg>) {
+    const SHORT: &str = "Machine Readable output.";
+    const LONG: &str = long!("\
+Machine Readable output, for integration with external tools.
+");
+    let arg = RGArg::switch("machine-readable")
         .help(SHORT).long_help(LONG);
     args.push(arg);
 }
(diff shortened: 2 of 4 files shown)


okdana:
I didn't look into the context for this at all, nor the Rust changes, but a problem that immediately jumps out is this:

If you use colons for delimiters, and you have the file name and the match on the same line, it becomes impossible to reliably parse lines where the file name contains colons. (If you didn't include the match you could at least work backwards.) Admittedly this is not an *incredibly* common use case, but it's one that would affect a project i personally work on.

If the idea is sound perhaps the option could at least take an optional delimiter (which you'd probably want to set to `$'\0'` or at least `$'\037'`, tbh).

Also your change to the completion function doesn't follow the existing conventions and is actually broken. It should rather be something like:

```
'--machine-readable[output in machine-readable format]'
```

If you do add an optional delimiter, maybe something like:

```
'--machine-readable=-[output in machine-readable format]::field delimiter'
```

BurntSushi:
I'm going to close this. If someone wants to work on this, we need a thorough specification. Personally, I would rather we hold off on this until the printer gets refactored.

timotheecour:
@BurntSushi would you consider merging this (after I address comments) if I made the flag experimental, eg: `--experimental-machine-readable` ; (along with clear comments in help that this flag can be removed or have it's API changed with no deprecation cycle)

This issue has been there for a very long time and IMO a working solution (that is subject to change) is better than no solution.

PR #806, opened 2018-02-14
Add termcolor docs explaining what intense does (closes #797)

### LINKED ISSUE #797: termcolor docs should explain what "intense" means

Does not apply.

`termcolor` has an `intense` and `set_intense` methods, but from skimming the docs I have no idea what that does.

balajisivaraman:
I went through the code (and searched the WWW) to understand what the `intense` parameter does in `termcolor`. Hopefully I've captured the intent behind them to the best of my understanding.

### DIFF
--- a/termcolor/src/lib.rs
+++ b/termcolor/src/lib.rs
@@ -1252,9 +1252,23 @@ impl ColorSpec {
     }
 
     /// Get whether this is intense or not.
+    ///
+    /// On Unix-like systems, this will output the ANSI escape sequence
+    /// that will print a high-intensity version of the color
+    /// specified.
+    ///
+    /// On Windows systems, this will output the ANSI escape sequence
+    /// that will print a brighter version of the color specified.
     pub fn intense(&self) -> bool { self.intense }
 
     /// Set whether the text is intense or not.
+    ///
+    /// On Unix-like systems, this will output the ANSI escape sequence
+    /// that will print a high-intensity version of the color
+    /// specified.
+    ///
+    /// On Windows systems, this will output the ANSI escape sequence
+    /// that will print a brighter version of the color specified.
     pub fn set_intense(&mut self, yes: bool) -> &mut ColorSpec {
         self.intense = yes;
         self


termcolor/src/lib.rs
```diff
@@ -1252,9 +1252,21 @@ impl ColorSpec {
     }
 
     /// Get whether this is intense or not.
+    ///
+    /// On Unix-like systems, this will output the ANSI escape sequence
+    /// that will print the 256-color version of the color specified.
```
BurntSushi: I'm not an ANSI expert, but [Wikipedia (see 8-bit section)](https://en.wikipedia.org/wiki/ANSI_escape_code) seems to just classify these as "intense" colors. I don't know where the canonical documentation is for this though.

I think your Windows description looks good though!

balajisivaraman: Hmmm... I looked at the same page and got confused by the fact that the escape sequence was for 256-colors. Sorry about that. It does say that he sequence of numbers we're using (9-15) are for high intensity colors.

I also scoured the web and couldn't find the canonical documentation for it. 😞 

I'll just change it to say: "On Unix-like systems, this will output the ANSI escape sequence that will print a high-intensity version of the color specified." (We could use the word gaudy, which is what those colors look like to me but we probably don't want to.)

I understand that's a very vague description, but I cannot come up something better unfortunately.

BurntSushi: Haha! Yes, that wording sounds good. :) If another enterprising individual can make these descriptions more precise in the future, then that's great, but these will do for now!

okdana: The macro is misleadingly named. The 'intense' colours are the standard 8 with the bold/intense parameter set (e.g., `\e[1;35m` for intense magenta), or in the 8-bit set (indexed from `0`) they are `8` through `15`. `write_intense!()` simply uses the more comprehensive foreground/background colour sequence which supports the standard 8, the intense 8, and then 240 more pre-defined colours. If you did `write_intense!("5")` you'd get non-intense magenta, and if you did `write_intense!("251")` you'd get a pre-defined medium grey. I think you should even be able to do `write_intense!("255;0;255")` for 24-bit magenta, since it uses the same leading sequence.

BurntSushi: Yeah basically my understanding is that on Unix "intense" is just another variant of a predefined color, where as on Windows it's an actual setting that is toggled on the console.

PR #811, opened 2018-02-17
globset: support backslash escaping

### LINKED ISSUE #526: support escaping meta characters in glob patterns
I have the following in [my global `.gitignore`](https://github.com/gabebw/dotfiles/blob/master/.gitignore):

```gitignore
# ...other things...
[
]
# ...other things...
```

This is because my text editor sometimes creates files with these exact names, and I want to ignore them. However, ripgrep is (I think!) parsing these lines as part of a regex:

```
$ rg anything
/Users/gbwilliams/.gitignore: line 56: error parsing glob '[': unclosed character class; missing ']'
./.gitignore: line 2: error parsing glob '[': unclosed character class; missing ']'
No files were searched, which means ripgrep probably applied a filter you didn't expect. Try running again with --debug.
```

-----------------------

To reproduce:

1. `mkdir test_dir`
1. `cd test_dir`
1. `echo [ > .gitignore`
1. `echo ] >> .gitignore`
1. `rg anything`
1. See an error

Version:

```
$ rg -V
ripgrep 0.5.2
```

Debug output:

```
$ rg --debug anything
DEBUG:grep::search: regex ast:
Literal {
    chars: [
        'a',
        'n',
        'y',
        't',
        'h',
        'i',
        'n',
        'g'
    ],
    casei: false
}
DEBUG:grep::literals: literal prefixes detected: Literals { lits: [Complete(anything)], limit_size: 250, limit_class: 10 }
DEBUG:globset: glob converted to regex: Glob { glob: "**/*.sw[po]", re: "(?-u)^(?:/?|.*/).*\\.sw[po]$", opts: GlobOptions { case_insensitive: false, literal_separator: false }, tokens: Tokens([RecursivePrefix, ZeroOrMore, Literal('.'), Literal('s'), Literal('w'), Class { negated: false, ranges: [('p', 'p'), ('o', 'o')] }]) }
DEBUG:globset: glob converted to regex: Glob { glob: "**/*.py[co]", re: "(?-u)^(?:/?|.*/).*\\.py[co]$", opts: GlobOptions { case_insensitive: false, literal_separator: false }, tokens: Tokens([RecursivePrefix, ZeroOrMore, Literal('.'), Literal('p'), Literal('y'), Class { negated: false, ranges: [('c', 'c'), ('o', 'o')] }]) }
DEBUG:globset: built glob set; 1 literals, 25 basenames, 2 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 2 regexes
DEBUG:ignore::dir: /Users/gbwilliams/.gitignore: line 56: error parsing glob '[': unclosed character class; missing ']'
DEBUG:globset: glob converted to regex: Glob { glob: "**/*.sw[po]", re: "(?-u)^(?:/?|.*/).*\\.sw[po]$", opts: GlobOptions { case_insensitive: false, literal_separator: false }, tokens: Tokens([RecursivePrefix, ZeroOrMore, Literal('.'), Literal('s'), Literal('w'), Class { negated: false, ranges: [('p', 'p'), ('o', 'o')] }]) }
DEBUG:globset: glob converted to regex: Glob { glob: "**/*.py[co]", re: "(?-u)^(?:/?|.*/).*\\.py[co]$", opts: GlobOptions { case_insensitive: false, literal_separator: false }, tokens: Tokens([RecursivePrefix, ZeroOrMore, Literal('.'), Literal('p'), Literal('y'), Class { negated: false, ranges: [('c', 'c'), ('o', 'o')] }]) }
DEBUG:globset: built glob set; 1 literals, 25 basenames, 2 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 2 regexes
/Users/gbwilliams/.gitignore: line 56: error parsing glob '[': unclosed character class; missing ']'
DEBUG:globset: built glob set; 0 literals, 1 basenames, 0 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 0 regexes
./.gitignore: line 2: error parsing glob '[': unclosed character class; missing ']'
DEBUG:ignore::walk: ignoring ./.gitignore: Ignore(IgnoreMatch(Hidden))
No files were searched, which means ripgrep probably applied a filter you didn't expect. Try running again with --debug.
```

bmalehorn:
From `man 7 glob`:

    One can remove the special meaning of '?', '*' and '[' by preceding
    them by a backslash, or, in case this is part of a shell command
    line, enclosing them in quotes.

Conform to glob / fnmatch / git implementations by parsing `\?`, `\*`
and `\[` as literal `?`, `*` and `[`. If a backslash is followed by
anything else, parse it as before.

Closes #526

### DIFF
--- a/globset/src/glob.rs
+++ b/globset/src/glob.rs
@@ -969,8 +969,9 @@ mod tests {
 
     #[derive(Clone, Copy, Debug, Default)]
     struct Options {
-        casei: bool,
-        litsep: bool,
+        casei: Option<bool>,
+        litsep: Option<bool>,
+        bsesc: Option<bool>,
     }
 
     macro_rules! syntax {
@@ -1000,11 +1001,17 @@ mod tests {
         ($name:ident, $pat:expr, $re:expr, $options:expr) => {
             #[test]
             fn $name() {
-                let pat = GlobBuilder::new($pat)
-                    .case_insensitive($options.casei)
-                    .literal_separator($options.litsep)
-                    .build()
-                    .unwrap();
+                let mut builder = GlobBuilder::new($pat);
+                if let Some(casei) = $options.casei {
+                    builder.case_insensitive(casei);
+                }
+                if let Some(litsep) = $options.litsep {
+                    builder.literal_separator(litsep);
+                }
+                if let Some(bsesc) = $options.bsesc {
+                    builder.backslash_escape(bsesc);
+                }
+                let pat = builder.build().unwrap();
                 assert_eq!(format!("(?-u){}", $re), pat.regex());
             }
         };
@@ -1017,11 +1024,17 @@ mod tests {
         ($name:ident, $pat:expr, $path:expr, $options:expr) => {
             #[test]
             fn $name() {
-                let pat = GlobBuilder::new($pat)
-                    .case_insensitive($options.casei)
-                    .literal_separator($options.litsep)
-                    .build()
-                    .unwrap();
+                let mut builder = GlobBuilder::new($pat);
+                if let Some(casei) = $options.casei {
+                    builder.case_insensitive(casei);
+                }
+                if let Some(litsep) = $options.litsep {
+                    builder.literal_separator(litsep);
+                }
+                if let Some(bsesc) = $options.bsesc {
+                    builder.backslash_escape(bsesc);
+                }
+                let pat = builder.build().unwrap();
                 let matcher = pat.compile_matcher();
                 let strategic = pat.compile_strategic_matcher();
                 let set = GlobSetBuilder::new().add(pat).build().unwrap();
@@ -1039,11 +1052,17 @@ mod tests {
         ($name:ident, $pat:expr, $path:expr, $options:expr) => {
             #[test]
             fn $name() {
-                let pat = GlobBuilder::new($pat)
-                    .case_insensitive($options.casei)
-                    .literal_separator($options.litsep)
-                    .build()
-                    .unwrap();
+                let mut builder = GlobBuilder::new($pat);
+                if let Some(casei) = $options.casei {
+                    builder.case_insensitive(casei);
+                }
+                if let Some(litsep) = $options.litsep {
+                    builder.literal_separator(litsep);
+                }
+                if let Some(bsesc) = $options.bsesc {
+                    builder.backslash_escape(bsesc);
+                }
+                let pat = builder.build().unwrap();
                 let matcher = pat.compile_matcher();
                 let strategic = pat.compile_strategic_matcher();
                 let set = GlobSetBuilder::new().add(pat).build().unwrap();
@@ -1127,12 +1146,24 @@ mod tests {
     syntaxerr!(err_range2, "[z--]", ErrorKind::InvalidRange('z', '-'));
 
     const CASEI: Options = Options {
-        casei: true,
-        litsep: false,
+        casei: Some(true),
+        litsep: None,
+        bsesc: None,
     };
     const SLASHLIT: Options = Options {
-        casei: false,
-        litsep: true,
+        casei: None,
+        litsep: Some(true),
+        bsesc: None,
+    };
+    const NOBSESC: Options = Options {
+        casei: None,
+        litsep: None,
+        bsesc: Some(false),
+    };
+    const BSESC: Options = Options {
+        casei: None,
+        litsep: None,
+        bsesc: Some(true),
     };
 
     toregex!(re_casei, "a", "(?i)^a$", &CASEI);
@@ -1245,6 +1276,17 @@ mod tests {
     #[cfg(not(unix))]
     matches!(matchslash5, "abc\\def", "abc/def", SLASHLIT);
 
+    matches!(matchbackslash1, "\\[", "[", BSESC);
+    matches!(matchbackslash2, "\\?", "?", BSESC);
+    matches!(matchbackslash3, "\\*", "*", BSESC);
+    matches!(matchbackslash4, "\\[a-z]", "\\a", NOBSESC);
+    matches!(matchbackslash5, "\\?", "\\a", NOBSESC);
+    matches!(matchbackslash6, "\\*", "\\\\", NOBSESC);
+    #[cfg(unix)]
+    matches!(matchbackslash7, "\\a", "a");
+    #[cfg(not(unix))]
+    matches!(matchbackslash8, "\\a", "/a");
+
     nmatches!(matchnot1, "a*b*c", "abcd");
     nmatches!(matchnot2, "abc*abc*abc", "abcabcabcabcabcabcabca");
     nmatches!(matchnot3, "some/**/needle.txt", "some/other/notthis.txt");
@@ -1289,13 +1331,20 @@ mod tests {
         ($which:ident, $name:ident, $pat:expr, $expect:expr) => {
             extract!($which, $name, $pat, $expect, Options::default());
         };
-        ($which:ident, $name:ident, $pat:expr, $expect:expr, $opts:expr) => {
+        ($which:ident, $name:ident, $pat:expr, $expect:expr, $options:expr) => {
             #[test]
             fn $name() {
-                let pat = GlobBuilder::new($pat)
-                    .case_insensitive($opts.casei)
-                    .literal_separator($opts.litsep)
-                    .build().unwrap();
+                let mut builder = GlobBuilder::new($pat);
+                if let Some(casei) = $options.casei {
+                    builder.case_insensitive(casei);
+                }
+                if let Some(litsep) = $options.litsep {
+                    builder.literal_separator(litsep);
+                }
+                if let Some(bsesc) = $options.bsesc {
+                    builder.backslash_escape(bsesc);
+                }
+                let pat = builder.build().unwrap();
                 assert_eq!($expect, pat.$which());
             }
         };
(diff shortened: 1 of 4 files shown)


BurntSushi:
@bmalehorn Thanks for putting the work into this! #526 raised a number of issues with this, particularly with respect to Windows. Could you explain how those issues are addressed in this PR? I also see a test that is only run on Unix, but no comment explaining why. Could you elaborate on that as well?

okdana:
Aside from the Windows path issue, if i'm reading it right, this doesn't actually conform to the real-world behaviour of either `fnmatch(3)` or Git: A `\` should escape *any* character that follows it, even if it's not a meta-character. In other words, not only does the pattern `foo\*bar` match the string `foo*bar`, `foo\bar` also matches `foobar`.

(For comparison, [here](https://github.com/BurntSushi/ripgrep/issues/526#issuecomment-310754910) is where i provided my own attempt at fixing it, and as mentioned i think it behaves pretty well on UNIX, but it leaves the Windows issues unaddressed just like this one does.)

bmalehorn:
@okdana That's not quite correct, `foo\*bar` matches `foo*bar` but `foo\bar` will still only match `foo\bar`. If the character after `\` isn't one of the special characters, we emit both characters.

However I'm realizing now what you two probably realized a while ago, that this is going to break `-g` on Windows. Even though `src\foo.rs` will still work on Windows with this commit, globs like `src\*.rs` won't. Instead, you'd have to type `src/*.rs` or `src[\]*.rs`.

I suppose the real problem stems from using the same glob code for both `-g` and gitignores. `rg -g` is considerate to Windows and allows backslashes. Meanwhile Git demands .gitignores be written only with forward slashes, freeing up backslashes for escaping. Ditto for Mercurial and .hgignores.

So I suppose the only ways to parse .gitignores 100% accurately are:

1. (this PR) make `\` unusable in some situations; annoy Windows users, OR
2. only apply this glob change for .gitignores; parse `-g` and .gitignores differently; be inconsistent

Both of these kind of suck so I can see why this issue hasn't been closed yet.

Does that explanation make sense to everyone? Perhaps it's best to leave everything as it is: don't annoy Windows users, don't be inconsistent, but do parse a rare .gitignore pattern incorrectly.

okdana:
>@okdana That's not quite correct, `foo\*bar` matches `foo*bar` but `foo\bar` will still only match `foo\bar`

How did you come to that conclusion?

```
% cat fnmatch.c
#include <fnmatch.h>
#include <stdio.h>
int main(int argc, char *argv[]) {
  printf("fnmatch() returned %d\n", fnmatch(argv[1], argv[2], 0));
}

% gcc -o fnmatch fnmatch.c

% ./fnmatch 'foo\bar' 'foo\bar'
fnmatch() returned 1
% ./fnmatch 'foo\bar' 'foobar'
fnmatch() returned 0
```

Or with Git:

```
% mkdir rg811 && cd rg811 && git init
Initialized empty Git repository in /private/tmp/rg811/.git/
% touch .gitignore foobar 'foo\bar'

% git status --porcelain
?? .gitignore
?? "foo\\bar"
?? foobar

% echo 'foo\bar' > .gitignore; git status --porcelain
?? .gitignore
?? "foo\\bar"

% echo 'foo\\bar' > .gitignore; git status --porcelain
?? .gitignore
?? foobar
```

I personally would like `\` to work as expected on UNIX, and since the `gitignore` spec doesn't support `\` as a path separator on Windows i feel like it's kind of problematic that `globset` does. If i were a Windows user i think i would just expect `rg` to behave like Git. But i can imagine that making a unilateral change like that could break stuff, so idk

bmalehorn:
@okdana Sorry if I wasn't clear, but I was describing the behavior of my patch, not `fnmatch` or `git`. But I see now how my patch differs from `fnmatch`.

> I personally would like \ to work as expected on UNIX, and since the gitignore spec doesn't support \ as a path separator on Windows i feel like it's kind of problematic that globset does. If i were a Windows user i think i would just expect rg to behave like Git. But i can imagine that making a unilateral change like that could break stuff, so idk

I propose a compromise: on Windows, globset would behave as-is, but on Unix, we would correct it to behave like fnmatch (like in your patch). Windows users could still use `\` in pathnames, and Unix users would get exactly fnmatch behavior. Sound reasonable?

okdana:
>Sorry if I wasn't clear, but I was describing the behavior of my patch

Oh, sorry.

>Does that sound like a good idea to anyone?

That's how PHP does it. Not sure if it's the best option or not, but it's an option...

BurntSushi:
> I propose a compromise: on Windows, globset would behave as-is, but on Unix, we would correct it to behave like fnmatch (like in your patch). Windows users could still use `\` in pathnames, and Unix users would get exactly fnmatch behavior. Does that sound like a good idea to anyone?

This sounds plausible, but we need to be fastidious in how we go about this. That, I suspect this should be a [configuration knob on the `Glob` type](https://docs.rs/globset/0.3.0/globset/struct.GlobBuilder.html) so that users of the library can choose what behavior they want explicitly, if necessary. I'd probably call it:

```
fn forward_slash_separator(&mut sefl, yes: bool) -> &mut GlobBuilder<'a> {
```

We should document this to say that when a forward slash is a separator, then it will never be interpreted as the start of an escape sequence. But when disabled, `\` may be used to escape meta characters. We should document this in the syntax description in the main docs of the `globset` crate.

I'm tempted to say that this should be disabled by default on all non-Windows platforms and enabled by default on Windows platforms, but that kind of conditional logic in a somewhat fundamental crate like this seems... unwise. Alas, we already have some conditional logic in that we always treat `\` as a separator on Windows: https://github.com/BurntSushi/ripgrep/blob/597bf04a56d43aa9c0eb0f8fbb90c9d51c53656c/globset/src/glob.rs#L730 --- And looking at that more, explicitly calling out "forward slash" in the API seems wrong too, since we've broken the "path separator" abstraction by doing so.

What about this instead?

```
/// When enabled, a forward slash (`\`) may be used to escape
/// special characters in a glob pattern. Additionally, this will
/// prevent `\` from being interpreted as a path separator on all
/// platforms.
///
/// This is enabled by default on platforms where `\` is not a
/// path separator and disabled by default on platforms where `\`
/// is a path separator.
fn forward_slash_escape(&mut sefl, yes: bool) -> &mut GlobBuilder<'a> {
```

Regrettably, there is no way around a breaking change here, which means this will require another semver bump. /sigh

BurntSushi:
@retep998 Do you have any opinions on how `\` is handled in glob patterns on Windows?

okdana:
Nit-pick: `\` is the back-slash, not forward-slash

An option like that sounds reasonable to me for whatever it's worth. The only concerns i can think of with the platform-specific method are (1) it might trick Windows users into thinking that Git supports `\` as a separator (which it apparently doesn't), and (2) it still leaves them out in the cold when it comes to being able to escape things.

I never use Windows, so i have zero personal stake in how well `rg` works there, but if i *were* a Windows user i think i'd find that limitation somewhat bothersome. On the other hand, like i mentioned in the other thread, most of the glob meta-characters are illegal in Windows file names anyway (or they were last time i used it), so maybe it doesn't actually matter that much in practice...

BurntSushi:
> Nit-pick: \ is the back-slash, not forward-slash

:rofl: Thanks for that. Derp.

> I never use Windows, so i have zero personal stake in how well rg works there, but if i were a Windows user i think i'd find that limitation somewhat bothersome. On the other hand, like i mentioned in the other thread, most of the glob meta-characters are illegal in Windows file names anyway (or they were last time i used it), so maybe it doesn't actually matter that much in practice...

Yeah, I share your trepidation, definitely. I don't use Windows either, so I basically have no idea what the common or expected usage patterns are. Bascially, I hear your argument, but I also think, "Shouldn't Windows users be able to use their native path separator? Maybe they are just annoyed that git doesn't allow it?" But yeah, hopefully @retep998 can shed some light. @roblourens might also have opinions!

retep998:
`.gitignore` is a cross platform file where the paths are expected to work everywhere, so requiring `/` in that is reasonable. However on the command line `git` is perfectly happy to work with my normal Windows paths that have `\` in them, and if you give `git` a glob path like `src\*.rs` on the command line, it will interpret `\` as a path separator and `*` is not escaped.

Any changes to make `\` an escape character on Windows would be exceptionally annoying for Windows users.

BurntSushi:
> Any changes to make `\` an escape character on Windows would be exceptionally annoying for Windows users.

I understand this for the CLI, definitely. But is it also true for `.gitignore` files as well? That is, the whole point of this PR is that `\` is used as an escape in globs in `.gitignore` files. Should we allow that on Windows too for `.gitignore`?

But that is a good point about the CLI. We may want to toggle this behavior differently depending on whether the globs are in .gitignore or the CLI.

retep998:
On Windows, in `.gitignore`, `\` is indeed used as an escape character and not a path separator.

BurntSushi:
All right, I guess that's good enough for me.

@bmalehorn I think my comment here still seems mostly relevant: https://github.com/BurntSushi/ripgrep/pull/811#issuecomment-366962156 That is, we still need to make `\` as an escape character be a toggleable feature. I think enabling it by default on Unix and disabling it by default on Windows makes the most sense for `globset`. Then in the `ignore` crate, we can enable it unconditionally regardless of platform.

bmalehorn:
This all makes sense and sounds like the most reasonable solution. I'll update this PR tomorrow.

bmalehorn:
I probably should have commented when I updated the PR. @BurntSushi thoughts on the updated PR?

PR #814, opened 2018-02-17
search: add support for counting individual matches (Fixes #566)

### LINKED ISSUE #566: Have a --count option that counts occurrences instead of matched lines:
Right now -c/--count counts line matches, it would be great to have an option to count matches and not lines matched. Looking at the example:

```bash
$ echo "foo foo" > example.txt
$ rg -c foo example.txt
1
$ rg -o foo example.txt
1:foo
1:foo
```

would be interesting to have a way to get the total matches (2) in this case.

balajisivaraman:
As discussed in #566, this PR adds a `--count-matches` flag that will count the individual matches instead of the matching lines.

- This does essentially the same thing `printer.matched` does for printing only matches, which is to use the Regex's `find_iter` method for getting the individual matches. We then increment a variable to keep track of the count, ignoring the actual match itself. Since this will always happen behind a flag, I hope this won't have any performance implications.
- I initially considered using the existing `match_count` variable in `search_stream` and `search_buffer` to keep track of individual matches if `--count-matches` were passed in. This would've been handy if we later wanted to use the individual match count for something like the `--stats` flag, as it would get returned all the way up to `main.rs`; we can then use this for our purposes.

  However, I decided against this because it conflicts with the existing `--max-count` argument, which terminates the search early if we've hit the max count of matching lines. As a result, `rg` still terminates for matching line count and not individual match count even after this change. I'm just throwing this out there for your consideration.

  Currently, I've renamed the existing `match_count` to `matching_line_count` to better reflect what it keeps track of. `match_count` is now an `Option<usize>` and used to keep track of individual match count.
- `-c/--count` and `--count-matches` will override each other.


src/search_buffer.rs
```diff
@@ -21,7 +21,8 @@ pub struct BufferSearcher<'a, W: 'a> {
     grep: &'a Grep,
     path: &'a Path,
     buf: &'a [u8],
-    match_count: u64,
+    matching_line_count: u64,
```
BurntSushi: Could you name this `match_line_count` so that it's more consistent with `match_count`?

src/search_buffer.rs
```diff
@@ -118,39 +129,53 @@ impl<'a, W: WriteColor> BufferSearcher<'a, W> {
...
+        if self.opts.invert_match && !self.opts.terminate(self.matching_line_count) {
             let upto = self.buf.len();
             self.print_inverted_matches(last_end, upto);
         }
-        if self.opts.count && self.match_count > 0 {
-            self.printer.path_count(self.path, self.match_count);
+        if self.opts.count && self.matching_line_count > 0 {
+            self.printer.path_count(self.path, self.matching_line_count);
+        } else if self.opts.count_matches
+            && self.match_count.map_or(false, |c| c > 0) {
```
BurntSushi: When wrapping a conditional to multiple lines, could you move the brace down to the next line so that the division between the conditional and the conditional body is more clear?

src/search_buffer.rs
BurntSushi: Also, in the future (no need to do this for this PR), it would be super helpful to put renames (and similarly uninteresting changes) in a separate commit. In this case, a commit that came before the principle change. The reason is that I don't really care about the mechanics of a name change so long as the compiler is happy, but it increases the noise in the diff among other changes that I do carefully want to look at.

It can be hard to do this though because it does require some forethought, so don't feel like I'm imposing a strong requirement! But if you think of it and it's not too much work, it definitely helps the review process.

src/search_buffer.rs
```diff
@@ -118,39 +129,53 @@ impl<'a, W: WriteColor> BufferSearcher<'a, W> {
...
+        self.matching_line_count = 0;
         self.line_count = if self.opts.line_number { Some(0) } else { None };
+        self.match_count = if self.opts.count_matches { Some(0) } else { None };
         let mut last_end = 0;
         for m in self.grep.iter(self.buf) {
             if self.opts.invert_match {
                 self.print_inverted_matches(last_end, m.start());
             } else {
                 self.print_match(m.start(), m.end());
+                self.count_individual_matches(m.start(), m.end());
```
BurntSushi: Would it be possible to move this call into `print_match` itself?

src/search_stream.rs
```diff
@@ -291,6 +305,7 @@ impl<'a, R: io::Read, W: WriteColor> Searcher<'a, R, W> {
                     self.print_after_context(start);
                     self.print_before_context(start);
                     self.print_match(start, end);
+                    self.count_individual_matches(start, end);
```
BurntSushi: Same as for the mmap searcher. Can you move this into the `print_match` method?

src/search_buffer.rs
balajisivaraman: Ah this does indeed make a lot of sense to me. I understand completely and will make the change in a separate commit for this PR also. No issues. :-) Good housekeeping is always welcome.

BurntSushi:
@balajisivaraman OK, I finally got a chance to merge this today in https://github.com/BurntSushi/ripgrep/commit/27fc9f2fd341bd3ed672f79d43bf983514188b96. All I did was fix up the conflicts with latest master and reformatted the commit message. (Commit messages should note the issues they close.) I also added an additional commit that causes `--count --only-matching` to behave the same as `--count-matches`. (The next release of ripgrep will be 0.9.0, so this is OK.)

PR #818, opened 2018-02-19
printer: add support for printing 0-based byte offset before matches

### LINKED ISSUE #812: Add option to report file position instead of line number

ripgrep 0.8.0 (rev 23d1b91ead)
+SIMD -AVX

Ubuntu 16.04

Please add a feature to report the file offset of the lines containing matches. GNU grep has this

```
       -b, --byte-offset
              Print the 0-based byte offset within the  input  file  before  each  line  of  output.   If  -o
              (--only-matching) is specified, print the offset of the matching part itself.

```

I have a process that generates terabytes of diagnostic output daily. I desire to scan all that
for certain patterns and then to have a script read forward from the lines the patterns are on.
It would be much faster to fseek() to the offset of each matching line than to read all the data
counting lines. rg would be the perfect tool for generating the list of interesting offsets if it offered
that functionality. Thanks.

balajisivaraman:
Closes #812 

As discussed in the issue thread, the only question is the order of line number, column and offset when all three are displayed. I've gone with `line number:column:offset` as the default for now.

### DIFF
--- a/tests/tests.rs
+++ b/tests/tests.rs
@@ -395,6 +395,16 @@ sherlock!(csglob, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
     assert_eq!(lines, "file2.html:Sherlock\n");
 });
 
+sherlock!(byte_offset_only_matching, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
+    cmd.arg("-b").arg("-o");
+    let lines: String = wd.stdout(&mut cmd);
+    let expected = "\
+sherlock:56:Sherlock
+sherlock:177:Sherlock
+";
+    assert_eq!(lines, expected);
+});
+
 sherlock!(count, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
     cmd.arg("--count");
     let lines: String = wd.stdout(&mut cmd);
--- a/complete/_rg
+++ b/complete/_rg
@@ -25,6 +25,7 @@ _rg() {
     '*--colors=[specify color settings and styles]: :->colorspec'
     '--column[show column numbers]'
     '(-A -B -C --after-context --before-context --context)'{-C+,--context=}'[specify lines to show before and after each match]:number of lines'
+    '(-b --byte-offset)'{-b,--byte-offset}'[print the 0-based byte offset for each matching line]'
     '--context-separator=[specify string used to separate non-continuous context lines in output]:separator'
     '(-c --count --passthrough --passthru)'{-c,--count}'[only show count of matches for each file]'
     '--debug[show debug messages]'
--- a/src/app.rs
+++ b/src/app.rs
@@ -509,6 +509,7 @@ pub fn all_args_and_flags() -> Vec<RGArg> {
     // Flags can be defined in any order, but we do it alphabetically.
     flag_after_context(&mut args);
     flag_before_context(&mut args);
+    flag_byte_offset(&mut args);
     flag_case_sensitive(&mut args);
     flag_color(&mut args);
     flag_colors(&mut args);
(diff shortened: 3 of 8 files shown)


tests/tests.rs
```diff
@@ -395,6 +395,26 @@ sherlock!(csglob, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
...
+sherlock!(byte_offset_only_matching, "Sherlock", ".", |wd: WorkDir, mut cmd: Command| {
+    cmd.arg("-b").arg("-o");
+    let lines: String = wd.stdout(&mut cmd);
+    let expected = "\
+sherlock:56:Sherlock
+sherlock:177:Sherlock
+";
+    assert_eq!(lines, expected);
+});
+
```
BurntSushi: For completeness, could you pass the `--no-mmap` flags to each of the above tests, and then copy the tests and change `--no-mmap` to `--mmap`? I am thinking that your implementation might actually be correct for `--mmap` but not for `--no-mmap`. I'm not sure a global byte offset is even tracked in the `--no-mmap` searcher, so this task might require a bit more work. :)

balajisivaraman: I'm really sorry about overlooking this. When I went into `search_stream`, I noticed that the change was pretty trivial and it made so much sense to me that I missed `search_buffer`. It's on me. I'll have a look at the `mmap` searcher to see what can be done. :+1:

balajisivaraman: Ah, I think I lucked out here. The iterator returned by `self.grep.iter(self.buf)` in `search_buffer.rs` provides us with the starting position of the matching line itself. From then on, we call `printer.matched` with the `start` and `end`, which thankfully works the same way it does for `search_stream`. Both the `--no-mmap` and `--mmap` tests that you asked to write also pass. :+1:

BurntSushi: Hmm, awesome! You might actually want to test this also in non-mmap searcher using a smaller capacity, such that the matches reported span multiple buffers. If that test passes, then I'll believe the implementation is correct. :)

BurntSushi: I believe there's a `smallcap` helper function in the tests somewhere, since there are other tests that proceed in a similar vein. But I can't remember. It's been a while since I looked at that code.

balajisivaraman: Hmm... No, you're right. I tried it with the `smallcaps` test available in `search_stream.rs` and it completely blew apart. (I got 0 as the offset for the next match instead of the offset from the beginning of the file.) The good news is that I can see why this is happening very clearly.

I now understand why you originally thought it will require additional work for the `--no-mmap` case. I'll have a look at it. Sorry. :+1:

balajisivaraman:
@BurntSushi, Thanks for the patient explanation and walkthrough yesterday. I've identified the bug and, hopefully, fixed it as well. :-)

One point to note is that within the searchers, I've named the variable that keeps track of the offset as `buffer_offset`. This made sense to me as it tracks the offset of the current buffer from the beginning of the file; and I can't come up with a better name for it.

BurntSushi:
@balajisivaraman OK, I finally had a chance to look into merging this. Overall, this looks really great. I merged this PR in b006943c01561a4ae5b928081d1bb4087b599e19 with a few minor tweaks and a bug fix:

1. I renamed `buffer_offset` to `byte_offset`, since the former isn't actually true for the stream searcher. (It is a byte offset for the entire file, not to a buffer.)
2. I changed the type of `byte_offset` to `u64` from `usize`. This is because `usize` is a platform specific type whose size is tied to the size of addressable memory. So on a 32 bit system, `usize`'s max value is `2^32-1`, which means the byte offset would overflow when searching files bigger than 4GB. Note that this is tricky to get right, since there are places in the searcher where `usize` is correct since they correspond to offsets into memory.
3. GNU grep's behavior when one of the `--context` flags is given is to emit a byte offset before each context line. I updated your patch to do that and fixed the test, which was easy to do thanks to your prior work. :-)

Thanks so much for working on this!

PR #833, opened 2018-02-23
Update snapcraft.yaml

anthonywong:
Update version number and don't declare alias as it has been deprecated.


BurntSushi:
Thanks!

Issue #835, opened 2018-02-25
Small performance regression in ignore-0.4.x


sharkdp:
#### What operating system are you using ripgrep on?

Arch Linux, x86_64

#### Describe your question, feature request, or bug.

I was running some benchmarks for `fd` and noticed that there was a small (~5%) performance regression when I updated from `ignore-0.2.2` to `ignore-0.4.x`. I investigated some more and ran the same benchmark (`fd -uu '[0-9]\.jpg$'` in my home folder) multiple times while changing the version of the ignore crate:

![image](https://user-images.githubusercontent.com/4209276/36641938-2a14fe2c-1a38-11e8-9870-f11e8f999707.png)

There are two things to notice:
1. The timing for `ignore-0.3.0` is not visisble because it's so fast (137 ms). This is due to a bug where not all results are shown (I didn't investigate further, but the git log seems to indicate that something was wrong with `ignore-0.3.0`).
2. There is a small (but statistically significant) performance regression that seems to have been introduced with commit 3cb4d13 (Custom ignore files). Interestingly, I'm running a search that doesn't even use ignore files (`-uu` is the same as in ripgrep).

#### What are the steps to reproduce the behavior?

I can reproduce this behavior with ripgrep itself:

```
▶ git checkout 51864c1; cargo install -f

...

▶ hyperfine --warmup 10 'rg -uu --files --iglob "*[0-9].jpg" ~'
Benchmark #1: rg -uu --files --iglob "*[0-9].jpg" ~

  Time (mean ± σ):     700.4 ms ±   5.3 ms    [User: 3.157 s, System: 2.310 s]
 
  Range (min … max):   691.4 ms … 709.0 ms

▶ git checkout 3cb4d13; cargo install -f

...

▶ hyperfine --warmup 10 'rg -uu --files --iglob "*[0-9].jpg" ~'
Benchmark #1: rg -uu --files --iglob "*[0-9].jpg" ~

  Time (mean ± σ):     725.4 ms ±   7.9 ms    [User: 3.318 s, System: 2.338 s]
 
  Range (min … max):   715.5 ms … 743.9 ms
```
(benchmarks performed with [hyperfine](https://github.com/sharkdp/hyperfine) on a warm disk cache)

Again, the difference is small (~4%) but significant.

sharkdp:
This seems to be caused by the additional `Gitignore` instance `custom_ignore_matcher` that has to be constructed in `Ignore::add_child_path`.

[...]

PR #836, opened 2018-02-25
Proof of concept: skip unused Gitignore matchers

sharkdp:
This is just a proof of concept. I'm not sure if this is the best approach. Even if it is, the code should probably be clean up.

This PR addresses the findings in #835 and disables unneeded `Gitignore` matchers completely. This leads to a performance improvement for use cases where one or more of the matchers (custom ignore files, `.ignore` files, `.gitignore` files, `.git/info/exclude`) are not needed.

### Benchmark 1: `rg -uu --files`

master
```
▶ hyperfine --warmup 10 'rg -uu --files'
Benchmark #1: rg -uu --files

  Time (mean ± σ):      1.298 s ±  0.067 s    [User: 4.348 s, System: 2.667 s]
 
  Range (min … max):    1.203 s …  1.398 s
```

this branch (~ 8% faster):
```
▶ hyperfine --warmup 10 'rg -uu --files'
Benchmark #1: rg -uu --files

  Time (mean ± σ):      1.201 s ±  0.032 s    [User: 3.732 s, System: 2.532 s]
 
  Range (min … max):    1.165 s …  1.264 s
```

### Benchmark 2: `rg -uu --files --glob "Cargo.toml"`

master:
```
▶ hyperfine --warmup 10 'rg -uu --files --glob "Cargo.toml"'
Benchmark #1: rg -uu --files --glob "Cargo.toml"

  Time (mean ± σ):     662.5 ms ±   2.2 ms    [User: 2.991 s, System: 2.198 s]
 
  Range (min … max):   659.3 ms … 665.7 ms
```

this branch (~ 10% faster):
```
▶ hyperfine --warmup 10 'rg -uu --files --glob "Cargo.toml"' 
Benchmark #1: rg -uu --files --glob "Cargo.toml"

  Time (mean ± σ):     602.6 ms ±   1.7 ms    [User: 2.527 s, System: 2.183 s]
 
  Range (min … max):   599.1 ms … 604.8 ms
```

### Benchmark 3: `rg --files --glob "Cargo.toml"`

As expected, there is no significant change if `-uu` is not used:

master:
```
▶ hyperfine --warmup 10 'rg --files --glob "Cargo.toml"'    
Benchmark #1: rg --files --glob "Cargo.toml"

  Time (mean ± σ):     101.1 ms ±   1.6 ms    [User: 420.4 ms, System: 327.0 ms]
 
  Range (min … max):    99.5 ms … 107.1 ms
```

this branch:
```
▶ hyperfine --warmup 10 'rg --files --glob "Cargo.toml"'
Benchmark #1: rg --files --glob "Cargo.toml"

  Time (mean ± σ):     101.6 ms ±   0.6 ms    [User: 419.2 ms, System: 331.5 ms]
 
  Range (min … max):   100.3 ms … 103.0 ms
```

### DIFF
--- a/ignore/src/dir.rs
+++ b/ignore/src/dir.rs
@@ -107,15 +107,15 @@ struct IgnoreInner {
     /// Ignore files used in addition to `.ignore`
     custom_ignore_filenames: Arc<Vec<OsString>>,
     /// The matcher for custom ignore files
-    custom_ignore_matcher: Gitignore,
+    custom_ignore_matcher: Option<Gitignore>,
     /// The matcher for .ignore files.
-    ignore_matcher: Gitignore,
+    ignore_matcher: Option<Gitignore>,
     /// A global gitignore matcher, usually from $XDG_CONFIG_HOME/git/ignore.
     git_global_matcher: Arc<Gitignore>,
     /// The matcher for .gitignore files.
-    git_ignore_matcher: Gitignore,
+    git_ignore_matcher: Option<Gitignore>,
     /// Special matcher for `.git/info/exclude` files.
-    git_exclude_matcher: Gitignore,
+    git_exclude_matcher: Option<Gitignore>,
     /// Whether this directory contains a .git sub-directory.
     has_git: bool,
     /// Ignore config.
@@ -210,34 +210,38 @@ impl Ignore {
         let mut errs = PartialErrorBuilder::default();
         let custom_ig_matcher =
             {
-                let (m, err) =
-                    create_gitignore(&dir, &self.0.custom_ignore_filenames);
-                errs.maybe_push(err);
-                m
+                if self.0.custom_ignore_filenames.len() > 0 {
+                    let (m, err) =
+                        create_gitignore(&dir, &self.0.custom_ignore_filenames);
+                    errs.maybe_push(err);
+                    Some(m)
+                } else {
+                    None
+                }
             };
         let ig_matcher =
             if !self.0.opts.ignore {
-                Gitignore::empty()
+                None
             } else {
                 let (m, err) = create_gitignore(&dir, &[".ignore"]);
                 errs.maybe_push(err);
-                m
+                Some(m)
             };
         let gi_matcher =
             if !self.0.opts.git_ignore {
-                Gitignore::empty()
+                None
             } else {
                 let (m, err) = create_gitignore(&dir, &[".gitignore"]);
                 errs.maybe_push(err);
-                m
+                Some(m)
             };
         let gi_exclude_matcher =
             if !self.0.opts.git_exclude {
-                Gitignore::empty()
+                None
             } else {
                 let (m, err) = create_gitignore(&dir, &[".git/info/exclude"]);
                 errs.maybe_push(err);
-                m
+                Some(m)
             };
         let ig = IgnoreInner {
             compiled: self.0.compiled.clone(),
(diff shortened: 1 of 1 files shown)


BurntSushi:
@sharkdp Thanks for this! I think I'd be OK with this PR, but just to cover our bases, did you explore whether this improvement was possible by just using `Gitignore::empty()`? That was basically my intent, where if a matcher wasn't needed, then we could create an empty one and the empty cases would just get folded into the matcher instead of needing to add a bunch of explicit case analysis everywhere.

Let's work through this. Here's the definition of `Gitignore::empty()`:

```rust
    pub fn empty() -> Gitignore {
        GitignoreBuilder::new("").build().unwrap()
    }
```

OK, so this is creating a new builder with an empty root, not adding any globs and then building it. That means running this code:

```rust
    pub fn new<P: AsRef<Path>>(root: P) -> GitignoreBuilder {
        let root = root.as_ref();
        GitignoreBuilder {
            builder: GlobSetBuilder::new(),
            root: strip_prefix("./", root).unwrap_or(root).to_path_buf(),
            globs: vec![],
            case_insensitive: false,
        }
    }
```

and this code

```rust
    pub fn build(&self) -> Result<Gitignore, Error> {
        let nignore = self.globs.iter().filter(|g| !g.is_whitelist()).count();
        let nwhite = self.globs.iter().filter(|g| g.is_whitelist()).count();
        let set =
            self.builder.build().map_err(|err| {
                Error::Glob {
                    glob: None,
                    err: err.to_string(),
                }
            })?;
        Ok(Gitignore {
            set: set,
            root: self.root.clone(),
            globs: self.globs.clone(),
            num_ignores: nignore as u64,
            num_whitelists: nwhite as u64,
            matches: Arc::new(ThreadLocal::default()),
        })
    }
```

The constructor seems pretty cheap to me. There's a call to `to_path_buf` which might invoke an allocation, although given it's an empty string, that probably gets  optimized out. (IIRC, `vec![]` doesn't actually allocate, for example, and `PathBuf` is certainly built on `Vec`.)

The `build` method is a little more interesting. The first couple lines filter over the globs, but we have none, so that should be negligible. Then we build the `GlobSet` itself. This is actually cheap since `GlobSet::new(&[])` is special cases to do no work and just return a set that never matches.

The builder otherwise does some cloning (but again, those should all be empty), and sets up a scratch buffer (which also defaults to empty).

So my question is: if we changed this

```rust
        let custom_ig_matcher =
            {
                let (m, err) =
                    create_gitignore(&dir, &self.0.custom_ignore_filenames);
                errs.maybe_push(err);
                m
            };
```

to be more in line with how the other matchers handle the empty case on master

```rust
        let custom_ig_matcher =
            if self.custom_ignore_filenames.is_empty() {
                Gitignore::empty()
            } else {
                let (m, err) =
                    create_gitignore(&dir, &self.0.custom_ignore_filenames);
                errs.maybe_push(err);
                m
            };
```

then does that solve the performance problem? Or are there costs in `Gitignore::empty()` that I am not accounting for?

(The reason why I think the above might work is because the `create_gitignore` function definitely does have overhead associated with it, and always does at least a syscall to open a file.

sharkdp:
@BurntSushi Thank you for the detailed answer.

I should have explained this better:

1. I'm pretty sure that the problem is *not* with the additional `Gitignore::empty` invocation. The expensive operation is the `matched(&path, is_dir)` call on the four `Gitignore` instances in (every loop iteration inside) `Ignore::matched_ignore`.

2. This PR does not only resolve the performance regression in #835 but improves the speed beyond that (if more than one matcher is disabled).

PR #837, opened 2018-02-26
Add scoop install instructions

alejandro5042:
Scoop, a popular Windows package manager, lists and regularly updates ripgrep. Would be useful to add instructions.


BurntSushi:
Thanks!

PR #844, opened 2018-03-05
doc: add windows scoop installation instruction

h404bi:


BurntSushi:
Looks like #837 beat ya to it! Thanks though!

PR #849, opened 2018-03-08
Added instructions for installing latest release with snap

hackel:
In the snap store, ripgrep 0.8.1 is only available as a candidate release, while the default (stable) release is 0.7.1.


BurntSushi:
Thanks!

PR #867, opened 2018-03-26
Fix bold, intense color on Windows 10 console.

ehuss:
There is an issue with the Windows 10 console where if you issue the bold
escape sequence after one of the extended foreground colors, it overrides the
color.  This happens in termcolor if you have bold, intense, and color set.
The workaround is to issue the bold sequence before the color.

This is for rust-lang/rust#49322.


BurntSushi:
Awesome, thank you!

Issue #74, opened 2016-09-25
Add support for replace in files


kfir-drivenets:
One of the usecases that many people have is a simple search and replace in multiple files.
Usually I achieve this functionality with `xargs sed` for example with `ag -l 'hello' | xargs sed -i "s/hello/bye/g`. 
Since rg contains the replace flag it will be nice to add support for this feature by replacing in the files directly. 

This will be a nice addition after https://github.com/BurntSushi/ripgrep/issues/26

BurntSushi:
I don't think I'd be willing to do this. It would change rg from a tool that will never modify your files to one that will modify your files. It is also a very complicated feature to get right to make sure you don't really mess up files on disk.

Just use `sed`.

samuelcolvin:
Real search and replace would be an absolutely excellent feature.

But I can well believe it would be painful and complex to implement.

Perhaps there could be a reasonable middle way: if `rg` could output in `.patch` file based on the replace argument the user could then review the patch and apply it with they're happy.

Something like:

```shell
rg '(foo) (bar)' --replace '$2 $1' --diff-output > rg.diff
cat rg.diff
...
patch < rg.diff
```

That has the advantage that the user couldn't easily mess up a directory by happening to add an argument and the business of modifying files is left to another command.

BurntSushi:
@samuelcolvin Why would someone do that over using `sed`?

I guess your suggestion does satisfy one of my primary concerns, but it doesn't seem like good UX to me.

samuelcolvin:
Because for me (and lots of other developers) `sed` is a complete anathema, it's not at all easy to get started with or reason with and it's documentation doesn't help much. I'm also not sure how well it deals with unicode.

Most people seem to open up sublime or atom or their IDE when they want to do a search and replace. For me that says a lot about the dearth of decent tools in the terminal for this.

I agree the UX isn't great but since some kind of review step is likely to be required anyway, it couldn't be that much more succinct. If you know what you want you could of course pipe the output of `rg` straight into patch.

(by the way, I'm a big fan of `rg`, sorry to sound like I'm demanding lots highly complex features. The current tool is already great.)

BurntSushi:
@samuelcolvin Thanks for explaining. I can appreciate that `sed` might be an opaque tool, however, its functionality is vast and I really don't want to start down the path of being a `sed` replacement. The patch idea is clever, but I don't think it's a particularly good fit for `ripgrep`.

[...]

PR #870, opened 2018-03-27
doc: add search and replace faq

z2oh:
Adds a brief solution to basic search and replace using ripgrep and sed. I originally wrote a version of this in https://github.com/BurntSushi/ripgrep/issues/74, where @BurntSushi expressed interest in this being added to the FAQ. Thanks for such a powerful tool!


BurntSushi:
@z2oh Thanks so much for this! I added a few touchups but otherwise merged this as-is. :-)

PR #892, opened 2018-04-21
ignore: impl Clone for DirEntry

### LINKED ISSUE #891: ignore crate: DirEntry is not Clone
Not a ripgrep issue, so excuse me _ignoring_ the standard boilerplate.

`ignore::walk::DirEntry` cannot be cloned, which is an issue if I want to use it in [findr](https://github.com/stevedonovan/findr).  `findr` uses the most excellent embedded mini-scripting language Rhai, and Rust structs must be clonable to inter-operate with it.

I've been using a combination of `walkdir` (which does provide a clonable `DirEntry`) and the `gitignore` crate.  In effect, I have to reproduce a (subset) of `ignore` functionality in `findr` - it would be rather nice to offload that to a crate that already knows about directory iteration that can ignore non-essential files. 

Edit: Had a look - not so trivial, since `Error` must be `Clone` and `std::io::Error` is not....

BurntSushi:
There is a small hiccup here in that a `DirEntry` can embed errors
associated with reading an ignore file, which can be accessed and logged
by consumers if desired. That error type can contain an io::Error, which
isn't cloneable. We therefore implement Clone on our library's error
type in a way that re-creates the I/O error as best as possible.

Fixes #891

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,7 +16,7 @@ dependencies = [
 
 [[package]]
 name = "atty"
-version = "0.2.8"
+version = "0.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
  "libc 0.2.40 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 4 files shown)


BurntSushi:
cc @stevedonovan

PR #894, opened 2018-04-24
Added verilog type

jounathaen:
This adds a verilog type (.v & .vh).
This also includes the typical SystemVerilog extensions (.sv & .svh).
System Verilog is an extension to Verilog, but I think you can combine both in one rg type.


BurntSushi:
Awesome, thanks!

PR #895, opened 2018-04-24
ignore: speed up Gitignore::empty

### LINKED ISSUE #835: Small performance regression in ignore-0.4.x

BurntSushi:
This commit makes Gitignore::empty a bit faster by avoiding allocation
and manually specializing the implementation instead of routing it through
the GitignoreBuilder.

This helps improve uses of ripgrep that traverse *many* directories, and
in particular, when the use of ignores is disabled via command line
switches.

Fixes #835, Closes #836

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,8 @@ Bug fixes:
   Fix problems with `--line-number-width` by removing it.
 * [BUG #832](https://github.com/BurntSushi/ripgrep/issues/832):
   Clarify usage instructions for `-f/--file` flag.
+* [BUG #835](https://github.com/BurntSushi/ripgrep/issues/835):
+  Fix small performance regression while crawling very large directory trees.
 * [BUG #851](https://github.com/BurntSushi/ripgrep/issues/851):
   Fix `-S/--smart-case` detection once and for all.
 * [BUG #852](https://github.com/BurntSushi/ripgrep/issues/852):
--- a/globset/src/lib.rs
+++ b/globset/src/lib.rs
@@ -288,6 +288,14 @@ pub struct GlobSet {
 }
 
 impl GlobSet {
+    /// Create an empty `GlobSet`. An empty set matches nothing.
+    pub fn empty() -> GlobSet {
+        GlobSet {
+            len: 0,
+            strats: vec![],
+        }
+    }
+
     /// Returns true if this set is empty, and therefore matches nothing.
     pub fn is_empty(&self) -> bool {
         self.len == 0
--- a/ignore/src/dir.rs
+++ b/ignore/src/dir.rs
@@ -209,7 +209,9 @@ impl Ignore {
     fn add_child_path(&self, dir: &Path) -> (IgnoreInner, Option<Error>) {
         let mut errs = PartialErrorBuilder::default();
         let custom_ig_matcher =
-            {
+            if self.0.custom_ignore_filenames.is_empty() {
+                Gitignore::empty()
+            } else {
                 let (m, err) =
                     create_gitignore(&dir, &self.0.custom_ignore_filenames);
                 errs.maybe_push(err);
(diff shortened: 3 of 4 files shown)


BurntSushi:
@sharkdp I think this should resolve the performance regression. I think my [initial analysis here](https://github.com/BurntSushi/ripgrep/pull/836#issuecomment-372030260) wasn't quite the whole story. In addition to the overhead of `create_gitignore`, it turns out that `Gitignore::empty()` had a fair bit of cost associated with it as well.

Here are the benchmarks (first is current master, second is your PR on top of current master and third is this PR):

```
$ hyperfine -s basic --warmup 10 'rg-baseline -uu --files -g "README.rst"'
Benchmark #1: rg-baseline -uu --files -g "README.rst"
  Time (mean ± σ):     596.9 ms ±   4.7 ms    [User: 3.815 s, System: 3.200 s]
  Range (min … max):   589.8 ms … 604.1 ms

$ hyperfine -s basic --warmup 10 'rg-sharkdp -uu --files -g "README.rst"'
Benchmark #1: rg-sharkdp -uu --files -g "README.rst"
  Time (mean ± σ):     544.0 ms ±   2.8 ms    [User: 3.287 s, System: 3.112 s]
  Range (min … max):   540.3 ms … 548.5 ms

$ hyperfine -s basic --warmup 10 'rg -uu --files -g "README.rst"'
Benchmark #1: rg -uu --files -g "README.rst"
  Time (mean ± σ):     554.9 ms ±   6.8 ms    [User: 3.365 s, System: 3.160 s]
  Range (min … max):   547.7 ms … 566.3 ms
```

One important observation here is that because `-u` is used, the calls to `matched` on the `Gitignore` matchers never actually take place since `has_any_ignore_rules` returns `false`.

Your PR is still a hair faster, but I'm not inclined to it since it spreads case analysis out too much. The likely cause of the performance difference is the additional function calls/copying when calling `Gitignore::empty` as opposed to just using `None`. One possible middle ground is to change `Gitignore` to be `pub struct Gitignore(Option<GitignoreInner>)` which would make the implementation of `Gitignore::empty()` equivalent to your PR without spreading out the case analysis. I'm not inclined to think it is worth it though.

sharkdp:
Very cool - thank you very much for looking into this!

I've run the benchmark again and see no significant difference between my PR and yours.

PR #898, opened 2018-04-27
add support for lz4 decoding

njaard:
Requires the `lz4` binary which I published in
https://crates.io/crates/lz4util. It has semantics
similar to `gzip`.

Add a test for lz4 files.

Update the readme


BurntSushi:
I'm not sure I follow the choices made in this PR. It seems like lz4 already has a standard command line tool: https://github.com/lz4/lz4 It is even already installed on my system as a core package (Archlinux). If we are going to add support for it, then it should be against the standard binary and not a binary released on crates.io yesterday.

njaard:
Since the `lz4` command that you cited is compatible with my lz4 (for decompression purposes) due to `gzip` compatibility, I can simply remove the reference to my crate in the manual. Is that acceptable?

BurntSushi:
@njaard OK, yes, that's fine. It wasn't clear from your PR whether your binary was compatible with the standard one. But yes, might as well remove the crate reference, which I think would be confusing.

Also, please install the standard lz4 binary in CI so that the test for it runs.

README.md
```diff
@@ -106,7 +106,7 @@ increases the times to `2.640s` for ripgrep and `10.277s` for GNU grep.
   automatically detecting UTF-16 is provided. Other text encodings must be
   specifically specified with the `-E/--encoding` flag.)
 * ripgrep supports searching files compressed in a common format (gzip, xz,
-  lzma or bzip2 current) with the `-z/--search-zip` flag.
+  lzma, bzip2 or [lz4](https://crates.io/crates/lz4util)) with the `-z/--search-zip` flag.
```
BurntSushi: Please follow the prevailing format of this file and wrap lines to 79 columns inclusive.

njaard:
* I removed the reference to my crate
* The line wraps to 79 characters, consequently
* added `liblz4-tool` as a dependency for travis
* found a few more instances to document lz4 support

src/app.rs
```diff
@@ -1436,7 +1436,7 @@ This flag can be used with the -o/--only-matching flag.
 fn flag_search_zip(args: &mut Vec<RGArg>) {
     const SHORT: &str = "Search in compressed files.";
     const LONG: &str = long!("\
-Search in compressed files. Currently gz, bz2, xz, and lzma files are
+Search in compressed files. Currently gz, bz2, xz, lzma, and lz4 files are
```
mqudsi: Can we sort this list alphabetically, just like it is in `FAQ.md`?

(Sorry for being so OCD!)

njaard: I kept the original ordering because it seemed to be by popularity

njaard:
@BurntSushi Just a quick reminder that I believe my pull request has satisfied all of your concerns.

njaard:
Great!

Also, they were your own Oxford commas :)

PR #899, opened 2018-04-29
Add support for commonly used files with Puppet

strangelittlemonkey:
Puppet is primarily written in it's own format of .pp files, but
custom facts and functions are often written in Ruby. The templating
language is ERB and so this will allow scanning of any of the three
most commonly used formats for Puppet specific things.

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -215,6 +215,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("pod", &["*.pod"]),
     ("protobuf", &["*.proto"]),
     ("ps", &["*.cdxml", "*.ps1", "*.ps1xml", "*.psd1", "*.psm1"]),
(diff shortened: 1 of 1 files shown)


ignore/src/types.rs
```diff
@@ -215,6 +215,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("pod", &["*.pod"]),
     ("protobuf", &["*.proto"]),
     ("ps", &["*.cdxml", "*.ps1", "*.ps1xml", "*.psd1", "*.psm1"]),
+    ("puppet", &[".erb", "*.pp", "*.rb"]),
```
CYBAI: Does the `.erb` miss a `*` here?

BurntSushi: Sigh. Yup. Fixed on master.

strangelittlemonkey: :(  I'm sorry, it was late.

PR #906, opened 2018-05-03
Update clap for better PowerShell completion

lzybkr:
When generating completions for PowerShell, the latest version of clap will use the short parameter help. 

PowerShell ISE, VSCode w/ the PowerShell extension, and PSReadLine 2.0.0 will all display this help when completing.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -51,7 +51,7 @@ same-file = "1"
 termcolor = { version = "0.3.4", path = "termcolor" }
 
 [dependencies.clap]
-version = "2.29.4"
+version = "2.31.2"
 default-features = false
 features = ["suggestions", "color"]
 


BurntSushi:
I'm not sure I understand this change. The `Cargo.lock` already indicates that the latest version of clap is being used.

lzybkr:
More like I don't understand Cargo. I made an incorrect assumption, other dependency resolvers I've worked with wouldn't have used the version I was expecting.

It doesn't seem harmful, but feel free to close if you don't want to merge.

BurntSushi:
Ah I see. Mostly, what I wanted to know is if you were observing the problem with the auto completion files even though the latest version of clap was being used.

But yeah, I think the idea here is that `Cargo.toml` specifies the minimum version of clap that ripgrep works with, for flexibility. With that said, CI doesn't run against the minimum version, so the numbers in the `Cargo.toml` become somewhat misleading.

I'm going to close this out for now, but I do expect to be bumping the minimum clap version at some point soon, since it has new features I'd like to use.

Thanks!

PR #911, opened 2018-05-04
Documentation Update

### LINKED ISSUE #910: Ignore Crate Case Insensitive
This is not directly related to ripgrep, hence the template is missing.

When using the `ignore` crate, I wanted to add a set of overrides using the `OverrideBuilder` type. When testing case-insensitivity, I noticed that it only worked when I called `case_insensitive(true)` before adding any overrides. I don't know if this is an issue or just something that should be documented in `OverrideBuilder`.

If this is a simple doc fix, I would be happy to submit a patch. Thanks!

gsquire:
A small documentation update for `OverrideBuilder` and `GitignoreBuilder`.

Closes #910

### DIFF
--- a/ignore/src/overrides.rs
+++ b/ignore/src/overrides.rs
@@ -140,6 +140,8 @@ impl OverrideBuilder {
 
     /// Toggle whether the globs should be matched case insensitively or not.
     /// 
+    /// When this option is changed, only globs added after the change will be affected.
+    ///
     /// This is disabled by default.
     pub fn case_insensitive(
         &mut self, yes: bool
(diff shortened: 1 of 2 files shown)


ignore/src/overrides.rs
```diff
@@ -140,7 +140,8 @@ impl OverrideBuilder {
 
     /// Toggle whether the globs should be matched case insensitively or not.
     /// 
-    /// This is disabled by default.
+    /// This is disabled by default. In order for case sensitivity to work, this function
+    /// must be called before any overrides are added.
```
BurntSushi: I think we can be more precise here:

```
/// When this option is changed, only globs added after the change will be affected.
///
/// This is disabled by default.
```

Also, can you add this to the gitignore builder too?

PR #920, opened 2018-05-13
Add details of BSD packages to README

wezm:


README.md
```diff
@@ -270,6 +270,24 @@ seem to work right and generate a number of very strange bug reports that I
 don't know how to fix and don't have the time to fix. Therefore, it is no
 longer a recommended installation option.)
 
+If you're a **FreeBSD** user, then you can install ripgrep from the [official ports](https://www.freshports.org/textproc/ripgrep/):
+
+```
+# pkg install ripgrep
```
BurntSushi: Can you use a `$` here? Or is the `#` intentional? Is it implying that the user should be `root`? Is it standard BSD convention to write it this way instead of `sudo`?

(Apologies, I am not experienced with BSDs or BSDisms. :-))

wezm: `#` was intentional to suggest a root shell. `sudo` is optional on NetBSD and FreeBSD so I wrote these examples without assuming it wasn't installed. It follows the [convention used in the FreeBSD handbook](https://www.freebsd.org/doc/en_US.ISO8859-1/books/handbook/book-preface.html#preface-conv-examples). OpenBSD has their `sudo` replacement `doas` in the base system so I used that in their example.

wezm:
> I had no idea ripgrep was in this many BSD package repos.

Yep it's there along with a [handful of other Rust tools](https://www.freshports.org/search.php?stype=depends_all&method=match&query=lang%2Frust&num=100&orderby=category&orderbyupdown=asc&search=Search&format=html&minimal=1&branch=head). :slightly_smiling_face:

README.md
BurntSushi: Ah awesome! I'm happy as long as this is the convention.

PR #925, opened 2018-05-24
Add MacPorts installation instructions to README

ylluminarious:
I just saw the recent addition of BSD installation instructions to the README file. So I thought, what the heck, why not add installation instructions for Macports as well?  :smiley:

This is a tiny change, so I think it should be okay to merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,7 +197,8 @@ $ brew tap burntsushi/ripgrep https://github.com/BurntSushi/ripgrep.git
 $ brew install ripgrep-bin
 ```
 
-If you're a **MacPorts** user, then you can install ripgrep from the [official ports](https://www.macports.org/ports.php?by=name&substr=ripgrep):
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -197,6 +197,12 @@ $ brew tap burntsushi/ripgrep https://github.com/BurntSushi/ripgrep.git
 $ brew install ripgrep-bin
 ```
 
+If you're a **MacPorts** user, then you can install ripgrep from the [official ports](https://www.macports.org/ports.php?by=name&substr=ripgrep):
```
BurntSushi: Could you please wrap this to 79 columns inclusive? If the URL is too long, then just put the `[...](...)` on its own line.

ylluminarious:
@BurntSushi No problem! Is it acceptable now?

ylluminarious:
@BurntSushi Thank you!

Issue #416, opened 2017-03-21
Carriage return handling on Windows


BatmanAoD:
On Windows, I noticed two surprising things about carriage returns (`\r`):

 * They aren't treated as word-boundaries (i.e. they won't cause the `\b` anchor to match). This means that `rg -w foo` fails to find `foo` at the end of a line in a file with Windows line-endings.
 * They aren't included in the end-of-line anchor `$`, which makes using the anchor more difficult.

I realize that this *could* be by design, since RipGrep may be using some standard definition of `\b` that explicitly does not include `\r`. That seems unintuitive, though, and I don't think I've seen that behavior in other regex engines running on Windows (e.g. in Python on Windows).

BurntSushi:
> They aren't treated as word-boundaries (i.e. they won't cause the \b anchor to match). This means that rg -w foo fails to find foo at the end of a line in a file with Windows line-endings.

This doesn't seem right. Could you please provide a minimal reproducible example? e.g., This works fine for me:

```
$ echo -e 'foo\n' | rg -w foo
1:foo
$ echo -e 'foo\r\n' | rg -w foo
1:foo
```

> They aren't included in the end-of-line anchor `$`, which makes using the anchor more difficult.

I think GNU grep handles this by stripping `\r` characters and dealing with the fallout. This seems hard, or at best, very annoying.

The semantics of the regex engine aren't going to change, sorry.

BatmanAoD:
Hm, you're right, I'm having trouble reproducing the `\b` issue with test files; I'll have to see if I can find out why I'm having issues with the file I was searching when I discovered the issue.

Regarding `$`, though, isn't the anchor essentially useless on Windows if `\r` causes a mismatch? `echo -e "foo\r\n" | <greptool> 'foo$'` seems to me like it should *always* match, regardless of the system. (I've checked that on Debian, GNU grep and ack both handle this case appropriately.)

Why not just let the anchor match when a character other than `\r` is followed by `\n` *or* `\r\n`? (I think you're right that this approach is not the one taken by grep, so I realize that there may be a good reason not to take this approach.)

I'm guessing that this behavior is due to the current behavior of rust-lang/regex ? If so, I can open an issue there, which seems like a more appropriate place (even though you're the primary author/contributor on that project as well).

BurntSushi:
The issue already exists: https://github.com/rust-lang/regex/issues/244

> Why not just let the anchor match when a character other than \r is followed by \n or \r\n?

[...]

PR #929, opened 2018-05-28
(Help needed) Windows eol anchor option

BatmanAoD:
Potential fix for #416 via a command-line option.

I decided it would be a good time to implement my old suggestion of just providing a command-line argument to modify the search pattern(s) now that there is config file support.

I've tested by hand that this works as expected on Windows, and the regression test I've added works on Linux.

However, the regression test fails on Windows with no information other than that the exit-code is `1`. Is there some problem with the test framework?

Here's the complete output from running the test on Windows:

```
   Compiling ripgrep v0.8.1 (file:///C:/Users/kjstrand/personal-projects/ripgrep)
    Finished dev [unoptimized + debuginfo] target(s) in 7.2 secs
     Running C:\Users\kjstrand\personal-projects\ripgrep\target\debug\deps\rg-275c12485214083c.exe

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 86 filtered out

     Running C:\Users\kjstrand\personal-projects\ripgrep\target\debug\deps\integration-3d1ca51b299d90b8.exe

running 1 test
test regression_416 ... FAILED

failures:

---- regression_416 stdout ----
        thread 'regression_416' panicked at '

==========
command failed but expected success!

Did your search end up with no results?

command: "C:\\Users\\kjstrand\\personal-projects\\ripgrep\\target\\debug\\deps\\../rg.exe" "--eol-anchor-include-cr" "--with-filename" "bar$"
cwd: C:\Users\kjstrand\personal-projects\ripgrep\target\debug\deps\ripgrep-tests\regression_416\0

status: exit code: 1

stdout:

stderr:

==========
', tests\workdir.rs:243:13
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    regression_416

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 157 filtered out
```


BatmanAoD:
There are also some stylistic issues you will probably want to weigh in on, e.g. whether the flag name is appropriate.

BurntSushi:
Thanks for working on this! I'm not sure when I'll get a chance to look at this. I'm also not 100% sold on this particular approach.

BatmanAoD:
I don't think it's ideal, which is why I hid it behind a long feature flag. But I think it's a lot better than nothing; the current behavior renders `$`

BatmanAoD:
It appears that on Windows, but on no other supported platform, the test framework requires one or more explicit paths to search (e.g. using `cmd.arg("./")`. So I've fixed the test that was failing.

src/args.rs
```diff
@@ -598,6 +600,20 @@ impl<'a> ArgMatches<'a> {
...
 
+    /// Returns the given pattern with `\r?` inserted before each `$` if the
+    /// --eol-anchor-include-cr flag is set. Otherwise, the pattern is
+    /// returned unchanged.
+    fn cr_anchr_pattern(&self, pat: String) -> String {
+        if self.is_present("eol-anchor-include-cr") {
+            // If `$` occurs at the end (or beginning), an empty
+            // split-element will be created, causing the `join` to
+            // perform the desired insertion at the correct location.
+            pat.split('$').collect::<Vec<&str>>().join(r"\r?$")
```
BurntSushi: This should be done by parsing the regex and doing a substitution at the HIR level.

BatmanAoD: @BurntSushi That seems sensible. Are you more amenable to the configuration-option design now, then?

BurntSushi: Not particularly... I just happened to see this particular issue in passing.

Honestly, this just isn't a priority for me now, and it needs thought.

BatmanAoD: I understand it's not a priority for you, but as I mentioned when I opened the original issue, it makes `$` essentially useless on Windows.

I agree that the *right* solution requires thought, and I can understand why you wouldn't want to introduce a half-baked solution just to deprecate it later. On the other hand, my suggested solution is extremely simple to implement and, I expect, to maintain, and personally I think it is reasonable to introduce command-line options that may be deprecated in favor of a better solution with a warning that this is the case.

I suppose that if you're definitely opposed to a partial solution, I can just continue to maintain my fork until there's a better solution ready to ship, but I don't want to publish a competing crate on Cargo or anything like that, so it's not terribly convenient.

BatmanAoD: Okay, I didn't know that. That's helpful; I'll try that out. Thank you.

PR #933, opened 2018-05-30
clippy: Fix some clippy lints

rumpelsepp:
I ran clippy and fixed some lints. :)

### DIFF
--- a/globset/src/lib.rs
+++ b/globset/src/lib.rs
@@ -449,7 +449,7 @@ impl GlobSet {
 
 /// GlobSetBuilder builds a group of patterns that can be used to
 /// simultaneously match a file path.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
 pub struct GlobSetBuilder {
     pats: Vec<Glob>,
 }
@@ -494,7 +494,7 @@ impl<'a> Candidate<'a> {
     /// Create a new candidate for matching from the given path.
     pub fn new<P: AsRef<Path> + ?Sized>(path: &'a P) -> Candidate<'a> {
         let path = path.as_ref();
-        let basename = file_name(path).unwrap_or(OsStr::new(""));
+        let basename = file_name(path).unwrap_or_else(|| OsStr::new(""));
         Candidate {
             path: normalize_path(path_bytes(path)),
             basename: os_str_bytes(basename),
@@ -567,7 +567,7 @@ impl LiteralStrategy {
     }
 
     fn add(&mut self, global_index: usize, lit: String) {
-        self.0.entry(lit.into_bytes()).or_insert(vec![]).push(global_index);
+        self.0.entry(lit.into_bytes()).or_insert_with(|| vec![]).push(global_index);
     }
 
     fn is_match(&self, candidate: &Candidate) -> bool {
@@ -591,7 +591,7 @@ impl BasenameLiteralStrategy {
     }
 
     fn add(&mut self, global_index: usize, lit: String) {
-        self.0.entry(lit.into_bytes()).or_insert(vec![]).push(global_index);
+        self.0.entry(lit.into_bytes()).or_insert_with(|| vec![]).push(global_index);
     }
 
     fn is_match(&self, candidate: &Candidate) -> bool {
@@ -621,7 +621,7 @@ impl ExtensionStrategy {
     }
 
     fn add(&mut self, global_index: usize, ext: String) {
-        self.0.entry(ext.into_bytes()).or_insert(vec![]).push(global_index);
+        self.0.entry(ext.into_bytes()).or_insert_with(|| vec![]).push(global_index);
     }
 
     fn is_match(&self, candidate: &Candidate) -> bool {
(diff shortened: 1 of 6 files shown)


globset/src/lib.rs
```diff
@@ -449,7 +449,7 @@ impl GlobSet {
 
 /// GlobSetBuilder builds a group of patterns that can be used to
 /// simultaneously match a file path.
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
```
BurntSushi: This is a change to the API. It seems fine in principle. But could you please remove the `derive(Default)` and write out an explicit `Default` impl that calls `GlobSetBuilder::new()`? This way, we make sure the implementations are exactly the same.

globset/src/lib.rs
```diff
@@ -494,7 +494,7 @@ impl<'a> Candidate<'a> {
     /// Create a new candidate for matching from the given path.
     pub fn new<P: AsRef<Path> + ?Sized>(path: &'a P) -> Candidate<'a> {
         let path = path.as_ref();
-        let basename = file_name(path).unwrap_or(OsStr::new(""));
+        let basename = file_name(path).unwrap_or_else(|| OsStr::new(""));
```
BurntSushi: I don't agree with this change. The status quo is simpler and shorter.

globset/src/lib.rs
```diff
@@ -567,7 +567,7 @@ impl LiteralStrategy {
     }
 
     fn add(&mut self, global_index: usize, lit: String) {
-        self.0.entry(lit.into_bytes()).or_insert(vec![]).push(global_index);
+        self.0.entry(lit.into_bytes()).or_insert_with(|| vec![]).push(global_index);
```
BurntSushi: I'm not sure I agree with this change either, nor the other uses of `or_insert_with`.

globset/src/pathutil.rs
```diff
@@ -7,9 +7,9 @@ use std::path::Path;
 /// If the path terminates in ., .., or consists solely of a root of prefix,
 /// file_name will return None.
 #[cfg(unix)]
-pub fn file_name<'a, P: AsRef<Path> + ?Sized>(
-    path: &'a P,
-) -> Option<&'a OsStr> {
+pub fn file_name<P: AsRef<Path> + ?Sized>(
+    path: &P,
+) -> Option<&OsStr> {
```
BurntSushi: Please add the explicit lifetime parameter back. :-)

globset/src/pathutil.rs
```diff
@@ -32,9 +32,9 @@ pub fn file_name<'a, P: AsRef<Path> + ?Sized>(
 /// If the path terminates in ., .., or consists solely of a root of prefix,
 /// file_name will return None.
 #[cfg(not(unix))]
-pub fn file_name<'a, P: AsRef<Path> + ?Sized>(
-    path: &'a P,
-) -> Option<&'a OsStr> {
+pub fn file_name<P: AsRef<Path> + ?Sized>(
+    path: &P,
+) -> Option<&OsStr> {
```
BurntSushi: Please add the explicit lifetime parameter back. :-)

grep/src/literals.rs
```diff
@@ -233,9 +233,9 @@ fn count_unicode_class(cls: &hir::ClassUnicode) -> u32 {
     cls.iter().map(|r| 1 + (r.end() as u32 - r.start() as u32)).sum()
 }
 
-/// Return the number of bytes in the given class.
+/// Return the number of bytes in the givep class.
```
BurntSushi: Typo?

grep/src/literals.rs
```diff
@@ -233,9 +233,9 @@ fn count_unicode_class(cls: &hir::ClassUnicode) -> u32 {
...
 fn count_byte_class(cls: &hir::ClassBytes) -> u32 {
-    cls.iter().map(|r| 1 + (r.end() as u32 - r.start() as u32)).sum()
+    cls.iter().map(|r| 1 + (u32::from(r.end()) - u32::from(r.start()))).sum()
```
BurntSushi: I don't understand this change. Can you justify it?

globset/src/lib.rs
academician: Can I ask why you disagree with it? It's a micro-optimization, but lazily creating the OsStr seems better to me. Eg: https://godbolt.org/g/YyXgWQ

grep/src/literals.rs
academician: IIUC, `as` can perform lossy conversions, whereas `From` is only implemented for lossless ones, so if the type of `r.end()` and `r.start()` changed to something like `u64` it would truncate. In this case that would almost certainly never happen, so it's unlikely to be a real problem here, but I think the idea is that it's usually the safer default so clippy lints against it.

globset/src/lib.rs
BurntSushi: I don't disagree with it per se. I just think the change should be justified. If you can produce a useful benchmark in the context of `globset` that shows a meaningful difference, then I'd be happy to reconsider it.

grep/src/literals.rs
BurntSushi: `ClassByteRange` is always going to be a byte range. But I don't feel that strongly. Happy to keep this.

globset/src/lib.rs
BurntSushi: RE your godbolt link: that sort of evidence is _basically_ meaningless to me because it's measuring the wrong thing. The codegen for globset's code in particular would be better, and a real benchmark is best.

BurntSushi:
@rumpelsepp Thanks for submitting this PR, but given the inactivity and my general disagreement with a lot of these lints, I think I'm just going to close this out.

With Clippy headed towards stable Rust, it's possible that I will capitulate and start running Clippy in CI along with a specific whitelist/blacklist of lints that I do and don't like. But until then, I think ad hoc fixes from Clippy probably just aren't worth it.

rumpelsepp:
No problem, sorry for being inactive on this. Just being quite busy the last weeks and I forgot about this... :) As you mentioned, maybe it makes sense to rely on clippy once it is available on stable. There is already some activity: https://internals.rust-lang.org/t/clippy-is-available-as-a-rustup-component/7967

PR #949, opened 2018-06-15
ignore/types: add support for a few Android platform file types

amhk:
Add the following file types to the built-in type list:

- Android AIDL files (*.aidl)
- Android platform makefiles (*.mk, *.bp)

Note: I went with "amake" as the name for the Android makefile pattern because it's quick to write. One could argue that something like "android-make" is easier to comprehend when skimming through the output of --type-list.


amhk:
>>Also, can I just say, it makes me unreasonably happy to see you following my commit message conventions. :-)

You're welcome! Thanks for creating and maintaining a great tool.

PR #952, opened 2018-06-16
simplify main error handling by using the new return Result from main feature

rtaycher:
return Result from main/increase minimum version to 1.26

Is printing `Error: StringError("no matches")` on no matches ok or should I put back `process::exit(1)` and not print anything?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -59,13 +59,13 @@ matrix:
     # Minimum Rust supported channel. We enable these to make sure ripgrep
     # continues to work on the advertised minimum Rust version.
     - os: linux
-      rust: 1.20.0
+      rust: 1.26.0
       env: TARGET=x86_64-unknown-linux-gnu
     - os: linux
-      rust: 1.20.0
+      rust: 1.26.0
       env: TARGET=x86_64-unknown-linux-musl
     - os: linux
-      rust: 1.20.0
+      rust: 1.26.0
       env: TARGET=arm-unknown-linux-gnueabihf GCC_VERSION=4.8
       addons:
         apt:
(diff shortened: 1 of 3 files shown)


BurntSushi:
Thanks for the PR! I appreciate that you're trying to improve things here, but there are several issues with this PR, which is why I'm just going to close this out. Those issues are:

* Compare the output of `rg foo --colors 'invalid'` on master and in this PR. Your PR shows the error as a debug representation. This is unacceptable for end users, and this is why `result-in-main`, as it currently exists, is not allowed in my projects.
* The simplification being made here is, IMO, nominal.
* Increasing the minimum Rust version to the _latest stable release_ isn't acceptable. While ripgrep doesn't have an official policy on this matter, requiring the latest stable release is definitely not the right thing to do.
* Printing an error message when no results are found represents a *significant behavioral change*. I can't imagine any reality in which I would accept a change like that, but in general, for significant changes you really need to open an issue so that it can be properly discussed.
* Your branch name starts with a `U+0096 START OF GUARDED AREA` character, which makes it unreasonably difficult to manage.

rtaycher:
thank you fro the explanation

rtaycher:
> Increasing the minimum Rust version to the latest stable release isn't acceptable. While ripgrep doesn't have an official policy on this matter, requiring the latest stable release is definitely not the right thing to do.

I'm sorry about this one, it may have been obvious but maybe it should be mentioned/semi-official (don't increase the minimum version to last stable release/last 2 stable releases)? I tried to find anything on ripgreps version policy in the rpo but could only find that it had been incremented in jumps by looking at the git history.

BurntSushi:
The only real policy is "be conservative, within reason." I don't really know how to articulate it more than that. It's a judgment call.

Issue #278, opened 2016-12-11
filters don't work correctly when a search path is specified


nu4nu:
ripgrep seems to ignore unintended files when the base directory is not the current directory.

How to reproduce:

```
% mkdir -p ../test/foo
% echo '.*' >../test/.gitignore
% echo baz >../test/foo/bar

% rg baz ../test
../test/foo/bar
1:baz

% rg baz ../test/foo
No files were searched, which means ripgrep probably applied a filter you didn't expect. Try running again with --debug.
```

I found the following lines in `ignore/src/dir.rs` don't make sense because `path` on the right side is a relative path from the current directory, not from `abs_parent_path`.
```
339         if let Some(abs_parent_path) = self.absolute_base() {
340             let path = abs_parent_path.join(path);
```

A possible fix to run the example above correctly is the following.  However, I'm not sure this fixes other cases.
```
let path = path.canonicalize().unwrap()
```

BurntSushi:
This is a tough one to fix.

First and foremost, the proposed solution can never work. Canonicalizing candidates paths does seems like the obvious solution---and I've been tempted to do it because it simplifies so much---but it has two significant problems:

1. Canonicalization erases symlinks and ignore files are symlink aware. If we erased symlinks, lots of ignore files would break.
2. Canonicalization is expensive.

Here are the key values in this particular case, assuming we've reproduced your test case inside `/tmp/ripgrep-278`, and searching from inside `/tmp/ripgrep-278/nested` for `../test/foo`:

* The absolute base path is `/tmp/ripgrep-278/test/foo`, which is derived by canonicalizing the search path `../test/foo`.
* The root directory that sets the context for the `.gitignore` file is `/tmp/ripgrep-278/test`.
* The candidate file path is `../test/foo/bar`.
* The path that we actually search is `foo/../test/foo/bar`, which is generated by joining the absolute base path (`/tmp/ripgrep-278/test/foo`) with the candidate file path (`../test/foo/bar`) and then stripping the gitignore directory path prefix (`/tmp/ripgrep-278/test`).

That last step is where things go bad. In particular, the text that we *actually* match should be just `foo/bar`.

More generally, relative components like `.` and `..` should probably never be present when doing a match, which potentially suggests that we need a `realpath` that does what it does today, but leaves symlinks in tact. However, doing that for every single file path is a non-starter. It would be too expensive.

[...]

PR #962, opened 2018-06-24
Use cwd to resolve paths in Ignore::matched

### LINKED ISSUE #829: Passing a path argument doesn't respect .gitignore
ripgrep 0.8.1
-SIMD -AVX

Archlinux (4.15.3-2-ARCH), also Ubuntu 16.04

If I am in a git repository and run `rg` with a directory as a PATH argument, then all files under that directory are searched, even if they would normally be ignored because of .gitignore rule. If I `cd` into the directory and run `rg` without a PATH argument, the files are excluded as I expect. changing into a subdirectory is a workaround but is sometimes less convenient than using a path argument.

The man page does say

> Paths specified expicitly on the command line override glob and ignore rules.

Which could be interpreted to mean that the if such an argument is specified, the ignore rules are ignored altogether. But that behaviour is unexpected to me, and the wording of the documentation isn't entirely clear on whether it overrides _all_ rules, or only rules that specifically ignore the specified path (but not children of that path).

If overriding all ignore rules is the intended behaviour there should at least be a way to turn the ignore rules back on, but `--ignore-vcs` does nothing when supplied with a PATH argument.

Interestingly, if I specify a glob such as `-g "*.c"`, that is respected for files under the path.

#### Steps to reproduce:

.gitignore:
```
/a/b/
```

a/b/test.txt:
```
Sample text
```

In root of repository:

`rg Sample` produces no output

```
$ rg Sample a
a/b/test.txt
1:Sample text
```

Adding `--ignore-vcs` produces the same output.

In `a` directory `rg Sample` also produces no output.

`rg --debug Sample a`

```
DEBUG/grep::search/grep/src/search.rs:195: regex ast:
Literal {
    chars: [
        'S',
        'a',
        'm',
        'p',
        'l',
        'e'
    ],
    casei: false
}
DEBUG/grep::literals/grep/src/literals.rs:38: literal prefixes detected: Literals { lits: [Complete(Sample)], limit_size: 250, limit_class: 10 }
DEBUG/globset/globset/src/lib.rs:396: glob converted to regex: Glob { glob: "**/.*~", re: "(?-u)^(?:/?|.*/)\\..*\\~$", opts: GlobOptions { case_insensitive: false, literal_separator: false }, tokens: Tokens([RecursivePrefix, Literal('.'), ZeroOrMore, Literal('~')]) }
DEBUG/globset/globset/src/lib.rs:401: built glob set; 0 literals, 0 basenames, 1 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 1 regexes
DEBUG/globset/globset/src/lib.rs:401: built glob set; 1 literals, 0 basenames, 0 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 0 regexes
a/b/test.txt
1:Sample text
```

Produced no output, since a/b/test.txt is ignored by the .gitignore file.

tmccombs:
Fixes #829 and #278.

This does change the Ignore struct to depend on the working directory at
the time of creation, I _think_ this is fine, since Ignore isn't
publicly accessible and the Walk structs already depend on the current
working directory implicitly.

I tried to make a minimal change to fix the issue. An alternative
implementation could be to call `current_dir` in the `matched_ignore`
method instead of in `add_parents`, but I'm not sure of the performance
implications of doing that.

Another possible solution would be to change the places that we call
`Ignore::matched` to change the path to be relative to the absolute_base
of the `Ignore`.


BurntSushi:
https://github.com/BurntSushi/ripgrep/pull/963#issuecomment-480499071

PR #965, opened 2018-06-25
Add --fixed-strings flag (fixes #964)

### LINKED ISSUE #964: add a --no-fixed-strings flag

ripgrep 0.8.1
-SIMD -AVX

Homebrew

macOS 10.13.5

I rarely ever search using regular expressions. Normally I'm wanting to find some usage of a function or string. I was hoping there was a way to set `rg --fixed-strings` as the default behavior. Perhaps with a `.rgrc` or an environment variable?

Here's what I had in mind.

```sh
$ cat ~/.rgrc 
{
  "defaultSearchBehavior": "fixed-strings"
}
$ rg ".i8n(" # Would not explode
```

okdana:
Simple as this probably? I didn't add a test since i don't see any for other options like this, but let me know if you'd like one.

I also didn't update the completion function. I guess we need to decide whether it's useful to show all of the `--no-*` options; i'm leaning towards not usually. But i should add them as hidden options at least so that the exclusions are accurate. I'll do that eventually...

(Fixes #964)

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -909,9 +909,17 @@ fn flag_fixed_strings(args: &mut Vec<RGArg>) {
 Treat the pattern as a literal string instead of a regular expression. When
 this flag is used, special regular expression meta characters such as .(){}*+
 do not need to be escaped.
+
+This flag can be disabled with --no-fixed-strings.
 ");
     let arg = RGArg::switch("fixed-strings").short("F")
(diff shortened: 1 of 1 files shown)


BurntSushi:
@okdana Hah whoops. Missed that! Should be fine!

jsatk:
Thanks for this.

Issue #690, opened 2017-11-23
^$ does not work when using --no-mmap


ericbn:
I found this while running the examples described at #441 

```
$ cat scratch
a
b

c


d
$ rg 'a' scratch
1:a
$ rg 'b' scratch
2:b
$ rg 'c' scratch
4:c
$ rg 'd' scratch
7:d
$ rg --mmap '^$' scratch
3:
5:
6:
8:
$ rg --no-mmap '^$' scratch
2:
4:
6:
7:
9:
$
```

By "does not work" I mean using `--no-mmap` gives an unexpected number of results and unexpected line numbers. I get these same unexpected results by running `rg '^$' scratch`.

I'm using ripgrep 0.7.1 under macOS.

BurntSushi:
I suspect this can be rolled up into the same bug as #441.

dennis-benzinger-hybris:
It's similar to #441 but the interesting point here is that phantom empty lines are also found elsewhere in the file. For example between line 1 and 2 (then 3). Try with `-C1`.

BurntSushi:
I believe this bug should be fixed in libripgrep.

PR #970, opened 2018-07-03
Improve zsh completion

okdana:
I'd been meaning to update this, and i think it's in a state now where it's much better than it was.

I propose that we give this a few days to sit before merging, just to make sure i haven't messed anything up (it is a decent number of lines changed). It's been working for me so far in zsh 5.3 and 5.5, though. If any zsh users watching the repo care to test, that would be pretty cool.

Full details:

* I've updated the function to use spec grouping with `_arguments`. This makes it so much easier (IMO) to read and maintain option exclusions. The down side is that grouping is only available in zsh 5.4+ (released in 2017); on older versions of zsh i simply strip the groups out. This makes option exclusivity pretty inaccurate on those older versions, but i think it's worth it — it's not like it breaks anything, it just continues to show options that might not be useful with options you've already supplied

* I mentioned previously that i wasn't sure whether to complete rarely used 'negation' options like `--messages` and `--no-fixed-strings`. I decided i would complete them, but only if it seems like the user is specifically requesting them, or if they've enabled a style requesting that they always be completed. So if you do `rg --`<kbd>Tab</kbd> you won't get the million `--no-whatever` options, but you will if you do `rg --n`<kbd>Tab</kbd>. To be clear, this only affects the (mostly newer) options that were added solely to negate the effects of other options; stuff that's more generally useful like `--no-heading` and `--no-mmap` is still completed unconditionally

* I've improved the accuracy of option exclusivity (for zsh 5.4+)

* I've improved several option/optarg descriptions

* I've improved completion of colour 'whens' (`always`, `never`, &c.) — they now have descriptions

* I've improved completion of colour specs, in particular the handling of the colons. This was the biggest irritation for me with previous iterations of the function tbh

* I've removed some unnecessary work-arounds

* I've generally replaced some uncommon/undesirable constructs by more idiomatic ones (`print`, quoting stuff, tag names, `ret` to support users with `prefix-needed` off, &c.)

ETA: The CI script failed on some of the AppVeyor platforms because they had an older version of zsh than i expected. Hopefully fixed.

### DIFF
--- a/ci/test_complete.sh
+++ b/ci/test_complete.sh
@@ -1,41 +1,42 @@
 #!/usr/bin/env zsh
 
+emulate zsh -o extended_glob -o no_function_argzero -o no_unset
+
 ##
 # Compares options in `rg --help` output to options in zsh completion function
 
-emulate -R zsh
-setopt extended_glob
-setopt no_function_argzero
-setopt no_unset
-
 get_comp_args() {
+    # Technically there are many options that the completion system sets that
+    # our function may rely on, but we'll trust that we've got it mostly right
     setopt local_options unset
 
     # Our completion function recognises a special variable which tells it to
     # dump the _arguments specs and then just return. But do this in a sub-shell
     # anyway to avoid any weirdness
     ( _RG_COMPLETE_LIST_ARGS=1 source $1 )
-    return $?
 }
 
 main() {
     local diff
-    local  rg="${${0:a}:h}/../target/${TARGET:-}/release/rg"
-    local _rg="${${0:a}:h}/../complete/_rg"
+    local  rg="${0:a:h}/../target/${TARGET:-}/release/rg"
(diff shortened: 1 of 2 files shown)


BurntSushi:
@okdana Thanks for doing this! As to your questions... They are good ones. My feeling is that the interaction between `--vimgrep` and `--passthru` will _probably_ be fixed in libripgrep (I haven't investigated, but it could be related to #441 and/or #690).

As for `--heading` and `--pretty`... They seem like conflicting options to me. I guess in my mind `--vimgrep` always intended to match a specific format to be read by vim itself, and options like `--heading` and `--pretty` kind of negate that. But yeah, maybe they should just work since it seems like there exists sensible behavior there.

In the course of working on libripgrep, I'm going to try and revisit the interactions between flags because it's going to get worse with the rise of support for additional output formats (thinking Ackmate and JSON lines).

Issue #918, opened 2018-05-11
Can't search in ZIP and tar.gz files


kenorb:
#### What version of ripgrep are you using?

```
$ rg --version
ripgrep 0.8.1
-SIMD -AVX
```

#### What operating system are you using ripgrep on?

```
$ macosver 
ProductBuildVersion 17C2120
ProductCopyright 1983-2017 Apple Inc.
ProductName Mac OS X
ProductUserVisibleVersion 10.13.2
ProductVersion 10.13.2
```

#### If this is a bug, what are the steps to reproduce the behavior?

```
$ echo test > test
$ zip test.zip test
  adding: test (stored 0%)
$ tar zcvf test.tar.gz test
a test
$ gzip test
$ rg -z test
./test.gz
1:test
 $ rg -zl test
test.gz
```

#### What is the actual behavior?

No `.zip` or `.tar.gz` are searched.

#### If this is a bug, what is the expected behavior?

Should search in `.zip` and `.tar.gz` files.

BurntSushi:
The documentation states:

```
Search in compressed files. Currently gz, bz2, xz, and lzma files are
supported. This option expects the decompression binaries to be available in
your PATH.
```

So this is expected behavior. ripgrep does not advertise support for searching through tar archives.

BurntSushi:
That's because tar archives are explicitly skipped:

https://github.com/BurntSushi/ripgrep/blob/64317bda9f497d66bbeffa71ae6328601167a5bd/src/decompressor.rs#L92-L95

I admit, it does seem like `rg -z -a test test.tar.gz` should work.

PR #974, opened 2018-07-09
add zip file support

derekmorr:
This adds support for searching in zip files for the `-z` option. It partially addresses #918.

I'm not sure if we should use the `-p` or `-c` option to unzip. I'd appreciate feedback on that.


okdana:
`-p` definitely seems preferable to `-c`. It's probably also worth mentioning that there are at least two or three different implementations of the `unzip` tool — but AFAIK they all support `-p`, so that's probably fine

derekmorr:
Do I need to update the ignore crate and add the unzip package to travis? I assume unzip is pretty common, but it might not be a bad idea to explicitly list it.

BurntSushi:
@derekmorr Thanks for this PR! Unfortunately, I'm not sure it's a good fit. Namely, as I understand it, `zip` is an _archive format_, which means it may contain multiple files. As far as I can see, this PR provides an implementation that always searches zip archives as if they contained a single file. I'm not convinced that this is actually what we want, and in particular, I worry that it could be quite confusing to end users.

Interestingly, #918 is slightly different than this. Namely, the file extensions used (`tar` and `gz`) decouple the notion of archive from compression, such that decompression can be done without unpacking the archive and the archive can be searched directly. In this instance, there is no confusion presented to the end user because the archive isn't being unpacked; it's just searching a `.tar` like any other file.

I'm honestly not sure how to proceed here. My inclination is to draw line and say that ripgrep should not be responsible for unpacking archive formats.

> Do I need to update the ignore crate and add the unzip package to travis? 

@derekmorr I'm not sure what you mean with respect to the `ignore` crate, but we should make sure the test will run on Windows CI. Unfortunately, Cargo's output won't tell you anything here, and as far as it's concerned, the test ran and passed even though for all intents and purposes, it might have been skipped. It might be wise to remove the skipping logic and see what happens.

derekmorr:
I see what you mean. It is possible that a user would only have a single file in a zip archive, but it's also possible they wouldn't. 

FWIW, I think there are valid use cases for wanting to search inside the contents of a .tar.gz or a zip file. I've certainly seen plenty of questions about how to do that online. There is a tool that greps inside the full content of zip archives -- zipgrep -- but it sounds like you don't want to incorporate that functionality into ripgrep. That's a reasonable position to take.

My question about the ignore crate was that `DEFAULT_TYPES` doesn't have an entry for "zip" and does it need to have one?

derekmorr:
Ok, fair enough. I'll close this PR.

daniejstriata:
It would've been useful if ripgrep can search inside a zip and list the files and string matched in the zip but that would mean that ripgrep would need to extract the data from the zip first. a helper tool like `rga` does that using rg already.
https://github.com/phiresky/ripgrep-all

garoto:
Workarounds for zip searching has been posted on the issue tracker many times. This is a closed PR.

Issue #902, opened 2018-04-30
Can't run ripgrep on temp files: "No such file or directory"


egoldshtrom:
#### What version of ripgrep are you using?

```
$ rg --version
ripgrep 0.8.1 (rev c8e9f25b85)
-SIMD -AVX
```

#### What operating system are you using ripgrep on?

Ubuntu 16.04

#### Describe your question, feature request, or bug.

When trying to run ripgrep on a file in `/tmp` I get the error: `No such file or directory (os error 2)`

#### If this is a bug, what are the steps to reproduce the behavior?

1. Create a file in `/tmp` with some content. I put the following into `/tmp/test_file`:
```
INFO example text
ERROR something bad
INFO example text2
```
2. Try to search for `ERROR` using ripgrep: `rg ERROR /tmp/test_file`
3. You should get the output:
```
/tmp/test_file: No such file or directory (os error 2)
No files were searched, which means ripgrep probably applied a filter you didn't expect. Try running again with --debug.
```

#### If this is a bug, what is the actual behavior?

ripgrep works for me on other files:
```
$ rg move config
57:# move focused window
58:bindsym $mod+Shift+h move left
...
```

I have permissions to the `/tmp` file:
```
$ ls -la /tmp
drwxrwxrwt 15 root        root           139264 Apr 30 10:39 .
...
-rw-rw-r--  1 egoldshtrom egoldshtrom        57 Apr 30 10:32 test_file
...
```

I can grep the file or if I pipe the contents into ripgrep I can use ripgrep on it:
```
$ grep ERROR /tmp/test_file
ERROR something bad

$ cat /tmp/test_file | rg ERROR
ERROR something bad
```

But if I try to use ripgrep directly on the file it fails:
```
$ rg ERROR /tmp/test_file
/tmp/test_file: No such file or directory (os error 2)
No files were searched, which means ripgrep probably applied a filter you didn't expect. Try running again with --debug.

$ sudo rg ERROR /tmp/test_file
/tmp/test_file: No such file or directory (os error 2)
No files were searched, which means ripgrep probably applied a filter you didn't expect. Try running again with --debug.

$ rg ERROR /tmp/test_file --debug
DEBUG/grep::search/grep/src/search.rs:195: regex ast:
Literal {
    chars: [
        'E',
        'R',
        'R',
        'O',
        'R'
    ],
    casei: false
}
[...]

PR #979, opened 2018-07-13
Tweak snapcraft.yaml

mati865:
I know you don't have time to work with snaps, is there any way I could help you?


BurntSushi:
> I know you don't have time to work with snaps, is there any way I could help you?

Seems unlikely. My position has changed on this from "I don't have time" to "actively discourage the use of snap." I've never been able to figure out how to get it working, and moreover, the use of snap has generated so many bug reports that I can do nothing about that I've had to put a special call out about snap in the issue template and I removed it from the list of suggested installation instructions: https://github.com/BurntSushi/ripgrep/commit/6ffb4b7466c914a2683847b5fb4e30ea16749e94#diff-0b6c1416b1293274c4e2d11fb49e1795

Maybe these problems are fixable, but I just don't have the bandwidth to fix them. For example, I understand exactly nothing about the change that's being submitted here. Why is it necessary? What does it do? What does it fix? None of this is obvious to me and none of it is in the commit message.

mati865:
@BurntSushi there is extremely simple reason why it didn't work before as I wrote in https://github.com/BurntSushi/ripgrep/issues/902#issuecomment-404798881

> For example, I understand exactly nothing about the change that's being submitted here. Why is it necessary? What does it do? What does it fix? None of this is obvious to me and none of it is in the commit message.

Snap by default creates wrappers for running package which can be useful for other apps to configure their dynamic libs location but totally redundant for static libs. This change gives very small benefit to startup time.
I can add commit message.

I haven't used this option but if https://dashboard.snapcraft.io/snaps/ripgrep/collaboration/ does what it looks to do I could take care of managing releases.

I'll understand if you don't want to touch it anymore, in this case feel free to close this PR.

BurntSushi:
@mati865 If you can improve the commit message and ideally include a link to the relevant snap documentation (so that I can start learning how to diagnose these problems myself), then I'm happy to merge this.

mati865:
Snap documentation is a bit lacking, discussion about this feature is available [here](https://forum.snapcraft.io/t/telling-snapcraft-to-skip-generating-wrapper-scripts/1635).

> so that I can start learning how to diagnose these problems myself

Those issues were related to AppArmor.
As long as this snap is using `classic` confinement AppArmor will let it do everything and `ripgrep` will behave just like normal app.

This section doesn't apply for `classic` snaps:
To improve security snaps are confined and given only explicitly requested (just like Android or iOS apps). Problematic snap `rg`requested only `home` interface which gives access only to user files (hence the permission errors). List of available interfaces is [here](https://docs.ubuntu.com/core/en/reference/interfaces/).
AppArmor debugging guide is available [here](https://docs.ubuntu.com/core/en/guides/intro/security)
To make `ripgrep` confined it would need interface giving read access to everything user can read. The problem is interface like that doesn't exists ([request was filed](https://bugs.launchpad.net/snappy/+bug/1603838)). Of course one [can create interface](http://www.zygoon.pl/2016/08/creating-your-first-snappy-interface.html) like that but it wouldn't differ from `classic` confinement too much and Ubuntu developer made valid point `the request is at odds with snappy's application isolation goals`.

Issue #978, opened 2018-07-12
Generic decoder program rather than/in addition to extension-based decompression


c-blake:
#### What version of ripgrep are you using?

ripgrep 0.8.1
-SIMD -AVX

#### How did you install ripgrep?

emerge sys-apps/ripgrep

#### What operating system are you using ripgrep on?

Gentoo linux (kernel 4.17.5); Rust version 1.27.0, but none of these questions should be relevant to this feature suggestion/request.

#### Describe your question, feature request, or bug.

In a recursive searching problem setting, one often has many types of (usually binary) files one would like to search through the "decoded" text of.  Here decoding is something that may even be idiosyncratic to a specific user on a specific system. For example, these may not only be compressed files - other encodings are possible/common such as PDF files where a text layer is extractable via a program like ``pdftotext`` in the poppler package or ``mutool draw`` or some similar tool.

``ripgrep`` already does a better job than most greps here, but it could be much better.  Besides an ever growing/ultimately unknowable set of desired encodings/translations, file name extensions are not the most reliable way to detect file types.  The ``file`` program or its ``libmagic`` are more complete, but is quite slow and even it may omit classes of file a user wants to use.  Further, the transformation that counts as decoding could be idiosyncratic.  I give a concrete example with unidiffs below.

In light of both of the above considerations, the most flexible/robust approach is akin to what ``less`` does with its ``LESSOPEN`` environment variable - allow a user specified program to both classify and decode files that ``ripgrep`` cannot.  The dowside of this relative to the current ``ripgrep`` filename extension oriented approach is a "double or triple ``exec``" cost, but there are many upsides -- total generality, no shared library dependencies and so on.  { Many might not even pay attention to these extra fork/exec costs.  For example, many programs use ``system(3)`` rather than fork and exec which already incurs an extra ``exec``.  These days ``gunzip`` is even a shell script wrapper causing an extra exec of /bin/sh.  But in a recursive grep context the costs surely add up. }

To be concrete consider something like this script which a hypothetical new ``GREP_OPEN`` or ``RIPGREP_OPEN`` variable could point to:
```sh
#!/bin/sh
case "$RIPGREP_INPUT" in
  *.gz) exec gzip -dc ;;                      # input name not needed for gzip..
  *.pdf) exec pdftotext "$RIPGREP_INPUT" - ;; #..but MIGHT be. So, ripgrep exports it
  *) exec cat ;; # unencoded
esac
```
Another possibility using ``file`` is:
```sh
[...]

PR #981, opened 2018-07-13
Try to fullfill all the goals of the generic decoder program feature

c-blake:
request: https://github.com/BurntSushi/ripgrep/issues/978

There are a a bunch of choices maybe I should mention in this PR.
I called it "-P,--preprocessor" to suggest its primary function.
I basically just imitated the way decompression was handled with
a new file ``src/preprocessor.rs`` instead of ``src/decompressor.rs``.

Currently, if both ``--search-zip`` and ``--preprocessor PROGRAM`` are
given,  the latter is used.  This seemed reasonable since preprocessing is more
general and probably involves more sophisticated users who can more easily
include whatever compression programs they want (or don't want) using whatever
dispatching algorithm they want.

For me, it compiles and runs fine on rust-1.27.1 both with no ``-P`` at all,
with ``-P program-using-only-stdin`` and ``-P program-using-argv1``.  The only
failing I see right now on Linux is that if you specify a bogus preprocessor
like ``rg -P /junk foo`` it does not error out at the very first file.

Oh, and while the shell script snippet formats fine in the --help output, the
auto-generated man page corrupts it into a 2- or 3-liner with some chars
dropped out.  Not sure what to do about that.  Could also just put that
material in GUIDE.md and drop it from the help.

Also, this is my very first significant stab at Rust work.  So, I may well
have done some things in an undesirable way.

### DIFF
--- a/complete/_rg
+++ b/complete/_rg
@@ -173,6 +173,7 @@ _rg() {
     + '(zip)' # Compressed-file options
     {-z,--search-zip}'[search in compressed files]'
     $no"--no-search-zip[don't search in compressed files]"
+    {-P,--preprocessor}'[search files needing preprocessing]'
 
     + misc # Other options — no need to separate these at the moment
     '(-b --byte-offset)'{-b,--byte-offset}'[show 0-based byte offset for each matching line]'
--- a/src/app.rs
+++ b/src/app.rs
@@ -540,6 +540,7 @@ pub fn all_args_and_flags() -> Vec<RGArg> {
     flag_regexp(&mut args);
     flag_replace(&mut args);
     flag_search_zip(&mut args);
+    flag_preprocessor(&mut args);
     flag_smart_case(&mut args);
     flag_sort_files(&mut args);
     flag_stats(&mut args);
@@ -1459,6 +1460,33 @@ This flag can be disabled with --no-search-zip.
     let arg = RGArg::switch("no-search-zip")
         .hidden()
         .overrides("search-zip");
+    args.push(arg);
+}
+
+fn flag_preprocessor(args: &mut Vec<RGArg>) {
+    const SHORT: &str = "search outputs of \"COMMAND FILE\" for each FILE";
+    const LONG: &str = long!("\
+For each input FILE, search the standard output of \"COMMAND FILE\" rather
+than the contents of FILE.  This option expects the COMMAND program to be
+available in your PATH.  An empty string COMMAND deactivats this feature.
+
+When searching over sets of files that may require one of several decoders
+as preprocessors, COMMAND should be a wrapper program or script which first
+classifies FILE based on magic numbers/content or based on the FILE name and
+then dispatches to an appropriate preprocessor.  Each COMMAND also has its
+standard input connected to FILE for convenience.
+
+For example, a Unix script for COMMAND might look like:
+case \"$1\" in
+  *.pdf) exec pdftotext \"$1\" - ;;
+  *) case $(file \"$1\") in
+       *Zstandard*) exec pzstd -cdqp8 ;;
+       *) exec cat ;;
+	 esac;
+esac
+");
+    let arg = RGArg::flag("preprocessor", "COMMAND").short("P")
+        .help(SHORT).long_help(LONG);
     args.push(arg);
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,8 @@ Feature enhancements:
   Add `--count-matches` flag, which is like `--count`, but for each match.
 * [FEATURE #880](https://github.com/BurntSushi/ripgrep/issues/880):
   Add a `--no-column` flag, which disables column numbers in the output.
+* [FEATURE #978](https://github.com/BurntSushi/ripgrep/issues/978):
+  Add a `--preprocessor` option to filter inputs with an arbitrary program.
 
 Bug fixes:
 
--- a/README.md
+++ b/README.md
@@ -107,6 +107,8 @@ increases the times to `2.640s` for ripgrep and `10.277s` for GNU grep.
   specifically specified with the `-E/--encoding` flag.)
 * ripgrep supports searching files compressed in a common format (gzip, xz,
   lzma or bzip2 current) with the `-z/--search-zip` flag.
+* ripgrep supports arbitrary input preprocessing filters which could be PDF
+  text extraction, less supported decompressing, decrypting, and so on.
 
 In other words, use ripgrep if you like speed, filtering by default, fewer
 bugs, and Unicode support.
--- a/src/args.rs
+++ b/src/args.rs
@@ -80,6 +80,7 @@ pub struct Args {
     types: Types,
     with_filename: bool,
     search_zip_files: bool,
+    preprocessor: PathBuf,
     stats: bool
 }
 
@@ -288,6 +289,7 @@ impl Args {
             .quiet(self.quiet)
             .text(self.text)
             .search_zip_files(self.search_zip_files)
+            .preprocessor(self.preprocessor.clone())
             .build()
     }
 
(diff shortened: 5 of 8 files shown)


src/app.rs
```diff
@@ -1462,6 +1463,33 @@ This flag can be disabled with --no-search-zip.
     args.push(arg);
 }
 
+fn flag_preprocessor(args: &mut Vec<RGArg>) {
+    const SHORT: &str = "search outputs of \"COMMAND FILE\" for each FILE";
```
okdana: Quoting `COMMAND FILE` doesn't seem right; place-holders aren't quoted anywhere else in the help text. Also you switch between `COMMAND FILE` and `COMMAND`; `COMMAND` seems simpler (or maybe `PROGRAM` if you're worried that's too vague). Also `deactivates` is misspelt

src/app.rs
```diff
@@ -1462,6 +1463,33 @@ This flag can be disabled with --no-search-zip.
...
+For example, a Unix script for COMMAND might look like:
+case \"$1\" in
+  *.pdf) exec pdftotext \"$1\" - ;;
+  *) case $(file \"$1\") in
+       *Zstandard*) exec pzstd -cdqp8 ;;
+       *) exec cat ;;
+	 esac;
+esac
+");
+    let arg = RGArg::flag("preprocessor", "COMMAND").short("P")
```
okdana: Usually when one option overrides another (as you said this does with `-z`), you put that in the help text and then specify it here. You can see examples of that elsewhere in the file

I actually would have suggested making them *conflict*, because it's not intuitively obvious how they should behave together, but then i thought that a lot of people probably add `-z` to their aliases/configs, so that might be too inconvenient. Not sure

complete/_rg
```diff
@@ -173,6 +173,7 @@ _rg() {
     + '(zip)' # Compressed-file options
     {-z,--search-zip}'[search in compressed files]'
     $no"--no-search-zip[don't search in compressed files]"
+    {-P,--preprocessor}'[search files needing preprocessing]'
```
okdana: If you're going to put these together, i would suggest renaming the group and changing the comment, since it's no longer accurate. Maybe `(preprocessor-zip) # File-preprocessing options` or something, to match the way `pretty-vimgrep` is done

Also, the spec isn't right. It should be something like:

```
{-P+,--preprocessor=}'[specify preprocessor utility]:preprocessor utility:_command_names -e'
```

(`_command_names -e` will prefer `PATH` executables, but also do the right thing if you try to complete an actual file path)

src/app.rs
c-blake: Ok.  Dequoted & fixed spelling.  I use COMMAND FILE when there is also a FILE in the sentence context and COMMAND when it's just about the program.

okdana: Ah, i think i see what you mean

phiresky:
Just as an example of why this would be awesome: Together with [this caching pdftotext wrapper](https://gist.github.com/phiresky/5025490526ba70663ab3b8af6c40a8db) as a preprocessor this is able improve on pdfgrep by orders of magnitude:

On a semi-large directory of pdfs:


`time pdfgrep -r IP >/dev/null`

`3,20s user 0,08s system 99% cpu `**3,284 total**

`time rg --preprocessor pdftotext-cached.sh IP >/dev/null`

`0,12s user 0,04s system 374% cpu `**0,035 total**


That's almost a 9000% performance improvement.

(Even without caching it only takes 0.5s, not sure what the pdfgrep people are doing)

c-blake:
Yeah...Besides lecture slides I have a slew of papers I've collected over the years, as I expect many others have.  I use this all the time with a custom GNU grep patch I did that does basically the same thing.  Line numbers might be nice and all, but honestly I mostly use this with a specific enough pattern and ``--files-with-matches``, myself.  My use case is usually..."What was that paper or two that did xyz again?".

As mentioned in the feature request, applications are really bounded only by your imagination.  The searching through only the parts of context diffs with actual changes, conceivably even (trustworthy) foreign language translations if there was a good library/CLI tool for that, ``rg -P enFrancais`` or whatever.

Caching transformers surely can speed things up at the expense of disk space to maintain the cache.  Personally, my archives are small enough that I just re-decode on the fly.  The parallel operation of ``rg`` really helps with transformation time, actually.  Indeed, if one is willing to spend the space for a full cache, maintaining it with ``make`` or whatever, it will always be faster to run ``rg`` on that shadow file hierarchy and transform the filenames in the output (if you even need to).

 To get even as low as 70 us/file + 0.34 sec/GB on my box at home, I have to have a statically linked classifier that does its own pass-through for "uncoded data" _and_ trim my environment to just $PATH.  That may not sound like much overhead, but the microseconds and milliseconds of overhead pile up when you have 10s of thousands of files on fast NVMe storage or buffered in RAM and _most but not all_ of that is uncoded data.  The per-byte costs come from copies over the pipe buffer.  My email history is sort of like that.  Anyway, the real user time of those overheads is usually smaller/around the same as my time to enter a pattern and interpret the results.  A less careful management of the overheads can easily blow that up by 10X or more (some fancy dynlinked bash script dispatcher type stuff to dynlinked cat with a big environment, etc.) which pushes it into annoying territory (at least for me).

complete/_rg
```diff
@@ -170,7 +170,8 @@ _rg() {
     {-w,--word-regexp}'[only show matches surrounded by word boundaries]'
     {-x,--line-regexp}'[only show matches surrounded by line boundaries]'
 
-    + '(zip)' # Compressed-file options
+    + '(input-decoding)' # Compressed-file options
```
okdana: Missed the comment here

c-blake: Oops.  Did you want that ``-E,--encoding`` option down in that input-decoding group, too?  Instead of "misc/other"?

okdana: No, these options were grouped together because they're completed exclusively of each other (that's what the `(...)` in the group name means). `-E` is independent and doesn't have any particular relationship with other options, so it can be dumped in with the miscellaneous stuff. I can see how the group name `input-decoding` might kind of imply that relationship, which is why i'd suggested `preprocessor-zip` or something more specific like that; doesn't matter too much tho

c-blake: Ah.  Ok.  I see.  Well, if someone optimizes some other common case and adds another option like --search-pdf (just as an example) we have a general group name.  Doesn't matter to me unless someone complains.

BurntSushi:
@c-blake Do you want to think of a different short option for this other than `-P`? I have something else in mind for use with `-P` so I'd rather not use it for this. I might even prefer not to allocate a short option for this at all, since it seems like a less common feature and short flag real estate is precious. Perhaps we could shorten the flag down to just `--pre`?

BurntSushi:
Some notes from working through the code and cleaning it up:

* `PreprocessorReader` does not attach the given file to stdin, which meant that preprocessor scripts couldn't actually read from stdin.
* The preprocessor isn't active when ripgrep is searching stdin. This seems OKish to me, but I've added it to the documentation.
* The preprocessor itself probably shouldn't be emitting log messages. Instead, it should return an error and the caller should decide how and when to print it. (In this case, that's the worker.) I see that this style was probably emulated from the decompressor implementation, but the decompressor should probably be updated in this vein as well. (Although, IIRC, the decompression stuff has slightly different constraints for when to emit error messages.)
* I've fixed up the ill-formatted script example for the man page.
* I've removed the `-P` short option and changed the long option to `--pre`.
* I've changed the representation of the preprocessor from `PathBuf` to `Option<PathBuf>` to better communicate that it is optional.
* I briefly investigated producing a better failure mode when the preprocessor command simply doesn't exist, but it looks like Rust's standard library doesn't expose its path resolution logic for commands, and I didn't want to try and re-litigate that. We could actually attempt to start the process, but then that would need to get added to the command interface, which is too deliciously simple to disturb IMO.

I've opened #989 with these changes.

PR #1007, opened 2018-08-07
add Idris, Dhall, and ATS

vmchale:
This adds support for Idris, Dhall, and ATS to the `ignore` crate.

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -149,7 +149,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("groovy", &["*.groovy", "*.gradle"]),
     ("h", &["*.h", "*.hpp"]),
     ("hbs", &["*.hbs"]),
-    ("haskell", &["*.hs", "*.lhs"]),
+    ("haskell", &["*.hs", "*.lhs", "*.cpphs", "*.c2hs", "*.hsc"]),
     ("hs", &["*.hs", "*.lhs"]),
     ("html", &["*.htm", "*.html", "*.ejs"]),
     ("idris", &["*.idr", "*.lidr"]),
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -98,6 +98,7 @@ use {Error, Match};
 
 const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("agda", &["*.agda", "*.lagda"]),
+    ("ats", &["*.ats", "*.dats", "*.sats", "*.hats"]),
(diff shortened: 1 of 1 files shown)


ignore/src/types.rs
```diff
@@ -107,7 +108,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("bazel", &["*.bzl", "WORKSPACE", "BUILD"]),
     ("bitbake", &["*.bb", "*.bbappend", "*.bbclass", "*.conf", "*.inc"]),
     ("bzip2", &["*.bz2"]),
-    ("c", &["*.c", "*.h", "*.H"]),
+    ("c", &["*.c", "*.h", "*.H", "*.cats"]),
```
BurntSushi: I don't think I've seen `cats` used for a C file before. Where does that come from?

vmchale: It is used with ATS to contain C code called by ATS code later, e.g. https://groups.google.com/forum/#!topic/ats-lang-users/On6L-1e6p_0

ignore/src/types.rs
```diff
@@ -149,7 +149,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("groovy", &["*.groovy", "*.gradle"]),
     ("h", &["*.h", "*.hpp"]),
     ("hbs", &["*.hbs"]),
-    ("haskell", &["*.hs", "*.lhs"]),
+    ("haskell", &["*.hs", "*.lhs", "*.cpphs", "*.c2hs", "*.hsc"]),
```
BurntSushi: Same question as for `cats` I think. When and where are these file extensions used?

ignore/src/types.rs
BurntSushi: Great. Thank you.

ignore/src/types.rs
vmchale: `.hsc` is for the `hsc2hs` preprocessor. You can see an example here: https://github.com/haskell-foundation/foundation/blob/master/foundation/Foundation/Foreign/MemoryMap/Posix.hsc

`.cpphs` is just for the `cpphs` preprocessor, e.g. here: https://github.com/vmchale/atspkg/blob/master/ats-pkg/internal/Quaalude.cpphs

`c2hs` is another pre-processor, and there is a nice example here: https://github.com/haskell/c2hs/blob/master/examples/libghttpHS/Ghttp.chs

BurntSushi: All righty, I buy it, thanks for showing me those!

vmchale: No problem!

PR #1010, opened 2018-08-10
Add *.sbt to scala type

woky:
Sbt is currently most used Scala build tool which uses *.sbt files, which are basically Scala.


BurntSushi:
Thanks1

PR #1017, opened 2018-08-19
libripgrep: PCRE2 support, multiline search, JSON output and more

### LINKED ISSUE #2: add context handling to memory map searching
The memory map searcher supports every option except for printing contexts. Specifically, if any of `-A`, `-B` or `-C` are provided, then memory map searching can't be used.

Incidentally, handling contexts in the memory map searcher should be much easier than in the streaming searcher.

### LINKED ISSUE #162: libripgrep

### LINKED ISSUE #176: support searching across multiple lines
Say for example I'm trying to find instances of `click` that reside in a `listeners` block, like so:

```
listeners: {
    foo: ...
    click: ....
}
```

According to the Rust regex docs, I should be able to do: `rg '(?s)listeners.+click'`, but this doesn't seem to work. Does ripgrep not support multiline regex?

### LINKED ISSUE #188: support fancier regex features such as look-around and backreferences
One area that makes `ag` unable to be replaced by `ripgrep` is that `ag` supports negative lookahead. For example:

```
ag 'foo(?!.*bar)'
```

This would find all rows with "foo", but not with "bar" somewhere after that on the same line.

`ripgrep` gives the following error message:

```
› rg 'foo(?!.*bar)'
Error parsing regex near 'foo(?!.*ba' at character offset 5: Unrecognized flag: '!'. (Allowed flags: i, m, s, U, u, x.)
```

I guess this is actually a problem in the `regex` crate, and I can open the same issue there if you wish.

### LINKED ISSUE #200: ripgrep doesn't stop when its pipe is closed

### LINKED ISSUE #244: Output format with parseable match indices?
For integration in other tools (editors...) it would be great to have a parseable output format that includes the start and end index for each match in a line. You can do it with colors (but see #242) but it's not really what colors are meant for.

ag has --ackmate, which outputs something like this:
```
$ ag --ackmate host
:src/util.rs
85;43 4,55 4:                                       ext_host: "localhost".into(),
```

### LINKED ISSUE #389: Different -w behavior from grep/git-grep

### LINKED ISSUE #416: Carriage return handling on Windows

### LINKED ISSUE #441: ^$ matches a phantom line at end of file

### LINKED ISSUE #643: improve stdin detection on Windows
Using version 0.6.0

I tried to parse some text and for each parsed term to call rg. Something like this (added newlines fo readibility):

```
rg -F ".EmailHtml.HandleAttribute" -g 2017-10-18* -B 3 |
rg ".*? machine (\d+) Back.*" -r "$1" |
powershell -command "$input | % { rg -g 2017-10-18* \"machine $_ Action \d+\" }"
```

However, this doesn't work. I'm able to get much more simple example like this:

```echo a|powershell -command "d:\prgs\rg.exe -g 2017-10-18* test "``` doesn't output anything
```powershell -command "d:\prgs\rg.exe -g 2017-10-18* test "``` - searches for "test" as expected (not that only the ```echo a|``` is missing there)

Does anybody here any idea why rg doesn't work when called from PowerShell when something is sent through pipeline into PowerShell?

### LINKED ISSUE #690: ^$ does not work when using --no-mmap

### LINKED ISSUE #764: Size of output is 5 times bigger with ripgrep than with thesilversearcher for color=always
This is very similar to this issue : https://github.com/BurntSushi/ripgrep/issues/696.

But I think I found out what the underlying problem is.

rg --color always --colors line:fg:yellow --colors line:style:bold --colors path:fg:green --colors path:style:bold --colors match:fg:black --colors match:bg:yellow --colors match:style:nobold '.' --no-heading --line-number >/tmp/foo_rg 
ag '.' --color --nogroup --column  > /tmp/foo_ag 

When I then run : 

```
cat /tmp/foo_rg | less -R
# and 
cat /tmp/foo_ag | less -R
```

I see that the output colors are "similar", but the size of the files are very different : 

```
ls -lah /tmp/foo*
-rw-r--r-- 1 edgar edgar  22M Jan 29 10:48 /tmp/foo_ag
-rw-r--r-- 1 edgar edgar 109M Jan 29 10:48 /tmp/foo_rg
```

And the number of lines in each file is the same.

I have taken a screenshot of the same lines of each efile for api/check.py: 
On top is the ripgrep output, On the bottom the ag output : 

selection_001

As you can see, ripgrep has a huge amount of escape sequences that ag doesn't have.

### LINKED ISSUE #917: trim whitespace from lines

```
ripgrep 0.8.1 (rev c8e9f25b85)
-SIMD -AVX 
```

From `.deb` binary.

Linux Mint 18.1 Serena.

First of all, thanks and congratulations for this tool.

I've searched in the help and in the issues, and as I didn't find anything related, I'm opening this suggestion.

I think that would be nice to add an option for trimming the output of the coincident lines. In other words, not to print the indentation of the lines. I think that this could improve the readability of the output.

Currently, for this search:

```
rg POST views/
```

I would get the following output:

```
views/modals/contact.twig
13:               <form action="/contact/send" method="POST" id="form-contact">
```
(Note the code indentation).

With an option for trimming (e.g. `--trim`), we would get the following:

```
views/modals/contact.twig
13:<form action="/contact/send" method="POST" id="form-contact">
```
(The indentation is now removed from the output).

I really think that this could be a very useful option, since almost every part of the code is indented, except class or function definitions.

**PD:** this could also work in combination with `--max-columns` option, so the trimming would be done before the output is filtered with this options. But as I know nothing about ripgrep internals and the feasibility and difficulty of this, I just leave it as a little extra suggestion. Actually this wouldn't be very relevant, since the `--max-columns` is thought for the output we actually don't want to see. So, if you find appropriate the original suggestion, just implement it as you think it's better for ripgrep. :) 

Thanks again and keep up the great work!

### LINKED ISSUE #922: "file length overflows usize"
**#### What version of ripgrep are you using?**

`ripgrep 0.7.1
-AVX -SIMD`.

**#### What operating system are you using ripgrep on?**

Operating System: 64-bit
Windows 7: Cygwin Terminal

**#### Describe your question, feature request, or bug.**

With specific files when trying to grep I get an error saying "file length overflows usize"

**#### If this is a bug, what are the steps to reproduce the behavior?**

$ ./rg -i -w "Gurkiran" "zoosk.txt"
file length overflows usize

### LINKED ISSUE #937: empty pattern doesn't display utf8

ripgrep 0.8.1
-SIMD -AVX

`cargo install ripgrep -f`

Linux Debian 9.

On empty patterns utf8 strings aren't presented with the correct encode.

```bash
❯ rg -S -n --color=always '' Cargo.toml
1:[package]
2:authors = ["Jo[0mo Bolila <j**>"]
..
❯ rg -S -n --color=always 'au' Cargo.toml
2:authors = ["João Bolila <j**>"]
```

Or with any utf8 content.

https://github.com/lotabout/skim/issues/83

Doesn't display correctly on empty patterns

Display always like when has a pattern match.

Thanks!

### LINKED ISSUE #940: Exit status should be independent of --passthrough

```
$ rg --version
ripgrep 0.8.1
-SIMD -AVX
```

Homebrew

macOS 10.13.4

The command
```
rg --passthrough searchterm file.txt
```
should exit with the same code as when the command is run without the flag.

Both `ack` and `ag` work this way.

```
$ echo hi > file
$ rg asdf file
$ echo $?
1
$ rg --passthrough asdf file
1:hi
$ echo $?
0
```

`ripgrep --passthrough` always exits with `0` (unless the file is empty).

```
$ rg --passthrough --debug asdf file
DEBUG/grep::search/grep/src/search.rs:195: regex ast:
Alternate(
    [
        Literal {
            chars: [
                'a',
                's',
                'd',
                'f'
            ],
            casei: false
        },
        StartText
    ]
)
DEBUG/globset/globset/src/lib.rs:401: built glob set; 0 literals, 0 basenames, 0 extensions, 0 prefixes, 0 suffixes, 1 required extensions, 0 regexes
1:hi
```

To exit with the same status as without the flag.

### LINKED ISSUE #980: ^$ erroneously matches around multiples of 4096

ripgrep 0.8.1
-SIMD -AVX

cargo install
rustc 1.27.0 (3eda71b00 2018-06-19)

`Linux 3.13.0-43-generic #72-Ubuntu SMP Mon Dec 8 19:35:06 UTC 2014 x86_64 x86_64 x86_64 GNU/Linux`
`Linux 4.14.34-v7+ #1110 SMP Mon Apr 16 15:18:51 BST 2018 armv7l GNU/Linux`
`Darwin 17.7.0 Darwin Kernel Version 17.7.0: Thu Jun 21 22:53:14 PDT 2018; root:xnu-4570.71.2~1/RELEASE_X86_64 x86_64`

```
~$ yes | rg -n ^$ | head 
2:
4098:
8195:
12292:
16389:
20486:
24583:
28680:
32777:
36874:
```

`yes | rg ^$` should not find any matches.

This also seems to break line numbers on any pattern that could possibly match a blank line.

(I was searching for `^.{0,12}$` in a file that contains no empty lines and was surprised to find empty lines. My workaround was to change the pattern to `^{1,12}$`, but I was looking at incorrect line numbers until then.)

```
~$ yes | rg -C1 -n ^$ | head 
1-y
2:
3-y
--
4097-y
4098:
4099-y
--
8194-y
8195:
```

This looks similar to #441, but that's only at the end of file.

### LINKED ISSUE #993: add -z/--null-data flag for reading large binary files

Version 0.8.1 [as built by NixOS](https://nixos.org/nixos/packages.html#ripgrep)

Hmm, so I think I just hit an issue similar (but not identical) to https://github.com/BurntSushi/ripgrep/issues/458, when running `rg [a string not on the disk] /dev/nvme0n1p3`. (this being a ~100G partition, on an 8G-RAM machine). The difference being that, here, rg gets oom-killed in a matter around 1min.

Dmesg output of the OOM killer http://xelpaste.net/1P14FN.

Run `rg [some search pattern] [some big partition]`. That said I can't reproduce on the computer on which I'm typing this, which has `sda*` partitions, and for which search finishes in a matter of seconds, so there may be something special about `nvme0n1p*`?

Ripgrep gets OOM-killed

Either ripgrep could have managed to actually perform the search (the pattern was something very simple, no regex or anything, so afaiu theoretically the search could be done with very limited memory use in 3 passes for the output, forwards until the pattern, then backwards until a newline, then forwards again while outputting until the next newline), which would likely be best, but quite potentially require large changes to the codebase.

Or ripgrep could display an error message saying “Sorry we couldn't allocate enough memory because {your file is too big, your pattern is too complex, …}”, so that the reason why ripgrep is OOM'ing is clear. And if not possible, even a “OOM” message would still be better than the SIGKILL rg currently receives through the OOM.

Anyway, these are just nitpicks, thank you for ripgrep!

### LINKED ISSUE #997: passthru replace

0.81

releases deb

Ubuntu 16.04

It would be really great if the replace flag worked with the passthru flag - that is, it will replace everything that matches the regex, but will still output non-matching lines.

I know that you can use sed for this, but ripgrep is much more convenient and user friendly, and also highlighting the replacements is great.

Use cases:
1. Multiple replaces - when you want to replace multiple things, you can't chain ripgreps since strings that aren't matched by the first replace won't be inputted into the second, so it won't replace them
2. Replacing paths - sometimes I read other people logs, and for the sake of convenience I want to replace their file paths with mine - no problem, just replace the prefix, but of course I can't do it currently in ripgrep since all other logs will be omitted.
3. This stack overflow question I stumbled upon could have bin solved with this feature: https://stackoverflow.com/questions/51504318/how-to-replace-a-with-o-with-0-in-a-word-list-and-remove-the-unchanged-wor

### LINKED ISSUE #1013: rg --version is misleading
When running `rg --version` with the provided ripgrep-0.9.0-x86_64-unknown-linux-musl binary I'm getting the following output:

```
ripgrep 0.9.0 (rev 6799dcfc0e)
-SIMD -AVX
```

This suggests that `SIMD` and `AVX` are disabled. However, the binary is supposedly using runtime detection of the corresponding hardware support, so this is misleading.

To fix this, the corresponding output code should check for hardware support like the ripgrep code actually using the hardware instructions. Currently, the output code only considers the corresponding Cargo features, ignoring runtime detection.

BurntSushi:
After a lot of work, libripgrep is finally read to be merged to master. This includes planned features such as multiline search and JSON output, but also a surprise: ripgrep now provides the ability to opt into using PCRE2 as the regex engine instead of Rust's default regex engine. This provides a way to use look-around and backreferences with ripgrep. This support should work on Windows, macOS and Linux.

libripgrep still needs high level documentation in the form of a cookbook and a guide, so it is not quite ready for wide use yet. However, the API itself is minimally documented, so ambitious individuals should feel empowered to try it out. :-)

Closes #162 (libripgrep)
Closes #176 (multiline search)
Closes #188 (opt-in PCRE2 support)
Closes #244 (JSON output)
Closes #416 (Windows CRLF support)
Closes #917 (trim prefix whitespace)
Closes #993 (add --null-data flag)
Closes #997 (--passthru works with --replace)

Fixes #2 (memory maps and context handling work)
Fixes #200 (ripgrep stops when pipe is closed)
Fixes #389 (more intuitive `-w/--word-regexp`)
Fixes #643 (detection of stdin on Windows is better)
Fixes #441, Fixes #690, Fixes #980 (empty matching lines are weird)
Fixes #764 (coalesce color escapes)
Fixes #922 (memory maps failing is no big deal)
Fixes #937 (color escapes no longer used for empty matches)
Fixes #940 (--passthru does not impact exit status)
Fixes #1013 (show runtime CPU features in --version output)

cc @roblourens


okdana:
This is huge and amazing, thank you!

I was wondering whether a `--no-multiline-dotall` would be appropriate? Obviously people can just flip it off with `(?-s)`, but they can do that with `(?s)` too, so maybe for UX consistency/discoverability it'd makes sense...?

I also had some feedback on the zsh completion — mostly just nit-picking, but a few more functional things (like exclusivity errors). Rather than flood you with a bunch of comments, i think i'll prepare a patch that can be applied against your branch.

One particular issue i do see with the completion is that, since `--pcre2-unicode` is enabled by default, `--no-pcre2-unicode` doesn't really qualify for `$no` — it should actually be the other way around. The problem with that (maybe one you ran into?) is that `--no-pcre2-unicode` doesn't have its own entry in the usage help, so the CI script won't deal with it properly.

We could work around that by giving `--no-pcre2-unicode` the 'canonical' spot in the usage help (and that might be more consistent with the way some of the other options are documented, tbh), but i think it's probably just time to make that script also extract long options from the option-description bodies. I'll do that.

Let me know how you feel about `--no-multiline-dotall` and i'll incorporate that into my changes.

I only looked at a few other things but so far everything else works great for me.

roblourens:
This is awesome. I know a lot of people will be very happy about being able to search with PCRE2 in vscode!

BurntSushi:
Thanks for the comments @okdana!

> I was wondering whether a `--no-multiline-dotall` would be appropriate?

So my thinking here is that `--multiline-dotall` probably isn't a flag folks are going to type often. If someone wants that behavior "on demand," then I think they're more likely to just add the inline `s` flag. Instead, I expect that flag would be added in an alias or a config, and when one wants to disable it, I'd expect folks to use the inline `s` flag for that too, instead of typing out `--no-multiline-dotall`.

But, it doesn't cost much to add the flag and maybe it's worth it for consistency reasons. We don't yet do it for every flag though I don't think, but it's probably pretty close.

> i think i'll prepare a patch that can be applied against your branch.

That's great, thanks! Would it work for you if I merged this PR and then you just submit the patch against master? I'm not planning on doing a release imminently or anything.

Also, in general, I've found it pretty difficult to update the completion script. I knew I probably got some stuff wrong and I'm grateful you've caught it. :-) I can't tell whether a little bit of documentation describing the syntax would fix my problem editing the completion rules, or if it would be better to try to generate the completion script like we generate the man page (potentially by adding addition metadata on `RGArg`).

> One particular issue i do see with the completion is that, since --pcre2-unicode is enabled by default, --no-pcre2-unicode doesn't really qualify for $no — it should actually be the other way around. The problem with that (maybe one you ran into?) is that --no-pcre2-unicode doesn't have its own entry in the usage help, so the CI script won't deal with it properly.

> We could work around that by giving --no-pcre2-unicode the 'canonical' spot in the usage help (and that might be more consistent with the way some of the other options are documented, tbh), but i think it's probably just time to make that script also extract long options from the option-description bodies. I'll do that.

Yeah, I'm not sure what I want to do here. You're right that making `--no-pcre2-unicode` be the "canonical" flag would fix this issue and would also be more consistent with how we do other things. My only rationale for doing it the way I did was because this way, `--pcre2-unicode` will show up right next to `--pcre2` when sorting flags alphabetically, which I think is nice. But maybe the fix there is to group options logically instead of alphabetically. Gah.

> One thing that a few tools do is provide a machine-readable method of listing all of their options. Symfony Console applications do this, and clang even has a fancy --autocomplete which does the work for their bash completion function. I wonder if Clap can be made to do something similar? (If it can, is it worth it? Not sure)

ripgrep does kind of have at least the _guts_ to make this happen now I think. This was in large part motivated by my desire to generate a quality man page directly from the flag definitions. This is why `RGArg` exists, which is basically a light layer right above clap that ripgrep can read at build time and do stuff with. If we could generate a completion script from that, I think that would be the best of all worlds. Even if `RGArg` as it exists today isn't enough, I think it is more than OK to add additional metadata just for generating completions such that flag definitions might need to be modified to add it. This also increases the locality trend, such that if you add a new flag to ripgrep, there are fewer places you need to edit to add it. (Today, there are two places: the definition itself and the zsh completion script.)

> PS: Treating passed-through lines as context is so much smarter than the old method (and totally obvious in hind sight), i like that a lot

Hah yeah. It also fixes the exit status bug nicely. It would not have been feasible to do in the old ripgrep code though!

@okdana Your patch LGTM, thank you. :-) I'm fine with taking in `--no-multiline-dotall`.

jessegrosjean:
Nice work and big thanks from me too! I've been using libripgrep in my own app, results are much better then I what I would have come up with by myself. It's also introduced me to the world of rust and having full control over memory and everything else. It's been and adventure, mostly good :).

okdana:
>That's great, thanks! Would it work for you if I merged this PR and then you just submit the patch against master?

Yeah, i can do that.

>Also, in general, I've found it pretty difficult to update the completion script. I knew I probably got some stuff wrong and I'm grateful you've caught it. :-) I can't tell whether a little bit of documentation describing the syntax would fix my problem editing the completion rules, or if it would be better to try to generate the completion script like we generate the man page (potentially by adding addition metadata on RGArg).

I worry that, unless it was extremely comprehensive, any automatic generation done by ripgrep would have similar limitations to the ones Clap had. Obviously it can be done, but i don't know, i have mixed feelings i guess.

Short of that, maybe it would help if i put together a little Markdown guide to explain in simple terms how the option specifications work, and what the conventions are? I think i could make something that's easier to follow than the official documentation, at least for the sub-set of features that most of our specs use. Then we could put that in the repo and link to it at the top of the function.

BurntSushi:
@okdana A small guide would be great! I think it would be nice to just inline it into a comment in the completion script though, just so that it's all in one place.

> I worry that, unless it was extremely comprehensive, any automatic generation done by ripgrep would have similar limitations to the ones Clap had. Obviously it can be done, but i don't know, i have mixed feelings i guess.

I'm not too familiar with the limitations that Clap had, but yeah, that's why I would expect we'd need to add additional structure to `RGArg` to make auto-generation as good as the manually curated script.

Either way, I suspect a mini-guide describing how the completion script works is an excellent next step. Thank you. :-)

kankri:
This is a bit off topic...

> Also, in general, I've found it pretty difficult to update the completion script. I knew I probably got some stuff wrong and I'm grateful you've caught it. :-) I can't tell whether a little bit of documentation describing the syntax would fix my problem editing the completion rules, or if it would be better to try to generate the completion script like we generate the man page (potentially by adding addition metadata on RGArg).

I'm surprised we are still using custom completion scripts for each command and each shell... Back when Bash didn't yet have support for user completion rules I had my own Bash fork which gave completions *and option help* when I typed `foo --bar<TAB>` by running `foo --help` and parsing and filtering the output. It worked great but I haven't bothered to keep maintaining it, most of the time I don't want to complete options, I just abbreviate them to keep command lines short.

To make things a bit less heuristic, each command could support some option, e.g. `foo --command-syntax-description`, which would output the option descriptions in a more machine readable form (JSON, YAML, ...). Or the description could be delivered as a separate file along with the binary and manual pages.

Maybe the Rust community can get a convention started?

BurntSushi:
@kankri I think it would be better if you sought out the Rust CLI working group: https://github.com/rust-lang-nursery/cli-wg There hasn't been too much discussion on completions there, so you might be able to get the ball rolling! https://github.com/rust-lang-nursery/cli-wg/search?q=completions&type=Issues

I personally don't have any plans to output ripgrep's options in a machine readable format. I do still think it's possible to generate completions with the same level of quality as our current manually curated completions, but this will probably require attaching more metadata to each flag, so it requires some work and I haven't thought through it completely yet. @okdana's excellent docs have honestly reduced my desire for this since I think I can now competently edit the completion file.

Otherwise, I don't really see any good reason to be "surprised." It's not surprising at all that good context sensitive completions require a lot more knowledge than what you typically see in a command's `--help` output.

PR #1018, opened 2018-08-20
Completion changes, &c., in light of libripgrep

okdana:
As requested in #1017, here's a proper PR for the feedback i had regarding zsh completion, documentation typos, and a `--no-multiline-dotall` option

Thanks again


BurntSushi:
Awesome, thank you!

PR #1020, opened 2018-08-20
Add zsh-completion reference guide

okdana:
As discussed in #1017 again — this adds a brief\* explanation of `_arguments` specifications to the zsh completion function, as a reference for other developers to use when making minor changes to the function. I mainly focussed on option specs, since that's the most common thing to change.

\* Not actually that brief, lol. Is it too wordy for a 'reference'? If you have suggestions as to how i could make it more 'skimmable' i'm definitely open to them. Or obv if you have any unanswered questions i can address them

ETA: Fixed very minor typo

### DIFF
--- a/complete/_rg
+++ b/complete/_rg
@@ -6,8 +6,8 @@
 # Run ci/test_complete.sh after building to ensure that the options supported by
 # this function stay in synch with the `rg` binary.
 #
-# @see http://zsh.sourceforge.net/Doc/Release/Completion-System.html
-# @see https://github.com/zsh-users/zsh/blob/master/Etc/completion-style-guide
+# For convenience, a completion reference guide is included at the bottom of
+# this file.
 #
 # Originally based on code from the zsh-users project — see copyright notice
 # below.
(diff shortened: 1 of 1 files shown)


complete/_rg
```diff
@@ -362,6 +362,157 @@ _rg_types() {
...
+#   '(-q -v --quiet --verbose)'{-q,--quiet}'[decrease output verbosity]'
+#   '!(-q -v --quiet --verbose)--silent'
+#   '(-q -v --quiet --verbose)'{-v,--verbose}'[increase output verbosity]'
+#   '--color=[specify when to use colors]:when:(always never auto)'
+#   '*:example file:_files'
+#
+# Although there may appear to be six specs here, there are actually nine; we
+# use brace expansion to combine specs for options that go by multiple names,
+# like `-q` and `--quiet`. This is customary, and ties in with the fact that zsh
+# merges completion possibilities together when they have the same description.
```
BurntSushi: Oh my. This one paragraph just clarified so many things.

okdana:
I'm not very knowledgeable about bash's completion system, but i know that it's functionally simpler than zsh's, and there's less abstraction built around it. It doesn't provide any of the fancy menu stuff that zsh has, and it's not especially configurable (mostly because there's not a lot to configure). I think it's possible to do some fancy things with it, but again since there's not much abstraction built around it you often have to do it manually (and so nobody bothers).

I think the `bash-completion` project provides some helpers to do certain things like handling long options with equals (`--color <TAB>` vs `--color=<TAB>`), but i haven't looked too much into that either.

zsh isn't super difficult to switch to, btw, if you ever feel compelled to try it. The default configuration is terrible, but if you set it up right it can be made to work almost like a super-set of bash. I was able to switch the default interactive shell in our product at work over to zsh without any of our tech-support people really even noticing the difference (besides the improved completion).

BurntSushi:
Drats. I was afraid you'd say that. I'm not a bash power user by any means, so I'm sure I could switch to zsh, but I fear the yak shave. Some day, perhaps. :-)

Issue #705, opened 2017-12-04
Ripgrep not working in OneDrive folders using Files On Demand


roblourens:
Ref: https://github.com/Microsoft/vscode/issues/35433

Files On Demand is a new OneDrive feature that only downloads files to the local disk when needed. It has caused some issues in Node/Electron land - https://github.com/nodejs/node/issues/12737, https://github.com/Microsoft/vscode/issues/27285.

When running ripgrep in a folder with this feature enabled, I just get `Access is denied. (os error 5)` immediately. Running with `--debug` doesn't produce any clues. I will investigate if you can point me in the right place to start looking?

BurntSushi:
I'm on mobile and headed to bed at the moment, but the first thing that
springs to mind is memory maps. If you run rg with --no-mmap (--mmap does
the opposite) then it will avoid using memory maps and instead use standard
file APIs. Linux, for example, has an issue where virtual files can't be
memory mapped. ripgrep detects this by looking at file size. If the size is
zero (which is what virtual files like /proc/cpuinfo report), then it
forcefully disables use of memory maps and everything works.

If memory maps are indeed the problem, and if there is a way to tell that
memory maps won't work with a stat call or similar then we can change
ripgrep's heuristics on Windows.

If this isn't the problem, then I personally would probably try to
reproduce the problem using a minimal Rust program, and then start
diagnosing from there. If it comes to that, how easy would it be for me to
reproduce the bug on my own Windows machine?

roblourens:
I get the same error with `--no-mmap`. Here is a fix in libuv for this onedrive feature: https://github.com/libuv/libuv/pull/1522/files, if this is a similar problem, it's probably in Rust's `fs`.

Repro steps are basically 

- Install the latest updates for Windows and OneDrive
- Open OneDrive settings, enable "Files On-Demand"
- Wait a couple minutes for OneDrive to do it's thing
- Invoke ripgrep in a OneDrive folder

roblourens:
Filed https://github.com/rust-lang/rust/issues/46484

roblourens:
Since there isn't much progress on the upstream issue, I'm trying to work around it in the ripgrep build for vscode. Would you be interested in a PR? The workaround is fairly simple, just need to check the `file_attributes` of a directory for the reparse point and directory flags: https://msdn.microsoft.com/en-us/library/windows/desktop/gg258117.aspx. And it needs to be done in a few places in the walkdir and ripgrep repos.

[...]

PR #1023, opened 2018-08-22
minor fixes & set version policy to "latest stable release of Rust"

### LINKED ISSUE #842: deb package lacks a man page

0.8.1

Linux

The deb package in ripgrep's release artifacts does not install the man page. This deb package is generated using the helpful [`cargo-deb`](https://github.com/mmstick/cargo-deb) tool. This tool basically "just works": I run it in the root of the ripgrep repo and out pops a deb package that can be installed in any Debian or Ubuntu system. This approach isn't idiomatic in the sense that every dependency of ripgrep is individually packaged, but the idiomatic approach is **extremely costly**. So costly, in fact, that I could never personally do it. But I can run `cargo deb`.

There is a small problem: the deb package doesn't contain a man page. I suspect the right answer here is to somehow teach `cargo deb` how to bundle a man page with the package it produces.

### LINKED ISSUE #953: `Walk` from `ignore` returns an Ok(...) and an Err(...) for the same permission-denied subdirectory
Skipped some of the preamble, since this isn't strictly a `ripgrep` question!

I'm using `ignore 0.4.2`, on Mac OSX 10.13.4.

When I use a `Walk` to iterate a directory containing a subdirectory with restricted permissions, it yields two separate entries for that subdirectory - one with an `Ok(DirEntry {...})` and one with an `Err(WithPath {...})`.

Also interestingly, calling the `DirEntry::error()` method from the `DirEntry` in the `Ok` returns `None`.

[This example repo](https://github.com/sashaweiss/ignore_permissions_example) will demonstrate this behavior!

I think I expected that calling `DirEntry::error()` on the `DirEntry {...}` in the `Ok` would give me back the same `WithPath` error, or some other indicator that `ignore` won't be able to recur into that directory.

Alternatively, if it's possible for a `Walk` to build a `DirEntry` even for a directory that is permission-restricted, maybe it makes sense for there to be another `Error` variant that contains a `DirEntry`? A `WithDirEntry {...}` perhaps?

I think in either case it'd make sense to eliminate the duplicate entry - in the first case by not returning the `Err`, and in the second case by not returning the `Ok`. Off the bat, not sure which would make the most sense. It's also not immediately clear to me how to make sure a solution to this makes sense for permission-restricted _files_, since so far my discussion of it has been focused on directories (and the fact that `ignore` won't recur into them).

My intuition says that the error comes about since `ignore` can detect the restricted directory while searching its parent directory, and that the error comes about when it tries to dive in - since this doesn't come up for restricted files. If you agree that this is something that should change, I'd be happy to try to help with a PR if you'd like!

Thanks in advance for your advice! `ignore` has been awesome to use and exactly what I needed - just want to confirm this behavior with you before building a workaround.

### LINKED ISSUE #984: DirEntry::path_is_symlink returns incorrect result for first result from ignore iterators
I'm using the `ignore` crate, version 0.4.2, but you don't seem to have a separate repository for it.

I added `ignore = "0.4"` to my `Cargo.toml`

    ssokolow@monolith ignore_test [master] % lsb_release -a
    No LSB modules are available.
    Distributor ID: Ubuntu
    Description:    Ubuntu 14.04.5 LTS
    Release:        14.04
    Codename:       trusty

Iterators from the `ignore` crate incorrectly identify the argument passed to them as a symlink when they return it during the first iteration.

This is actually quite the annoyance for a project I've just started, because it means I'm going to need to design some kind of minimally ugly hack to do a second syscall to get the *actual* value but only on the first iteration to avoid needlessly bogging down the following iterations.

    use std::fs::metadata;

    extern crate ignore;
    use ignore::Walk;

    fn main() {
        for result in Walk::new("./") {
            if let Ok(entry) = result {
                let verification = metadata(entry.path()).unwrap().file_type().is_symlink();
                println!("{}: {}", entry.path_is_symlink() == verification, entry.path().display());
            }
        }
    }

    ssokolow@monolith ignore_test [master] % cargo run
       Compiling ignore_test v0.1.0 (file:///home/ssokolow//src/ignore_test)
        Finished dev [unoptimized + debuginfo] target(s) in 4.72 secs
         Running `target/debug/ignore_test`
    false: ./
    true: ./Cargo.lock
    true: ./src
    true: ./src/main.rs
    true: ./target
    true: ./target/debug
    [output truncated for brevity as the rest is irrelevant]

(Note that the first entry is `false`, indicating that the result returned by `ignore`'s `path_is_symlink` doesn't match the result returned by explicitly querying the filesystem.)

    ssokolow@monolith ignore_test [master] % cargo run
       Compiling ignore_test v0.1.0 (file:///home/ssokolow//src/ignore_test)
        Finished dev [unoptimized + debuginfo] target(s) in 4.72 secs
         Running `target/debug/ignore_test`
    true: ./
    true: ./Cargo.lock
    true: ./src
    true: ./src/main.rs
    true: ./target
    true: ./target/debug
    [output truncated for brevity as the rest is irrelevant]

(Note that the first entry is `true` as well)

### LINKED ISSUE #1019: feedback wanted: do you care which version of the Rust compiler ripgrep targets?
In this issue, I'm asking for feedback from any user, and I'm particularly interested in getting feedback from folks that maintain ripgrep in some package repository.

What I'd like to know is: do you care which version of the Rust compiler ripgrep targets, and if so, how does it impact you?

There are at least a couple different aspects to this question. For example:

* That we document a minimum Rust compiler version at all.
* That we support Rust compilers older than a few months.

My policy for this has been relatively ad hoc, but fairly conservative in the context of the Rust ecosystem.  I've mostly done this out of "good sense," in that it seems better for ripgrep to compile on more versions of the Rust compiler than not. However, what I'd like to know is whether it's worth doing this. For example, if ripgrep changed to "only guaranteed to compile on the latest stable release of the Rust compiler," how would that impact you?

What I'm trying to figure out here is what kind of room I have to maneuver. For example, sometime around the end of the year, Rust 2018 will be released, and ripgrep will eventually want to move to that. What I'd like to figure out is _when_ that kind of move would be acceptable, and in order to do that, I need to ask people how the minimum supported Rust version impacts them. So any kind of experience you can offer would be great!

cc @svenstaro @roblourens @dstcruz @joshtriplett @infinity0 @carlwgeorge @ignatenkobrain @cuviper @sylvestre

### LINKED ISSUE #1022: organized flags in man page according to logical section?
The number of flags ripgrep has is truly staggering. A quick count appears to put it at `80`. Right now, these flags are listed alphabetically in the man page. I'm wondering whether it would be better to group them into logical sections.

For example, when PCRE2 support is enabled with the `--pcre2`, its Unicode support is enabled by default. ripgrep provides a flag, `--no-pcre2-unicode`, to disable PCRE2's Unicode support. However, the man page lists this as `--pcre2-unicode`, and states that it is enabled by default and can be disabled by using `--no-pcre2-unicode`. I did it this way so that the flag would be "discoverable" next to the `--pcre2` flag. But perhaps the better thing to do here is to just group flags.

I am somewhat inspired to do this based on ripgrep's zsh [completion script](https://github.com/BurntSushi/ripgrep/blob/master/complete/_rg), which defines its own groups. I'm not sure we want to make the groups in the man page as granular as that, but with the number of flags ripgrep has, it seems like a decent idea.

@okdana I'm especially interested to hear your opinion on this. :-)

BurntSushi:
This PR provides a smattering of minor fixes and cleanups. The biggest change here is actually putting #1019 into action and changing ripgrep's minimum Rust version policy to the latest stable release.

Fixes #842 
Fixes #953 
Fixes #984 
Fixes #1019 
Fixes #1022 

ref #705


complete/_rg
```diff
@@ -167,7 +167,7 @@ _rg() {
 
     + '(pcre2-unicode)' # PCRE2 Unicode options
     $no'(--no-pcre2-unicode)--pcre2-unicode[enable PCRE2 Unicode mode (with -P)]'
-    '(--no-pcre2-unicode)--no-pcre2-unicode[disable PCRE2 Unicode mode (with -P)]'
+    '(--pcre2-unicode)--no-pcre2-unicode[disable PCRE2 Unicode mode (with -P)]'
```
BurntSushi: cc @okdana I made this change based on your docs. Let me know if I got this wrong!

okdana: Oh, i was the one who got it wrong actually, sorry. :/

Neither of those exclusions are necessary since we've got the parentheses around the group name, but i *had* intended to exclude `--no-pcre2` (because there's no reason you'd change the PCRE2 Unicode mode if you were just going to disable PCRE2 anyway). I guess i just wasn't paying attention.

So both of the options in that group should actually have `(--no-pcre2)`.

PR #1027, opened 2018-08-26
add --one-file-system, --sort and --sortr

### LINKED ISSUE #321: Switch to avoid crossing filesystem boundaries
It's often useful to search only files that reside on the same filesystem as the search paths, same as ``find -xdev`` or ``ag --one-device``. I find ag's "one-device" switch to be misleading, as it's more correctly described by find as "do not cross filesystem boundaries", since the restriction applies to each search path in turn (which can span more than *one* device).

### LINKED ISSUE #404: sort output by file last modified time
I'm throwing this here for consideration. When I'm searching through my source files, I often not only want to find all occurrences, but when multiple hits are involved, I also want to find the most recent modification (by mtime). It would be nice to be able to sort output files (*with context*), sorted by time and reverse time.

Afaik, none of the various searching tools allow to do this. Not easily at least.

To do it currently, I actually do two passes:

- find file candidates fist
- sort files by mtime
- search within files

This is necessary as I obviously like to see the match context as well (otherwise I could stop at step 2). But having to find candidates first introduces extra overhead that cannot be avoided.

This requires buffering matches before showing the output, of course, but the search can be ran in parallel still.

BurntSushi:
Closes #321, Closes #404 

cc @okdana Mind checking if I got the completions right? If I got it this time, then I think I've grok'd it.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1,9 +1,9 @@
 [[package]]
 name = "aho-corasick"
-version = "0.6.6"
+version = "0.6.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "memchr 2.0.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "memchr 2.0.2 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
@@ -29,7 +29,7 @@ name = "base64"
 version = "0.9.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "byteorder 1.2.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "byteorder 1.2.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "safemem 0.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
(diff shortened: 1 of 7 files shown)


okdana:
No errors in the stuff you added, but:

* The exclusions for `--threads` should be changed from `(--sort-files)` to `(sort)` to make those groups mutually exclusive

* Since you're keeping `--sort-files` around for a while, i might suggest retaining a hidden spec for it, which would look like `!(threads)--sort-files`. (You could also leave one for `--no-sort-files`, but it would be irritating because then you can't just make the whole group exclusive — and it's got be a pretty rarely used option anyway — so i'd leave that out)

* I'd suggest changing the optarg description `sortby` to `sort method`; that'll make more sense in the description message that zsh produces

okdana:
btw, to give some context for the changes you're making:

screen shot 2018-08-25 at 23 02 29

BurntSushi:
@okdana I think I've fixed all of that!

Also, I spent half a day yak shaving. My default shell is now zsh. Let's see how this goes. :-)

BurntSushi:
@okdana I am pretty blown away by the auto completing for `--colors` specifically. Very nicely done.

okdana:
Thanks! Let me know if you need anything re: zsh

BurntSushi:
@okdana I think I might take you up on that! I sent you mail. :-)

PR #1043, opened 2018-09-08
grep: remove unneeded dependencies

ignatenkobrain:


BurntSushi:
I think removing `atty` is OK, but the others will be used fairly soon, so I'd rather just leave them there.

PR #1054, opened 2018-09-14
Ripgrep is also available in Ubuntu (from Cosmic)

sylvestre:


README.md
```diff
@@ -302,6 +301,11 @@ If you run Debian Buster (currently Debian testing) or Debian sid, ripgrep is
 $ sudo apt-get install ripgrep
 ```
 
+From Cosmic (18.10), ripgrep is also [available in Ubuntu](https://launchpad.net/ubuntu/+source/rust-ripgrep) (same packaging as Debian).
+```
+$ sudo apt-get install ripgrep
```
BurntSushi: Oh interesting. What is the difference between the `ripgrep` and `rust-ripgrep` packages?

BurntSushi:
woohooo! Thanks @sylvestre!

est31:
Nice, just thought today to check again on whether ripgrep will arrive in the next ubuntu or not. Thanks @sylvestre and @infinity0 ! Looking forward to the Ubuntu cosmic release.

sylvestre:
rust-ripgrep is the name of the source package.  We are prefixing all rust packages to avoid conflicts in the naming (example rust-num). This is usually transparent for the user.
ripgrep is the name of the binary package.

README.md
sdroege: The source package is called `rust-ripgrep`, the binary package is called `ripgrep`

PR #1058, opened 2018-09-17
Upgrade minimum reqiured version of Rust to 1.29

### LINKED ISSUE #916: Zombie processes cause rg --files to hang in /proc

```
ripgrep 0.8.1 (rev ab64da73ab)
-SIMD -AVX
```

```
> uname -s -r -v -m -p -i -o
Linux 4.16.7-1-ARCH #1 SMP PREEMPT Wed May 2 21:12:36 UTC 2018 x86_64 unknown unknown GNU/Linux
```

#### Describe the bug.

Even if it does not seem like a good idea in general to search the whole file system without excluding special folders like `/proc` (see also #311: *"TL;DR is: don't search /proc"*), I think that users will attempt to do so.

We recently tracked down one specific reproducible state which causes the `ignore` crate to hang while traversing `/proc` (see https://github.com/sharkdp/fd/issues/288 for more details). The same bug can be reproduced with `rg --files`.

1. Make sure that there are no zombie processes on your system (`ps -ef | rg '<defunct>'`)

2. Run `rg --files > /dev/null 2>&1` inside `/proc`. Everything should work fine (process exits after a few seconds).

3. Create a zombie process on purpose
    a. Copy the code from https://stackoverflow.com/a/25228579/704831 into a file called `zombie.c`
    b. Compile it `gcc -o zombie zombie.c`
    c. Run it: `./zombie`

4. Run `rg --files > /dev/null 2>&1` inside `/proc` again. This seems to hang indefinitely.

5. Get the PID of the zombie process via `ps -ef | rg '<defunct>'`. Call it `$PID`.

6. Run `rg --glob "\!$PID" --files > /dev/null 2>&1`. This should work fine, indicating that the directory traversal hangs in `/proc/$PID`.

`rg --files` hangs indefinitely inside the `/proc` folder.

`rg --files` traverses the `/proc` folder in finite time.

Note: `find`, for example, does not seem to have this problem and finishes in finite time even in the presence of zombie processes.

sharkdp:
This upgrades the minimum required version of Rust to 1.29 in order to fix #916 (Zombie processes cause `rg --files` to hang in `/proc`).

I have manually confirmed that the bug is fixed when compiling ripgrep with 1.29.

It's probably better to merge this just before the next release in order for the current Rust version to still be in present in the README(?).

See also:
- Rust compiler bug ticket: https://github.com/rust-lang/rust/issues/50619
- Rust compiler PR with the fix: https://github.com/rust-lang/rust/pull/50630

closes #916

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,13 +62,13 @@ matrix:
     # Minimum Rust supported channel. We enable these to make sure ripgrep
     # continues to work on the advertised minimum Rust version.
     - os: linux
-      rust: 1.28.0
+      rust: 1.29.0
       env: TARGET=x86_64-unknown-linux-gnu
     - os: linux
-      rust: 1.28.0
+      rust: 1.29.0
       env: TARGET=x86_64-unknown-linux-musl
     - os: linux
-      rust: 1.28.0
+      rust: 1.29.0
       env: TARGET=arm-unknown-linux-gnueabihf GCC_VERSION=4.8
       addons:
         apt:
--- a/README.md
+++ b/README.md
@@ -337,7 +337,7 @@ If you're a **NetBSD** user, then you can install ripgrep from
 
 If you're a **Rust programmer**, ripgrep can be installed with `cargo`.
 
-* Note that the minimum supported version of Rust for ripgrep is **1.28.0**,
+* Note that the minimum supported version of Rust for ripgrep is **1.29.0**,
   although ripgrep may work with older versions.
 * Note that the binary may be bigger than expected because it contains debug
   symbols. This is intentional. To remove debug symbols and therefore reduce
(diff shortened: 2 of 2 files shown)


BurntSushi:
> It's probably better to merge this just before the next release in order for the current Rust version to still be in present in the README(?).

Ah right yeah. Basically, to keep things sensible, I'd like to make sure any `0.x.y` releases build on the same version as the `0.x.0` release. So I won't bump the minimum Rust version until I know the next release is `0.11`.

I will just do that as part of the normal release process, so I'm going to close this. In theory, it could be months before a `0.11` release. :) Thank you for the reminder though and staying on top of this!

sharkdp:
> I'd like to make sure any `0.x.y` releases build on the same version as the `0.x.0` release

Ah, that makes sense.

> In theory, it could be months before a `0.11` release.

ripgrep 0.10 was six days early :smile:

PR #1071, opened 2018-09-27
Add Buildstream type

bochecha:
BuildStream is a Free Software tool for building/integrating software stacks.: https://buildstream.gitlab.io/buildstream/

It uses recipes written in YAML, in files with the `.bst` extension.

---

I wonder about the name if the type. I used `buildstream` because that's the name of the project, but the command line is `bst` so that's another possible name.

The latter is also shorter, which is nice when using `rg --type=bst …`.

Do you have a policy on this? Can we add both?

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -107,6 +107,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("awk", &["*.awk"]),
     ("bazel", &["*.bzl", "WORKSPACE", "BUILD"]),
     ("bitbake", &["*.bb", "*.bbappend", "*.bbclass", "*.conf", "*.inc"]),
+    ("buildstream", &["*.bst"]),
     ("bzip2", &["*.bz2"]),
     ("c", &["*.c", "*.h", "*.H", "*.cats"]),
     ("cabal", &["*.cabal"]),


BurntSushi:
> Do you have a policy on this? Can we add both?

Not particularly. My general policy has been a bit vague. Ideally, if we add very short abbreviations, then they should be popular enough that virtually anyone will recognize them. My sniff test has been that it's something I would recognize. I also have been trying to avoid multiple names that refer to the same thing, although there are some that exist, I'd prefer not to add more.

Unfortunately, I've never heard of buildstream before. Who is using it? I'm hesitant to add a short name.

bochecha:
> Unfortunately, I've never heard of buildstream before. Who is using it? I'm hesitant to add a short name.

Buildstream is mostly developed by a company called Codethink, so I guess they use it for their customers?

That being said, in terms of public open source projects using it, there's at least:

* the Freedesktop Sdk: https://gitlab.com/freedesktop-sdk/freedesktop-sdk/
* GNOME: https://gitlab.gnome.org/GNOME/gnome-build-meta/
* the Trustable project: http://trustable.gitlab.io/

Buildstream is a buildsystem for stacks of projects (e.g operating systems, container images, …) rather than individual projects (e.g the Autotools, Meson, Cargo, …). It pilots the latter for each projects integrated in the stack, builds everything reproducibly in a controlled environment in a build sandbox, etc…

Really impressive tech to build OSes with. :slightly_smiling_face:

BurntSushi:
Let's just go with the longer name for now. There are other things in the file types with names that are similarly long. Plus, if you have zsh completion setup, then file types will auto-complete. :-)

PR #1088, opened 2018-10-21
Docs: add instructions for openSUSE 15.0

mbologna:


README.md
```diff
@@ -263,6 +263,15 @@ $ sudo dnf copr enable carlwgeorge/ripgrep
 $ sudo dnf install ripgrep
 ```
 
+If you're an **openSUSE Leap 15.0** user, you can install ripgrep from the
+[utilities repo](https://build.opensuse.org/project/show/utilities/ripgrep):
```
BurntSushi: I get a 404 error for this.

README.md
```diff
@@ -263,6 +263,15 @@ $ sudo dnf copr enable carlwgeorge/ripgrep
...
+
+```
+$ sudo zypper ar https://download.opensuse.org/repositories/utilities/openSUSE_Leap_15.0/utilities.repo
```
BurntSushi: Does this need to be updated for every new major release of openSUSE?

README.md
```diff
@@ -263,6 +263,15 @@ $ sudo dnf copr enable carlwgeorge/ripgrep
 $ sudo dnf install ripgrep
 ```
 
+If you're an **openSUSE Leap 15.0** user, you can install ripgrep from the
```
BurntSushi: It looks like we already have openSUSE Tumbleweed instructions. Is there any way to collapse them into one section? What's the difference? (I don't know what the difference between them is.

mbologna: Leap and Tumbleweed are very different: the difference is verbosely explained in https://en.opensuse.org/Portal:Tumbleweed and https://www.opensuse.org/#Leap

Maybe we can create a super section called "openSUSE" and then introduce the two flavors to get ripgrep

README.md
mbologna: Yes, this depends on the release

README.md
mbologna: You are right. Fixed

PR #1100, opened 2018-11-03
Support Brotli and Zstd decompression

### LINKED ISSUE #1099: Add zstd (.zst) to the --search-zip support (extension of #539)

git master

Just need an additional list entry in https://github.com/BurntSushi/ripgrep/blob/master/grep-cli/src/decompress.rs#L349 and a small update to the manual page.  Kudos to @balajisivaraman for already having done the work to easily add more decompressors in a general fashion.

Something like:

```
+ const ARGS_ZSTD: &[&str] = &["zstd", "-d", "-c"];
...
+     cmd("*.zst", ARGS_ZSTD),
```

I can't find where the OPTIONS (`-z, --search-zip`) portion of the manual page is stored in the directory hierarchy, but that should be updated as well.

Should be pretty trivial!  Thanks for considering it.

okdana:
Fixes #1099

I added Brotli because it feels kind of related in design, utility, and commonness to Zstd. Let me know if that was a bad choice

I also updated the ignore types. Not sure how useful it is to have those in there, but they were updated the last time a compression format was added, so...


BurntSushi:
(CI failure looks unrelated.)

okdana:
This one does lack the `.zstd` extension that @chipturner included, which seems like a good addition. If we do go with mine i should probably add that in

BurntSushi:
@okdana Ah yeah, that'd be great!

okdana:
Updated and rebased

PR #1111, opened 2018-11-17
Attempt to add tar support

upsuper:
Some preliminary attempt of adding tar support.

The current approach is simple, but probably not ideal. We probably should handle tarball like directory rather than hacking into the file searching...

I can investigate further if you agree this is something worth doing.


BurntSushi:
Thanks for working on this! This really should have an issue for it where we can discuss what the actual semantics should be, if any. I've historically been skeptical of adding tar archive support, but if we can nail down what the desired behavior should be in detail, then I could be convinced otherwise. :)

upsuper:
Filed #1112 for this.

BurntSushi:
I'm going to close this for now. In terms of the implementation here, this mostly looks good, but unfortunately it does not interact well with how ignore rules are currently implemented. There's more discussion in #1112.

Issue #1116, opened 2018-11-21
ignore crate: Show matched FileTypeDef?


firesock:
#### What version of ripgrep are you using?

ignore-0.4.4

#### How did you install ripgrep?

ignore crate via crates.io

#### What operating system are you using ripgrep on?

Linux

#### Describe your question, feature request, or bug.

Is it feasible to make ignore's `Walk` return `DirEntry`'s that are aware of the `FileTypeDef` they matched on, if any?

My use case is to use `Walk` to iterate and depending on the name registered with the matching `FileTypeDef` do some different logic.

As far as I can tell, this information isn't accessible easily right now. Even if I used `Types` and called `matched`, the `Glob` struct doesn't return the `FileTypeDef` and hence the name registered.

firesock:
Okay thanks, is the intended goal to use `Types` and `matched` after a `DirEntry` is returned, and not expose it on `DirEntry` at all?

BurntSushi:
If you can make that work, then I think that would be ideal. Exposing it through the directory iterator would require more API design work.

PR #1120, opened 2018-11-25
ignore/types: Add method to retrieve matching file type def from a Glob

firesock:
Implements first solution from #1116

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -346,6 +346,16 @@ impl<'a> Glob<'a> {
     fn unmatched() -> Glob<'a> {
         Glob(GlobInner::UnmatchedIgnore)
     }
+
+    /// Return the file type defintion that matched, if any
+    pub fn file_type_def(&self) -> Option<&FileTypeDef> {
+        match self {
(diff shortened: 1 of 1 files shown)


ignore/src/types.rs
```diff
@@ -346,6 +346,16 @@ impl<'a> Glob<'a> {
     fn unmatched() -> Glob<'a> {
         Glob(GlobInner::UnmatchedIgnore)
     }
+
+    /// Return the file type defintion that matched, if any
```
BurntSushi: Could you fill out the docs a bit more here? Please also use a style consistent with the comments in the surrounding code (e.g., use punctuation).

I don't think you need much more here. But instead of saying, "if any," maybe we can say when a file type def is actually returned, e.g., if there was a match. When there's no match, then there's no available file type def.

BurntSushi:
I ended up just touching up the docs myself and committing the change here https://github.com/BurntSushi/ripgrep/commit/44a9e377371006f01f2e425fbf8d49e43bc14a9b. Thanks!

PR #1122, opened 2018-11-27
Fix #1121 - avoid triggering windows antimalware when in --files mode

### LINKED ISSUE #1121: rg --files triggers Windows Antimalware
https://github.com/BurntSushi/ripgrep/issues/600 has regressed, it looks like that fix got lost

image

I confirmed that commenting out this section fixes the problem: https://github.com/BurntSushi/ripgrep/blob/master/ignore/src/walk.rs#L1407

So we just need a fix similar to the earlier one that avoids this check when in `--files` mode.

roblourens:
This looks basically equivalent to the earlier fix for #600

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -782,7 +782,7 @@ impl ArgMatches {
             .max_filesize(self.max_file_size()?)
             .threads(self.threads()?)
             .same_file_system(self.is_present("one-file-system"))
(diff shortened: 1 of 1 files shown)


roblourens:
We are now shipping this in vscode and it seems to fix the issue.

BurntSushi:
Thanks!

PR #1130, opened 2018-12-01
--files-without-match / --files-with-matches issue when providing only a single path

### LINKED ISSUE #1106: --files-without-match reports an error when used with a single file path

ripgrep 0.10.0
-SIMD -AVX (compiled)

cargo install ripgrep

Windows 10

I'm trying to get a list of files that contain one word, but not another word. To do that I call "rg -l foo" to get a list of files that I pass on to another call with "rg --files-without-match bar". The call looks something like this: rg -l foo | xargs -i -d '\n' rg --files-without-match bar {}

This doesn't work because --files-without-match cannot be used with a single file. The error message is " output kind PathWithoutMatch requires a file path". However I don't understand why that shouldn't be possible.
Maybe there's a better way to do this? If not I'd really appreciate it if that could be implemented.

Xervon:
This should fix #1106 

When determining if file names should be output despite --with-filename not being set explicitly the length of the paths array should be larger than 0 rather than 1 to activate file name output.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -1448,6 +1448,8 @@ impl ArgMatches {
         } else {
             self.is_present("with-filename")
             || self.is_present("vimgrep")
+            || self.is_present("files-with-matches")
+            || self.is_present("files-without-match")
             || paths.len() > 1
             || paths.get(0).map_or(false, |p| p.is_dir())
         }
--- a/src/args.rs
+++ b/src/args.rs
@@ -1448,7 +1448,7 @@ impl ArgMatches {
         } else {
             self.is_present("with-filename")
             || self.is_present("vimgrep")
-            || paths.len() > 0
+            || paths.len() > 1
             || paths.get(0).map_or(false, |p| p.is_dir())
         }
     }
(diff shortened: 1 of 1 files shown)


src/args.rs
```diff
@@ -1448,7 +1448,7 @@ impl ArgMatches {
         } else {
             self.is_present("with-filename")
             || self.is_present("vimgrep")
-            || paths.len() > 1
+            || paths.len() > 0
```
BurntSushi: Thanks for working on this bug! But I don't think this is right. `rg foo file` should not be printing the file path by default.

Xervon:
I obviously also didn't call `cargo test` yesterday. This should be better. At least `cargo test` doesn't complain locally as it would have done yesterday.

Xervon:
Should I also add another test for this?

BurntSushi:
Thanks for this attempt! Unfortunately, this wasn't quite right. The issue was a bit more subtle than this. See https://github.com/BurntSushi/ripgrep/commit/a7f2d482342eb2250b5a32ee03a8fe16990228dc for more details!

PR #1146, opened 2018-12-21
Move from `trim_right` to `trim_end`

jwbowen:
The `trim_right` method is deprecated in favor of `trim_end` for
`std::string::String` to more naturally support right-to-left
languages. The `trim_end` method was introduced in Rust 1.30.0, and the
current stable version is 1.31.1, so it feels safe to make the change
at this point in time.

### DIFF
--- a/ignore/src/gitignore.rs
+++ b/ignore/src/gitignore.rs
@@ -423,7 +423,7 @@ impl GitignoreBuilder {
             return Ok(self);
         }
         if !line.ends_with("\\ ") {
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thank you, but the code churn isn't worth it at this time. I don't know whether there will be a ripgrep 0.10 patch release or not, which must continue to compile on the same Rust version as 0.10.0. Feel free to re-submit once ripgrep has bumped its minimum Rust version though!

PR #1158, opened 2019-01-10
Easily build static binaries and start developing

geropl:
With this PR everyone browsing GitHub can easily try out ripgrep and start contributing by using Gitpod, a free dev environment service with rich GitHub integration.

The PR contains the config for Gitpod that automatically executes `cargo build` and `cargo test --all` and allows to build static binaries using `cargo build --release --target x86_64-unknown-linux-musl`, without the need for any setup on the user side.

Here is a screenshot:
image

[Open in Gitpod](https://gitpod.io#snapshot/74880e80-9911-4ca2-8b91-490228153d2b)

### DIFF
--- a/.gitpod.yml
+++ b/.gitpod.yml
@@ -1,0 +1,5 @@
+image:
+  file: gitpod.Dockerfile
+tasks:
+- init: cargo build
+  command: cargo test --all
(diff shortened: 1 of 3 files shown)


svenefftinge:
Gitpod / Theia is actually using ripgrep for search. Full circle! :)

BurntSushi:
I'm happy to hear that Theia is using ripgrep! :-)

With that said, I don't think this functionality is a good fit for ripgrep. I'm not aware of this being an issue in practice, and I'd prefer not to maintain the Dockerfile for this.

PR #1170, opened 2019-01-21
Add --git-ignore-case-insensitive

### LINKED ISSUE #1164: Feature request: case-insensitive .gitignore matching on Windows

ripgrep 0.10.0 (rev 8a7db1a918)
-SIMD -AVX (compiled)

Downloaded ripgrep-0.10.0-i686-pc-windows-msvc.zip from the release page and installed rg.exe to a location in my path.

Windows 10 Pro
Version 1803 (OS Build 17134.523)

#### Describe your feature request.

I'd like ripgrep to support case-insensitive .gitignore matching.

On Windows, the file system is case insensitive which inspired this issue #163 and the subsequent --iglob flag. However, it looks like the flag can't be used to make exclusions from the .gitignore case insensitive.

Here's an example on the Windows command line. You'll notice that `git status` doesn't list `myfile.txt`, but ripgrep does (of course on linux this isn't the case, and the file is listed twice).

```
> mkdir rgtest
> cd rgtest
> git init
> echo foo > myfile.txt
> echo MYFILE.txt > .gitignore
> git status -s
?? .gitignore
> rg -l foo
myfile.txt
```

Sorry if I'm missing something obvious, or if this is covered by another issue. I tried searching the man page and didn't find any configuration regarding .gitignore, and found *many* issues related to case insensitive file matching but none as far as I could tell that asked for this specific feature.

davidtorosyan:
This closes #1164

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -235,6 +235,10 @@ Like `.gitignore`, a `.ignore` file can be placed in any directory. Its rules
 will be processed with respect to the directory it resides in, just like
 `.gitignore`.
 
+To process `.gitignore` and `.ignore` files case insensitively, use the flag
+`--git-ignore-case-insensitive`. This is especially useful on case insensitive
+file systems like those on Windows and macOS.
+
 For a more in depth description of how glob patterns in a `.gitignore` file
 are interpreted, please see `man gitignore`.
 
--- a/complete/_rg
+++ b/complete/_rg
@@ -99,6 +99,10 @@ _rg() {
     {-L,--follow}'[follow symlinks]'
     $no"--no-follow[don't follow symlinks]"
 
+    + '(git-ignore-case)' # Ignore-file case sensitivity options
+    '--git-ignore-case-insensitive[process ignore files case insensitively]'
+    '--git-ignore-case-sensitive[process ignore files case sensitively]'
+
     + glob # File-glob options
     '*'{-g+,--glob=}'[include/exclude files matching specified glob]:glob'
     '*--iglob=[include/exclude files matching specified case-insensitive glob]:glob'
--- a/ignore/src/dir.rs
+++ b/ignore/src/dir.rs
@@ -73,6 +73,8 @@ struct IgnoreOptions {
     git_ignore: bool,
     /// Whether to read .git/info/exclude files.
     git_exclude: bool,
+    /// Whether to ignore files case insensitively
+    git_ignore_case_insensitive: bool,
 }
 
 /// Ignore is a matcher useful for recursively walking one or more directories.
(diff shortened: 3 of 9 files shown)


okdana:
I think that based on the names of the other ignore-file-related options it would probably just be like `--ignore-case-insensitive`, right? And you usually don't need to document the option that negates it separately, especially when it's the default behaviour; see how options like `--no-ignore-messages` and `--no-multiline-dotall` are implemented, for example. (That's related to the `$no` thing in the completion function)

davidtorosyan:
I considered `--ignore-case-insensitive`, but thought it might be confused as a variant of `--ignore-case`. I didn't use `$no` because `--ignore-case` and `--case-sensitive` don't.

Maybe `--ignore-iglob` and `--no-ignore-iglob`?

okdana:
Ah, i didn't consider the name confusion. Mixing the concepts of ignore files and command-line globbing seems a bit suspect to me as well, but i'm not sure.

I do feel like the option should be documented/completed more like `--[no-]ignore-messages` than `--case-insensitive` &al., though, in any case. Those match-case options aren't quite analogous (there's three of them, for one thing).

davidtorosyan:
Both good points.

I think the clearest then would probably be `--[no-]ignore-file-case-insensitive`, but I'll wait for @BurntSushi to weigh in before updating the PR.

BurntSushi:
@davidtorosyan Thanks so much for working on this! I think you got the general approach correct, but there were a lot of details that I wanted to iron out, so I just did them myself here in https://github.com/BurntSushi/ripgrep/commit/718a00f6f2f88238546f7d33c1ea52217002495e. You can see my changes off your PR here: https://gist.github.com/BurntSushi/6456ad9be5ce650fa36bf02956cee14e

In summary:

* There were some formatting issues that were inconsistent with the rest of the code. In particular, there were some very long lines. I'd like to adhere to 79 columns inclusive.
* There were several breaking changes made to the public API of the `ignore` crate. This is not something we should do lightly. I do see why you did it though, because avoiding that did require a touch of refactoring that wasn't at all obvious. (`ignore` badly needs a facelift.)
* I decided on just using `--ignore-file-case-insensitive` as the flag name in order to avoid confusion with `--ignore-case`.
* The flag overrides just mean that all/most flags can be overridden. Following the pattern of `--ignore-case` is understandable, but `--case-sensitive` was added before we started getting more rigorous about creating flags for disabling other flags. These days, we consistently try to just prefix flags with a `--no-` in order to indicate disabling them. Moreover, since most flags have that variant, we hide them from the listing of flags since there are so many.
* Short names are in short supply. This definitely does not deserve a short name. There are literally only a few short flag names left.

I think that's it! Nice work!

BurntSushi:
@davidtorosyan Yup! Basically, you want to take a look at the public API in the docs, e.g., https://docs.rs/ignore. Your changes made sense though---those methods should not have been returning `Result`s. I added a TODO comment to update them on the next breaking release.

PR #325, opened 2017-01-16
Enable LTO (Link Time Optimization)

wdv4758h:
With this optimization, the size of stripped ripgrep's binary can reduce from `2137936` to `2117424` bytes on my machine 😃

PR #1187, opened 2019-02-07
Use LTO in release mode

lilianmoraru:
LTO in release mode, for your consideration :)
[More on `incremental = false`](https://github.com/jwilm/alacritty/issues/1115#issuecomment-460822456)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,6 +78,8 @@ pcre2 = ["grep/pcre2"]
 
 [profile.release]
 debug = 1
+lto = true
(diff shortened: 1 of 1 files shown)


BurntSushi:
Sorry, but I'm going to have to decline. This has already been [discussed](https://github.com/BurntSushi/ripgrep/pull/325).

If there's new evidence to bring to the table, then I'd be happy to review and revisit it.

lilianmoraru:
Should of measured first, sorry.

Issue #1144, opened 2018-12-16
Multiple testsuite failures on big-endian targets


glaubitz:
The testsuite of ```ripgrep``` 0.10.0 is failing on all of Debian's big-endian targets:

```
failures:

---- feature::f34_only_matching_line_column stdout ----
thread 'feature::f34_only_matching_line_column' panicked at '
printed outputs differ!

expected:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
sherlock:1:57:Sherlock
sherlock:3:49:Sherlock

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

got:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
sherlock:1:57:Sherlock
sherlock:2:49:Sherlock

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
', tests/feature.rs:110:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: std::panicking::begin_panic_fmt
   7: integration::feature::f34_only_matching_line_column::{{closure}}
             at tests/feature.rs:110
   8: integration::feature::f34_only_matching_line_column
             at tests/macros.rs:7
   9: integration::feature::f34_only_matching_line_column::{{closure}}
             at tests/macros.rs:5
  10: core::ops::function::FnOnce::call_once
             at libcore/ops/function.rs:238
  11: <F as alloc::boxed::FnBox<A>>::call_box
  12: __rust_maybe_catch_panic

---- feature::f917_trim stdout ----
thread 'feature::f917_trim' panicked at '
printed outputs differ!

[...]

PR #1194, opened 2019-02-10
run cargo update

llogiq:
This (besides other updates) pulls the bytecount fix for #1144


BurntSushi:
@llogiq Thanks! I prefer to do this in a way that maintains a clean history (in addition to reverting the patch to `grep-searcher` to work-around the bug).

Issue #1207, opened 2019-02-28
finding zero bytes in utf-16 encoded files


LesnyRumcajs:
#### What version of ripgrep are you using?
ripgrep 0.10.0
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)

#### How did you install ripgrep?
cargo install ripgrep

#### What operating system are you using ripgrep on?

Fedora 29

#### Describe your question, feature request, or bug.
I'm struggling to find files that contain 00 bytes. I created an UTF-16 LE file with text `test` (hexdump)
```
00000000: fffe 7400 6500 7300 7400 0a00            ..t.e.s.t...
```
Given that there are 00 bytes inside I'm issuing a command `rg -cuuu '(?-u:\x00)'` but get no results at all. It works for searching for `t`, like 
```
rg -cuuu '(?-u:\x73)'
test.txt:1
```
From my understanding the `-uuu` flag along with some UTF escaping should do the trick. It works fine for non-zero bytes. It also works for binary files (tried with a file comprising of a single 00 byte). Am I missing something?

lespea:
I think ripgrep is seeing the BOM and properly decoding the test as utf-16 (hence no null bytes).  Try using the `--no-encoding` flag?

```
    -E, --encoding <ENCODING>
            Specify the text encoding that ripgrep will use on all files searched. The
            default value is 'auto', which will cause ripgrep to do a best effort automatic
            detection of encoding on a per-file basis. Automatic detection in this case
            only applies to files that begin with a UTF-8 or UTF-16 byte-order mark (BOM).
            No other automatic detection is performend.

            Other supported values can be found in the list of labels here:
            https://encoding.spec.whatwg.org/#concept-encoding-get

            For more details on encoding and how ripgrep deals with it, see GUIDE.md.

            This flag can be disabled with --no-encoding.
```

LesnyRumcajs:
@lespea I'm pretty sure that's the case (removing BOM makes the command find the bytes, it's treating it as a plain binary). I don't know how to get over it though. Using `--no-encoding` as in `rg --no-encoding -cuuu '(?-u:\x00)'` still doesn't detect null bytes in UTF-16 LE file.

BurntSushi:
[...]

PR #1208, opened 2019-03-04
add option to disable bom sniffing

LesnyRumcajs:
https://github.com/BurntSushi/ripgrep/issues/1207

### DIFF
--- a/complete/_rg
+++ b/complete/_rg
@@ -378,7 +378,7 @@ _rg_encodings() {
     shift{-,_}jis csshiftjis {,x-}sjis ms_kanji ms932
     utf{,-}8 utf-16{,be,le} unicode-1-1-utf-8
     windows-{31j,874,949,125{0..8}} dos-874 tis-620 ansi_x3.4-1968
-    x-user-defined auto
+    x-user-defined auto none
   )
 
   _wanted encodings expl encoding compadd -a "$@" - _encodings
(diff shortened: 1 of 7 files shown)


BurntSushi:
@LesnyRumcajs Thanks so much for this! I ended up re-working this PR a bit and re-submitted it: #1237. I think I might have no explained the right path to take very well. In particular, a problem here is that `EncodingMode` was added to the public API of the `grep-searcher` crate _and_ a breaking change was introduced by changing the method signature of the `SearcherBuilder::encoding` builder. What I had intended was for `EncodingMode` to be an internal detail used in ripgrep's core `src/args.rs`. That's what I meant by "forwarding" the `bom_sniffing` option that you added to `encoding_rs_io` up through `grep-searcher::SearcherBuilder`.

More subtle than that was that when BOM sniffing was disabled, the BOM was still being stripped, and I don't think that's correct since the point of this is to get at the raw bytes directly, even if there's a BOM and including the BOM. This led to uncovering another subtle but in `encoding_rs_io`, which I fixed in [this commit](https://github.com/BurntSushi/encoding_rs_io/commit/4146e3d96a131944fe0e8576e71ad8c951472b11).

Thanks again for your work on this. Mostly everything you wrote I kept; it just needed to be shuffled around a bit. :-)

PR #1234, opened 2019-04-03
ci: switch to xenial

BurntSushi:
Rust is having problems with trusty, in particular, see this bug I
filed: https://github.com/rust-lang/rust/issues/59411

This was purpotedly fixed in
https://github.com/rust-lang/rust/pull/59468,
but it appears the issue is still occurring.

This commit tries to update to Ubuntu 16.04 in the hope that it will fix
this problem.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: rust
+dist: xenial
 env:
   global:
     - PROJECT_NAME: ripgrep


BurntSushi:
Well... That was easy.

lilianmoraru:
Also, Trusty is EOF in a few days.

BurntSushi:
@lilianmoraru Tell that to Travis CI. :-) Trusty is still the default Linux flavor used by them. (I've been very leisurely considering switching to Azure in anticipation of Travis CI going into the shitter.)

PR #1239, opened 2019-04-09
ignore/types: add lockfiles

tpai:
Add npm, yarn, composer and ruby gems lockfiles

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -161,7 +161,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("js", &[
         "*.js", "*.jsx", "*.vue",
     ]),
-    ("json", &["*.json", "composer.lock"]),
+    ("json", &["*.json"]),
     ("jsonl", &["*.jsonl"]),
     ("julia", &["*.jl"]),
     ("jupyter", &["*.ipynb", "*.jpynb"]),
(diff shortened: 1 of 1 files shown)


ignore/src/types.rs
```diff
@@ -196,6 +196,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
         "OFL-*[0-9]*",
     ]),
     ("lisp", &["*.el", "*.jl", "*.lisp", "*.lsp", "*.sc", "*.scm"]),
+    ("lockfiles", &["composer.lock", "package-lock.json", "yarn.lock", "Gemfile.lock"]),
```
BurntSushi: Maybe call this `lock` instead? It's a bit shorter.

Also, you might as well add `Cargo.lock` to this.

And please make sure the style stays consistent with the surrounding code. i.e., No lines over 79 columns please.

ignore/src/types.rs
```diff
@@ -161,7 +161,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("js", &[
         "*.js", "*.jsx", "*.vue",
     ]),
-    ("json", &["*.json", "composer.lock"]),
+    ("json", &["*.json"]),
```
BurntSushi: I'm not so sure about this change. We also have `Cargo.lock` listed under the `toml` type, so this change is inconsistent with that. The thing is that we know `composer.lock` to be a JSON file, right? So it really should be searched when you ask to search for JSON files. Similarly for `Cargo.lock`, which we know to be TOML.

Maybe leave `composer.lock` here? We can still keep a separate overlapping `lock` type.

tpai: I see, I will leave `composer.lock` with no change.

PR #1242, opened 2019-04-09
Support for files for type `java`

hupfdule:
- `*.jspx` for XHTML JSP files
- `*.properties` for Java Properties files (resource bundles, etc.)
- `pom.xml` for Maven project build files

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -156,7 +156,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("hs", &["*.hs", "*.lhs"]),
     ("html", &["*.htm", "*.html", "*.ejs"]),
     ("idris", &["*.idr", "*.lidr"]),
-    ("java", &["*.java", "*.jsp"]),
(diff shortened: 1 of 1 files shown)


ignore/src/types.rs
```diff
@@ -156,7 +156,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("hs", &["*.hs", "*.lhs"]),
     ("html", &["*.htm", "*.html", "*.ejs"]),
     ("idris", &["*.idr", "*.lidr"]),
-    ("java", &["*.java", "*.jsp"]),
+    ("java", &["*.java", "*.jsp", "*.jspx", "*.properties", "pom.xml"]),
```
BurntSushi: I'm not well versed in Java, but [according to this page](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html), `pom.xml` is an XML file. If I were a Java programmer, I'd be surprised if `rg -tjava` searched XML data. (Just like I'd be surprised if `rg -trust` searched `Cargo.toml`.)

BurntSushi:
I merged this PR, but dropped the `pom.xml` entry.

hupfdule:
Yes, I think that's ok.
Thank you!

PR #1243, opened 2019-04-09
Support for files for type `xml`

hupfdule:
- `*.dtd` for Document Type Definitions
- `*.xsl` and `*.xslt` for XSL Transformation descriptions
- `*.xsd` for XML Schema definitions
- `*.xjb` for JAXB bindings
- `*.rng` for Relax NG files
- `*.sch` for Schematron files

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -301,7 +301,10 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("vimscript", &["*.vim"]),
     ("wiki", &["*.mediawiki", "*.wiki"]),
     ("webidl", &["*.idl", "*.webidl", "*.widl"]),
-    ("xml", &["*.xml", "*.xml.dist"]),
+    ("xml", &[
+        "*.xml", "*.xml.dist", "*.dtd", "*.xsl", "*.xslt", "*.xsd", "*.xjb",
+        "*.rng", "*.sch",
+    ]),
     ("xz", &["*.xz", "*.txz"]),
     ("yacc", &["*.y"]),
     ("yaml", &["*.yaml", "*.yml"]),


ignore/src/types.rs
```diff
@@ -301,7 +301,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("vimscript", &["*.vim"]),
     ("wiki", &["*.mediawiki", "*.wiki"]),
     ("webidl", &["*.idl", "*.webidl", "*.widl"]),
-    ("xml", &["*.xml", "*.xml.dist"]),
+    ("xml", &["*.xml", "*.xml.dist", "*.dtd", "*.xsl", "*.xslt", "*.xsd", "*.xjb", "*.rng", "*.sch"]),
```
BurntSushi: I think these all look good to me.

Could you please use a style consistent with the surrounding code? In particular, please wrap code to 79 columns inclusive.

Thanks!

hupfdule: No problem, I will update it.

Do you want me to add another commit or can i force-push the change to the existing commit (rewriting history there)?

BurntSushi: Force push please, if convenient, in order to keep this as a single commit. (Otherwise, I will squash this into one commit. I try to keep ripgrep's revision history very clean.)

hupfdule: Done.

ignore/src/types.rs
```diff
@@ -301,7 +301,8 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("vimscript", &["*.vim"]),
     ("wiki", &["*.mediawiki", "*.wiki"]),
     ("webidl", &["*.idl", "*.webidl", "*.widl"]),
-    ("xml", &["*.xml", "*.xml.dist"]),
+    ("xml", &["*.xml", "*.xml.dist", "*.dtd", "*.xsl", "*.xslt", "*.xsd",
+        "*.xjb", "*.rng", "*.sch"]),
```
BurntSushi: Sorry to be a pain about this, but please format the code in a style that is consistent with the surrounding code. Please consult other file types, such as `systemd`, for how to format the list when it gets to be too long.

hupfdule: No problem. Didn't notice the difference.

It did this change now.

(BTW: The definition of `asp` is 87 characters long. ;-))

BurntSushi: Yes, I don't always catch violations of the line lengths, especially if they're close. (Because the GitHub UI doesn't make it easy to see.) And I rarely edit this file myself any more, so I didn't even know about it. :-)

Issue #1078, opened 2018-10-05
-M should show a preview of the line  if the line is too long


dylan-chong:
#### What version of ripgrep are you using?

ripgrep 0.10.0
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)

#### How did you install ripgrep?

Homebrew

#### What operating system are you using ripgrep on?

Mac high sierra 10.13.6

#### Describe your question, feature request, or bug.

When I specify a line length with -M, and a line is too long, I cannot see
anything about what the line is. The silver searcher does what I believe to be
a more useful behaviour, which is show a preview of the line for the first lot
of characters. This is much more useful because I can tell whether this line is
generated JavaScript or not, especially if I specify the line length of -M to
be something higher like 100.

Please change the behaviour to be more like the silver searcher

```
# Current behaviour
> \rg -M 15 Hello
ios/ToDoLater.xcodeproj/project.pbxproj
696:[Omitted long line with 1 matches]

# Ideal behaviour
> \ag -W 15 Hello
ios/ToDoLater.xcodeproj/project.pbxproj
696:                    productName  [...]

# What the line looks like without the column limit
> \rg Hello
ios/ToDoLater.xcodeproj/project.pbxproj
696:                    productName = "Hello World";
```

BurntSushi:
I don't think I'm willing to change the existing behavior here.

The problem I have with ag's behavior is that it makes it much less obvious that a line has been truncated. I guess this is more of a display/UX problem that could be solved as opposed to a problem with the idea itself.

If we do this, it will probably have to be behind a flag, e.g., `--max-columns-preview`.

okdana:
For whatever it might be worth, i found ripgrep's `-M` surprising and unhelpful when i first switched over as well. Not just because my brain's been tainted by `ag` use (though that may be part of it), but because the proposed (`ag`-like) behaviour seems consistent with the way the other `--max-*` options work; `--max-count` just stops after that many matches, and `--max-depth` just stops after that many directory levels, so i expected `--max-columns` to just stop after that many columns.

I know how much you care about BC, and i don't often run into files with stupidly long lines anyway, so i never mentioned it — but i ended up just removing `-M` from my alias because, whenever i did encounter them, having it there was more irritating than not.

I don't currently have any strong opinions on the UX/BC stuff other than that, but there's another perspective for you, in any case

BurntSushi:
[...]

PR #1245, opened 2019-04-14
better binary filtering UX + hybrid regexes + max column preview + and more

BurntSushi:
This PR fixes several bugs and adds a few new features that I want to get into the next release. Notably, this prepares the next release to be `ripgrep 11.0.0` and _not_ `ripgrep 0.11.0`.

### DIFF
--- a/complete/_rg
+++ b/complete/_rg
@@ -85,7 +85,7 @@ _rg() {
 
     + '(file-name)' # File-name options
     {-H,--with-filename}'[show file name for matches]'
-    "--no-filename[don't show file name for matches]"
+    {-I,--no-filename}"[don't show file name for matches]"
 
     + '(file-system)' # File system options
     "--one-file-system[don't descend into directories on other file systems]"
--- a/complete/_rg
+++ b/complete/_rg
@@ -148,6 +148,10 @@ _rg() {
     $no"--no-crlf[don't use CRLF as line terminator]"
     '(text)--null-data[use NUL as line terminator]'
 
+    + '(max-column-preview)' # max column preview options
+    '--max-column-preview[show preview for long lines (with -M)]'
+    $no"--no-max-column-preview[don't show preview for long lines (with -M)]"
+
     + '(max-depth)' # Directory-depth options
     '--max-depth=[specify max number of directories to descend]:number of directories'
     '!--maxdepth=:number of directories'
--- a/complete/_rg
+++ b/complete/_rg
@@ -43,6 +43,7 @@ _rg() {
     + '(exclusive)' # Misc. fully exclusive options
     '(: * -)'{-h,--help}'[display help information]'
     '(: * -)'{-V,--version}'[display version information]'
+    '(: * -)'--pcre2-version'[print the version of PCRE2 used by ripgrep, if available]'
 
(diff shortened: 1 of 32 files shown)


complete/_rg
```diff
@@ -270,6 +270,8 @@ _rg() {
     '*--ignore-file=[specify additional ignore file]:ignore file:_files'
     '(-v --invert-match)'{-v,--invert-match}'[invert matching]'
     '(-M --max-columns)'{-M+,--max-columns=}'[specify max length of lines to print]:number of bytes'
+    '--max-column-preview[show preview for long lines when limit is enabled]'
```
okdana: These two should be made exclusive with each other, probably by breaking them out into a group with `+ '(max-column-preview)'` or similar

Also, i would suggest the description `show preview for long lines (with -M)` — that matches the wording for similar options, and might be a little clearer

complete/_rg
```diff
@@ -182,6 +182,7 @@ _rg() {
     + '(pcre2)' # PCRE2 options
     {-P,--pcre2}'[enable matching with PCRE2]'
     $no'(pcre2-unicode)--no-pcre2[disable matching with PCRE2]'
+    '--pcre2-version[print the version of PCRE2 used by ripgrep, if available]'
```
okdana: If i'm understanding correctly (that it has basically the same usage semantics as `--version`), this belongs under the `+ '(exclusive)'` group

complete/_rg
```diff
@@ -44,6 +44,10 @@ _rg() {
     '(: * -)'{-h,--help}'[display help information]'
     '(: * -)'{-V,--version}'[display version information]'
 
+    + '(hybrid)' # hybrid regex options
+    '--auto-hybrid-regex[dynamically use PCRE2 if necessary]'
```
okdana: Kind of a nit-pick, but describing the affected functionality differently between the enable and disable options feels a little odd (and inconsistent with other options)

okdana:
`--max-column-preview` works really nicely, thank you for going to the trouble of implementing it (and the other stuff too, obv)

BurntSushi:
@okdana Thanks very much for your review. :-) I think everything you pointed out should be fixed! And thanks for testing the preview functionality. I'm near certain that there will be bugs there because correctly testing the color spans was more work than I wanted to do.

BurntSushi:
@okdana If all goes well with the max column preview, I'm contemplating turning it on by default in ripgrep 12. You were completely right about it being the "expected" behavior here. (I've put it in my .ripgreprc.)

okdana:
Oh, actually, on reflection, do you think it should be `--max-columns-preview` (with an `s`) instead? That was the name you originally suggested in #1078, and it seems to make sense since it enables a `preview` for `--max-columns`. I'm sure you put some more thought into the name, but i thought i'd bring it up now, just in case, before it's completely set in stone

BurntSushi:
I don't think I thought about it too much. I just felt like `--max-column-preview` sounded better in my head. Do you like `--max-columns-preview` better?

okdana:
For what it's worth, yeah, i think so

BurntSushi:
@okdana All righty, SGTM. Done!

PR #1287, opened 2019-05-28
Point regex doc link to the version ripgrep uses

skierpage:
Cargo.toml says regex 1.0.5, and its doc is different, e.g. adds "symmetric differences" under https://docs.rs/regex/1.0.5/regex/#character-classes

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -110,7 +110,7 @@ colors, you'll notice that `faster` will be highlighted instead of just the
 
 It is beyond the scope of this guide to provide a full tutorial on regular
 expressions, but ripgrep's specific syntax is documented here:
-https://docs.rs/regex/0.2.5/regex/#syntax
+https://docs.rs/regex/*/regex/#syntax
 
 
 ### Recursive search


GUIDE.md
```diff
@@ -110,7 +110,7 @@ colors, you'll notice that `faster` will be highlighted instead of just the
 
 It is beyond the scope of this guide to provide a full tutorial on regular
 expressions, but ripgrep's specific syntax is documented here:
-https://docs.rs/regex/0.2.5/regex/#syntax
+https://docs.rs/regex/1.0.5/regex/#syntax
```
BurntSushi: Nice catch. Could you please use `*` here instead of `1.0.5` though? In particular, the `Cargo.toml` merely specifies the minimal version of regex that ripgrep can compile with. The [actual version used is in `Cargo.lock`](https://github.com/BurntSushi/ripgrep/blob/d1e4d28f3090cd0ce6e2eb7931e6b897e9235c30/Cargo.lock#L766). I'm pretty good about keeping dependencies up to date, so I think just having the docs link to the latest version in the README is the most convenient choice here.

skierpage: Done, I think (I've never updated a GitHub pull request, thanks for your patience).

I also made a [pull request for docs.rs About ](https://github.com/rust-lang/docs.rs/pull/362) to mention *|latest|newest works in links, that was news to me :smiley:.

PR #1307, opened 2019-06-19
Enable testing on ARM, in CI

lilianmoraru:
A few issues here.
It seems to fail a test in CI because the QEMU version is old and it does not support ARM very well.

There are at least a few ways to fix this(that I can think of):
- There is no QEMU ppa - so, either it is created by someone(I personally don't know how) or have a precompiled QEMU archive be downloaded.
- Switch the ARM build to another CI, like Azure Pipelines and have it there run on a newer version of Linux?
- Disable some tests on ARM(cfg over some tests)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,10 @@ matrix:
           - docbook-xsl
           - xsltproc
           - libxml2-utils
+          - qemu-system-arm
+          - qemu-user
+          - qemu-user-static
+          - binfmt-support
     # Beta channel. We enable these to make sure there are no regressions in
     # Rust beta releases.
     - os: linux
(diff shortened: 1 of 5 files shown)


lilianmoraru:
[`cross`](https://github.com/rust-embedded/cross) also supports running tests in QEMU through a newer Dockerfile(QEMU from Ubuntu 18.04).
`ripgrep` should probably switch to that in order to support a lot more targets.

BurntSushi:
Thanks! I don't know when (or if) I'll get a chance to dig into this. Honestly, it is _not_ a goal of mine to support lots of targets. Every target added means a higher chance of something going wrong in the release process that I have to debug, and the release process is already laborious enough. If people really want ARM binaries, then they'll need to setup their own infrastructure for building them.

BurntSushi:
I'm unfortunately going to close this. I just don't have the bandwidth to maintain releases for more targets. Sorry. If you wanted to maintain this on your own in a separate repo, and it was kept up to date, I would be willing to link it from the README.

PR #1327, opened 2019-07-24
Support compiling `ignore` crate to wasm32

tiziano88:
Currently the crate assumes that exactly one of `cfg(windows)` or
`cfg(unix)` is true, but this is not actually the case, for instance
when compiling for `wasm32`.

Implement the missing functions so that the crate can compile on other
platforms, even though those functions will always return an error.

### DIFF
--- a/ignore/src/walk.rs
+++ b/ignore/src/walk.rs
@@ -379,7 +379,18 @@ impl DirEntryRaw {
         })
     }
 
-    #[cfg(not(unix))]
+    // Placeholder implementation to allow compiling on non-standard platforms (e.g. wasm32).
+    #[cfg(not(any(windows, unix)))]
+    fn from_entry_os(
+        depth: usize,
+        ent: &fs::DirEntry,
+        ty: fs::FileType,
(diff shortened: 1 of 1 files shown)


ignore/src/walk.rs
```diff
@@ -416,6 +427,17 @@ impl DirEntryRaw {
...
+
+    // Placeholder implementation to allow compiling on non-standard platforms (e.g. wasm32).
+    #[cfg(not(any(windows, unix)))]
+    fn from_path(
+        depth: usize,
+        pb: PathBuf,
+        link: bool,
+    ) -> Result<DirEntryRaw, Error> {
+        Err(Error::Io(io::Error::new(
+            io::ErrorKind::NotFound, "unsupported platform")))
```
BurntSushi: I don't think `NotFound` is the right error kind for this. Can you use `Other` instead?

tiziano88: Done.

BurntSushi:
This PR is on crates.io in `ignore 0.4.8`.

PR #1328, opened 2019-07-25
Update deps

tiziano88:
- Relax `crossbeam-channel` constraint slightly
- Run `cargo update`.

### DIFF
--- a/ignore/Cargo.toml
+++ b/ignore/Cargo.toml
@@ -18,7 +18,7 @@ name = "ignore"
 bench = false
 
 [dependencies]
-crossbeam-channel = "0.3.6"
+crossbeam-channel = "0.3"
 globset = { version = "0.4.3", path = "../globset" }
 lazy_static = "1.1"
 log = "0.4.5"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,25 +2,24 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.3"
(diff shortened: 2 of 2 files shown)


ignore/Cargo.toml
```diff
@@ -18,7 +18,7 @@ name = "ignore"
 bench = false
 
 [dependencies]
-crossbeam-channel = "0.3.6"
```
BurntSushi: Why are you relaxing this version constraint? Have you tested that `ignore` works with `crossbeam-channel` `0.3.0` through `0.3.5`?

tiziano88: The reason is that I have a patched (for unrelated reasons) version of `crossbeam-channel` `0.3.9` already in my deps tree, but `ignore` is still depending on `0.3.6.` and therefore not picking up my local patch.

To your question, no, I have not actually checked that it works with these versions, I am just assuming that minor version increments would be compatible. If you want to do this more carefully, feel free to close this PR and do it in your own time, I agree it makes sense to inspect dependencies for a binary like `rg` that gets installed by so many people.

Thanks, and keep up the good work!

BurntSushi: `0.3.6` is equivalent to `^0.3.6`, which is equivalent to `>= 0.3.6, < 0.4.0`, so there should be no problem with Cargo using `0.3.9` with `ignore` due to this specific constraint. Something else must be going on?

But yeah, all right, I'm going to close this. I'll see about doing a dependency update soonish.

Issue #1329, opened 2019-07-27
Feature request: Add EDN to type list


KingMob:
#### What version of ripgrep are you using?

ripgrep 11.0.1
-SIMD -AVX (compiled)
+SIMD -AVX (runtime)

#### How did you install ripgrep?

Homebrew

#### What operating system are you using ripgrep on?

macOS 10.14.6

#### Describe your feature request.

I would like to add extension support for searching files in the [Extensible Data Notation (EDN)](https://github.com/edn-format/edn) format. They end in `.edn`. It's essentially Clojure's version of JSON.

The only question I'm unclear on is whether it should get its own type-list entry (like JSON) or be grouped with the rest of the Clojure extensions. Given how strongly associated it is with Clojure, I'm leaning to adding it there.

BurntSushi:
Generally, you don't need to open issues for this. Just open a PR with what you think it should be. If you aren't sure, then ask some other folks in the Clojure ecosystem to weigh in.

PR #1330, opened 2019-07-27
Add .edn suffix to Clojure type

KingMob:
EDN is Clojure's JSON equivalent, and used heavily for Clojure configuration files these days.

In response to #1329


KingMob:
Well, I totally guessed wrong on that one. I asked around, and by a vote of 6-1, more people prefer to keep them separate. I'll update the PR.

BurntSushi:
@KingMob Sounds good. Do you perhaps have a link to that vote? If not that's okay, but if so it would be good to include it here for provenance sake.

KingMob:
Hmm, I can add some Slack links, but two are private, and the other is on an unpaid plan, so the links will eventually decay, I assume. Want screenshots?

This was nothing so formal as a website and a poll; I asked at the public Clojurians Slack channels #edn and #off-topic, my coworking group's Slack (we have a few Clojurists), and the internal Cognitect Slack.

While not a super-high turnout, people like Alex Miller (a lead dev of Clojure), Billy Meier (author of the clj Liberator web framework), and Michiel Borkent (author of jet, a JSON/EDN/Transit conversion tool) all weighed in, FWIW.

Here's a link to the discussion in the public Clojurians #off-topic channel: https://clojurians.slack.com/archives/C03RZGPG3/p1564419595224700

PR #1337, opened 2019-08-02
ignore: rework inter-thread messaging

zsugabubus:
Change the meaning of `Quit` message. Now it means terminate. The final
"dance" is unnecessary, because by the time quitting begins, no thread
will ever spawn a new `Work`.

Replace that heuristic spin-loop, with blocking receive.

Optimize use of atomic operations.

### DIFF
--- a/ignore/src/walk.rs
+++ b/ignore/src/walk.rs
@@ -7,10 +7,9 @@ use std::path::{Path, PathBuf};
 use std::sync::Arc;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::thread;
-use std::time::Duration;
 use std::vec;
 
-use channel;
+use channel::{self, TryRecvError};
 use same_file::Handle;
 use walkdir::{self, WalkDir};
 
@@ -1157,8 +1156,7 @@ impl WalkParallel {
             return;
         }
         // Create the workers and then wait for them to finish.
-        let num_waiting = Arc::new(AtomicUsize::new(0));
-        let num_quitting = Arc::new(AtomicUsize::new(0));
+        let num_running = Arc::new(AtomicUsize::new(threads));
         let quit_now = Arc::new(AtomicBool::new(false));
         let mut handles = vec![];
         for _ in 0..threads {
@@ -1167,11 +1165,7 @@ impl WalkParallel {
                 tx: tx.clone(),
                 rx: rx.clone(),
                 quit_now: quit_now.clone(),
-                is_waiting: false,
-                is_quitting: false,
-                num_waiting: num_waiting.clone(),
-                num_quitting: num_quitting.clone(),
-                threads: threads,
+                num_running: num_running.clone(),
                 max_depth: self.max_depth,
                 max_filesize: self.max_filesize,
                 follow_links: self.follow_links,
@@ -1201,7 +1195,7 @@ enum Message {
     /// Work items for entries that should be skipped or ignored should not
     /// be produced.
     Work(Work),
-    /// This instruction indicates that the worker should start quitting.
+    /// This instruction indicates that the worker should quit.
     Quit,
 }
 
@@ -1280,21 +1274,12 @@ struct Worker {
     tx: channel::Sender<Message>,
     /// The receive side of our mpmc queue.
     rx: channel::Receiver<Message>,
-    /// Whether all workers should quit at the next opportunity. Note that
-    /// this is distinct from quitting because of exhausting the contents of
-    /// a directory. Instead, this is used when the caller's callback indicates
-    /// that the iterator should quit immediately.
+    /// Whether all workers should terminate at the next opportunity. Note
+    /// that we need this because we don't want other `Work` to be done after
+    /// we quit. We wouldn't need this if have a priority channel.
     quit_now: Arc<AtomicBool>,
-    /// Whether this worker is waiting for more work.
-    is_waiting: bool,
-    /// Whether this worker has started to quit.
-    is_quitting: bool,
     /// The number of workers waiting for more work.
-    num_waiting: Arc<AtomicUsize>,
-    /// The number of workers waiting to quit.
-    num_quitting: Arc<AtomicUsize>,
-    /// The total number of workers.
-    threads: usize,
+    num_running: Arc<AtomicUsize>,
     /// The maximum depth of directories to descend. A value of `0` means no
     /// descension at all.
     max_depth: Option<usize>,
@@ -1315,20 +1300,19 @@ impl Worker {
     /// The worker will call the caller's callback for all entries that aren't
     /// skipped by the ignore matcher.
     fn run(mut self) {
-        while let Some(mut work) = self.get_work() {
+        'get_work: while let Some(mut work) = self.get_work() {
             // If the work is not a directory, then we can just execute the
             // caller's callback immediately and move on.
             if work.is_symlink() || !work.is_dir() {
                 if (self.f)(Ok(work.dent)).is_quit() {
                     self.quit_now();
-                    return;
                 }
                 continue;
             }
             if let Some(err) = work.add_parents() {
                 if (self.f)(Err(err)).is_quit() {
                     self.quit_now();
-                    return;
+                    continue;
                 }
             }
             let readdir = match work.read_dir() {
@@ -1336,7 +1320,6 @@ impl Worker {
                 Err(err) => {
                     if (self.f)(Err(err)).is_quit() {
                         self.quit_now();
-                        return;
                     }
                     continue;
                 }
@@ -1349,7 +1332,7 @@ impl Worker {
                         Err(err) => {
                             if (self.f)(Err(err)).is_quit() {
                                 self.quit_now();
-                                return;
+                                continue;
                             }
                             false
                         }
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks. This will take me a while to get to, because I'll need to carefully review this to make sure it's correct. It is very very easy to introduce subtle bugs into this code, and more importantly, those bugs can be hard to catch.

zsugabubus:
I tried to explain every subtle details in comments, to be sure it's correct.

ignore/src/walk.rs
```diff
@@ -1470,64 +1453,44 @@ impl Worker {
...
+                Err(TryRecvError::Empty) => {
+                    // If it was the last running thread, then no more work
+                    // can arrive, thus we can safely start quitting.
+                    // Otherwise, a thread may spawn new works to be done.
+                    if self.waiting() == 1 {
+                        // Every other thread is blocked at the next recv().
+
+                        // Send the initial quit message and quit.
+                        self.tx.send(Message::Quit).unwrap();
+                        return None;
```
BurntSushi: Nice. This is clever. I think I buy it.

ignore/src/walk.rs
```diff
@@ -1470,64 +1453,44 @@ impl Worker {
...
                     }
+
+                    // Wait for next `Work` or `Quit` message.
+                    value = self.rx.recv().map_err(|_| TryRecvError::Disconnected);
```
BurntSushi: Why the `map_err` here instead of an `unwrap`?

zsugabubus: A few lines below I wrote an `unreachable!()` assertion that also covers `try_recv()` and this way `recv()` errors. So you will get a meaningful error message when panicking (although it can never occur).

zsugabubus: Although that `uncreachable()!` statement below is unnecessary, and we could use `unwrap` everywhere, I just wanted to explicitly distinguish `Err(TryRecvError::Empty)` from `Err(TryRecvError::Disconnected)` cases. In the future maybe `Disconnected` brach will be the quit.

BurntSushi: Right, yeah, I saw that. But then it becomes harder to distinguish between "channel somehow got disconnected from a bug" and "this particular blocking `recv` failed." That is, forwarding the error like this strips information from the inevitable backtrace as far as I can tell. I'd rather see it panic closer to where the error occurred.

BurntSushi: Yeah I like the idea of distinguishing the error cases. That part is good. :)

zsugabubus:
I also expect some minimal(?) speedups, because that two expensive atomic operations gone before `return Some(work)`.

BurntSushi:
Keep in mind that atomic operations are themselves likely quite inexpensive compared with processing the work itself, so I'd be surprised to see any performance changes here. Other than perhaps latency being better with quitting, but it's hard to reason about that.

My longer term plans for this code is to either find a way to switch to `rayon`, or be more principled about work stealing and work distribution.

ignore/src/walk.rs
zsugabubus: > I'd rather see it panic closer to where the error occurred

I have no so much experience in Rust. It seems very reasonable.

On the other hand, what if you want to call `drop` on the channel intentionally and you want it to quit instead of panicking. Then it's better to forward the error instead of `unwrap`. And I think, knowing what `(try_)recv` panicked is secondary. The most important information is that it hase been disconnected.

But it's hard to reason about, which would be the better choice, because it cannot occur now.

BurntSushi: > And I think, knowing which (try_)recv panicked is secondary. The most important information is that it hase been disconnected.

If you use `recv().unwrap()` instead of `recv().map_err(...)`, then we get both. Where as with the `map_err`, you only know that the channel was disconnected.

> But it's hard to reason about, which would be the better choice, because it cannot occur now.

Right, _unless_ there is a bug. So basically, this is about choosing a failure mode that provides the most information and is thus easier to debug, if and when it does occur.

zsugabubus:
Thanks!

ignore/src/walk.rs
```diff
@@ -1459,64 +1442,44 @@ impl Worker {
...
+                        // Every other thread is blocked at the next recv().
+
+                        // Send the initial quit message and quit.
+                        self.tx.send(Message::Quit).unwrap();
+                        return None;
                     }
+
+                    // Wait for next `Work` or `Quit` message.
+                    value = Ok(self.rx.recv().expect("channel disconnected while worker is alive"));
+                    self.resume();
```
BurntSushi: @zsugabubus So I unfortunately had to revert this PR. While I wasn't able to reproduce any problems locally, I was observing transient CI failures that could only be explained by the parallel directory traversal accidentally dropping work. After staring at this patch for a bit, I think I convinced myself that the problem is a race between when `value` is populated from the blocking receive here, the call to `self.resume()` and the call to `self.waiting()` above.

In particular, I believe there is a possible interleaving where `self.waiting() == 1` is true but there is actually still more work to be done. Once `value = ...;` is unblocked, it is possible for another worker to think that it was the last one running before `self.resume()` is called. This in turn causes the `Quit` domino effect to kick in, which in turn is guaranteed to cause all other works to stop even if there is still outstanding work to be done.

I believe I had similar issues in my initial implementation of parallel traversal, and that was what provoked the introduction of the "waiting" and "quitting" dance. AIUI, you probably would need something similar here to fix the race, I think.

Using blocking semantics however is still probably better than the implementation I have now, which inserts sleeps in order to avoid burning the CPU. Nevertheless, I'm going to stick with the old implementation for now unless you want to take another crack at it.

zsugabubus: But new works can’t only origin from a (running) worker? (Did something change since or I had some fatal assumptions?)

If every other worker is blocked at that `recv`, they surely won’t add anything… and before the last one goes to sleep (blocking) can safely start killing its brothers (initiate quitting).

The thing you talk about could only occur only if something other than a worker starts pushing new works, no (*)? This is the only code segment that accesses `tx` from outside, but it surely finishes before workers start at all (it just puts on root directories):
https://github.com/BurntSushi/ripgrep/blob/72f5a25435f3dad1c69d16010b0aa796f2523f70/ignore/src/walk.rs#L1140-L1144

*: Or if there is a bug(?) in channels implementation and makes `try_recv` reporting no work (empty), even though there would we works; and then randomly wakes up a blocking `recv` instead.

zsugabubus: On the other hand, I took a short glimpse into the documentation of channels, and if everything is true, we could completely eliminate `resume()`s and `waiting()`s by [`is_empty()`](https://docs.rs/crossbeam-channel/0.4.0/crossbeam_channel/struct.Receiver.html#method.is_empty). So instead `self.waiting() == 1`, `self.rx.is_empty()` could be written simply.

BurntSushi: So first I want to say that since reverting it, all transient failures I was observing have completely stopped. (The nature of the transient failures was such that the parallel traverser was returning only a subset of file paths that it should be returning. So it was either skipping some file paths or it was stopping early.) So I think, at least from my perspective, there is clear evidence that there is a problem with this change.

The key to my thinking behind the problem is that there is an _unsynchronized_ race between the `self.waiting() == 1` check and `value = ...;` being unblocked and `self.resume()` being called. Between those statements, there exists a point in time in which there is work remaining in a worker, but where that worker is currently regarded as `waiting.` All that needs to happen for that to occur is this:

1. Consider that all workers, except for one, are `waiting`.
2. The last remaining worker finds one more job to do and sends it on the channel.
3. One of the previously `waiting` workers wakes up from the job that the last running worker sent, but `self.resume()` has not been called yet.
4. The last worker, from (2), calls `get_work` and sees that the channel has nothing on it, so it executes `self.waiting() == 1`. Since the worker in (3) hasn't called `self.resume()` yet, `self.waiting() == 1` evaluates to true.
5. This sets off a chain reaction that stops all workers, despite that fact that (3) got more work (which could itself spawn more work).

BurntSushi: We definitely cannot use `self.rx.is_empty()`. The fact that the channel is empty does not imply that the traversal has finished. It could simply be that a worker is running and is _about_ to produce more work.

zsugabubus: Cried and thought a bit and I thought of something. Really hope not too big bullshit: Decrement a counter before `tx.send`s and decrement it when we have completely finished; basically a counter for “works pending”.

BurntSushi: Possibly... I'd have to think more carefully about it and don't have the energy at the moment. My initial instinct is that it sounds like it might have the same problem though.

zsugabubus: Some thoughts to make it clearer:
- The last running worker processes the last work. Counter is 1. (1 pending.)
- It adds a new work (W), but before it do so, it increments the counter by one (2 pending); so it ensures that `try_recv` + quit testing won’t initiate a quitting–like before–because even if (W) has been finished, counter will be >=1. Counter will be >=1 until this newly added work gets done too. And now it doesn’t matter which blocking or non-blocking `recv` will be waken up and when, worker won’t exit only if it got an empty response and no pending works.
- Here, no pending work really mean that `rx.recv` will block indefinitely, so we don’t need to fear.

If implemented well, it sounds obvious to me in plain English: If there are no pending works, there are nothing to do. We can quit now.

It was a bit hard to explain because my limited English skills and because it’s a tautology.

BurntSushi: I think I get it! And then we can remove the `waiting`/`resume` stuff entirely. Great. I have it implemented in #1491.

PR #1345, opened 2019-08-07
minor update to manpage to reflect glob matching behavior change

LawAbidingCactus:
This is a minor update to ripgrep's manpage to reflect the changes in #762. The manpage's use of `!git/*` in glob patten matching examples, which is presumably intended to exclude all `git` directories from ripgrep's search, is incorrect. This might be a source of confusion for people configuring ripgrep who may not be familiar with `.gitignore` semantics.

### DIFF
--- a/doc/rg.1.txt.tpl
+++ b/doc/rg.1.txt.tpl
@@ -143,16 +143,16 @@ would behave identically to the following command
 
 same with using globs
 
-    --glob=!git
+    --glob=!.git
 
 or
 
     --glob
-    !git
+    !.git
 
 would behave identically to the following command
 
-    rg --glob '!git' foo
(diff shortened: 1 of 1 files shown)


doc/rg.1.txt.tpl
```diff
@@ -143,16 +143,16 @@ would behave identically to the following command
 
 same with using globs
 
-    --glob=!git/*
+    --glob=!git
```
BurntSushi: Ah interesting, yeah, this makes sense to me. We might as well go the whole way here and use `.git` instead of `git`. What do you think?

LawAbidingCactus: Seems fine! That's what I usually use to exclude `.git` dirs anyway.

PR #1359, opened 2019-08-29
Split sbt out of scala file type

MaT1g3R:
sbt is a build tool for Scala. It is often useful to search only in sbt
files and not Scala files (e.g. looking for dependencies)

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -247,7 +247,8 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("ruby", &["Gemfile", "*.gemspec", ".irbrc", "Rakefile", "*.rb"]),
     ("rust", &["*.rs"]),
     ("sass", &["*.sass", "*.scss"]),
-    ("scala", &["*.scala", "*.sbt"]),
+    ("sbt", &["*.sbt"]),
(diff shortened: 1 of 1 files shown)


MaT1g3R:
I think filter only sbt files is a very useful feature, since there can be a lot of these files in large Scala projects. How do you feel about adding `*.sbt` back to Scala file type, and just add the new `sbt` type?

BurntSushi:
@MaT1g3R I'm not quite sure how I feel about it. It is very easy to just write `-g '*.sbt'` (or define your own file type via an alias or `.ripgreprc`). ripgrep generally doesn't have file types for every possible build system within every programming language, and I'm not sure if I want to start that precedent.

MaT1g3R:
Since `gradle` is grouped under groovy, I think it's reasonable that `sbt` is its own file type, since `sbt` is more of a programming language than a config file.

Adding this change will introduce the file type to other programs like `fd`. `fd` doesn't seems to be configured via `.ripgreprc` (It would be weird if it did)

BurntSushi:
I'm going to close this because I haven't received any credible guidance on whether removing `sbt` from the Scala type is appropriate or not.

I'd probably be okay with having a separate `sbt` type that overlaps with `scala`.

PR #1365, opened 2019-09-03
Make the parallel walker visit untraversable directories before failing

### LINKED ISSUE #1346: WalkParallel does not show unaccessible directories.
If a directory is not accessible the dta produced by WalkParallel is different from the one created by WalkDir: the unaccessible directory is not listed.

How to reproduce:
Prepare the env  (I use Linux)
```bash
# create dir 
mkdir imnotlisted
# make it unaccessible
sudo chmod 700 imnotlisted
sudo chown root:root imnotlisted 
```

Using the small programm below the parallel version
```bash
./small_program parallel imnotlisted
```
outputs:
```
Err(WithDepth { depth: 0, err: WithPath { path: "iminvisible", err: Io(Os { code: 13, kind: PermissionDenied, message: "Permission denied" }) } })
```

but the sequential version
```bash
./small_program walkdir imnotlisted
```
outputs
```
Ok(DirEntry { dent: Walkdir(DirEntry("iminvisible")), err: None })
Err(WithPath { path: "iminvisible", err: Io(Custom { kind: PermissionDenied, error: Error { depth: 0, inner: Io { path: Some("iminvisible"), err: Os { code: 13, kind: PermissionDenied, message: "Permission denied" } } } }) })
```

Which is the desired behaviour? In ripgrep this is not a problem.

-----------------
the "small program below"

```rust
extern crate crossbeam_channel as channel; 
extern crate ignore; 
 
use std::env; 
use std::thread; 
 
use ignore::WalkBuilder; 
 
fn main() { 
    let mut path = env::args().nth(1).unwrap(); 
    let mut parallel = false; 
    let (tx, rx) = channel::bounded::<Result<ignore::DirEntry,ignore::Error>>(100); 
    if path == "parallel" { 
        path = env::args().nth(2).unwrap(); 
        parallel = true; 
    } else if path == "walkdir" { 
        path = env::args().nth(2).unwrap(); 
    } 
 
    let stdout_thread = thread::spawn(move || { 
        for dent in rx { 
            println!("{:?}",dent); 
        } 
    }); 
 
    if parallel { 
        let walker = WalkBuilder::new(path).threads(6).build_parallel(); 
        walker.run(|| { 
            let tx = tx.clone(); 
            Box::new(move |result| { 
                use ignore::WalkState::*; 
 
                tx.send(result).unwrap();
                Continue
            }) 
        }); 
    } else { 
        let walker = WalkBuilder::new(path).build(); 
        for result in walker { 
            tx.send(result).unwrap(); 
        } 
    } 
    drop(tx); 
    stdout_thread.join().unwrap(); 
} 
```
(edit: fixed small program...was all lowercase)

jakubadamw:
This PR fixes an inconsistency between the serial and the parallel
directory walkers around visiting a directory for which the user holds
insufficient permissions to descend into.

The serial walker does produce a successful entry for a directory that
it cannot descend into due to insufficient permissions. However, before
this change that has not been the case for the parallel walker, which
would produce an `Err` item not only when descending into a directory
that it cannot read from but also for the directory entry itself.

This change brings the behaviour of the parallel variant in line with
that of the serial one.

Fixes #1346.

### DIFF
--- a/ignore/src/walk.rs
+++ b/ignore/src/walk.rs
@@ -1325,16 +1325,7 @@ impl Worker {
                     return;
                 }
             }
-            let readdir = match work.read_dir() {
-                Ok(readdir) => readdir,
-                Err(err) => {
-                    if (self.f)(Err(err)).is_quit() {
-                        self.quit_now();
-                        return;
-                    }
-                    continue;
-                }
-            };
+
             let descend = if let Some(root_device) = work.root_device {
                 match is_same_file_system(root_device, work.dent.path()) {
                     Ok(true) => true,
@@ -1352,6 +1343,7 @@ impl Worker {
             };
 
             let depth = work.dent.depth();
+            let readdir = work.read_dir();
             match (self.f)(Ok(work.dent)) {
                 WalkState::Continue => {}
                 WalkState::Skip => continue,
@@ -1363,6 +1355,18 @@ impl Worker {
             if !descend {
                 continue;
             }
+
+            let readdir = match readdir {
+                Ok(readdir) => readdir,
+                Err(err) => {
+                    if (self.f)(Err(err)).is_quit() {
+                        self.quit_now();
+                        return;
+                    }
+                    continue;
+                }
+            };
+
             if self.max_depth.map_or(false, |max| depth >= max) {
                 continue;
             }
@@ -2070,5 +2074,43 @@ mod tests {
         let mut builder = WalkBuilder::new(td.path());
         builder.follow_links(true).same_file_system(true);
         assert_paths(td.path(), &builder, &["same_file", "same_file/alink"]);
+    }
+
+    #[cfg(target_os = "linux")]
+    fn chmod<P: AsRef<Path>>(path: P, mode: libc::mode_t) -> Result<(), std::io::Error> {
+        use std::os::unix::ffi::OsStringExt;
+        let bytes = path.as_ref().as_os_str().to_os_string().into_vec();
+        let cstring = std::ffi::CString::new(bytes).unwrap();
+        let ret = unsafe { libc::chmod(cstring.as_ptr(), mode) };
+        if ret < 0 {
+            Err(std::io::Error::last_os_error())
+        } else {
+            Ok(())
+        }
+    }
+
+    #[test]
+    #[cfg(target_os = "linux")]
+    fn no_read_permissions() {
+        struct ChmodOnDrop(std::path::PathBuf, libc::mode_t);
+
+        impl Drop for ChmodOnDrop {
+            fn drop(&mut self) {
+                chmod(&self.0, self.1).unwrap();
+            }
+        }
+
+        let td = tmpdir();
+        let dir_path = td.path().join("a");
+        mkdirp(&dir_path);
+        wfile_size(dir_path.join("foo"), 0);
+
+        {
+            chmod(&dir_path, 0).unwrap();
+            let _chmod_on_drop = ChmodOnDrop(dir_path, libc::S_IRUSR | libc::S_IWUSR | libc::S_IXUSR);
+
+            let builder = WalkBuilder::new(td.path());
+            assert_paths(td.path(), &builder, &["a"]);
+        }
     }
 }
--- a/ignore/src/walk.rs
+++ b/ignore/src/walk.rs
@@ -1342,8 +1342,14 @@ impl Worker {
                 true
             };
 
-            let depth = work.dent.depth();
+            // Try to read the directory first before we transfer ownership
+            // to the provided closure. Do not unwrap it immediately, though,
+            // as we may receive an `Err` value e.g. in the case when we do not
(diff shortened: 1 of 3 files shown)


jakubadamw:
I should probably note that the new test will always fail when running it as a root. Is that a problem?

ignore/src/walk.rs
```diff
@@ -1363,6 +1355,18 @@ impl Worker {
             if !descend {
                 continue;
             }
+
+            let readdir = match readdir {
```
BurntSushi: Could you maybe add a short comment about the ordering here? e.g., Why `readdir` is being handled here instead of above when `work.read_dir()` is called?

ignore/src/walk.rs
```diff
@@ -2071,4 +2075,42 @@ mod tests {
...
+        wfile_size(dir_path.join("foo"), 0);
+
+        {
+            chmod(&dir_path, 0).unwrap();
+            let _chmod_on_drop = ChmodOnDrop(dir_path, libc::S_IRUSR | libc::S_IWUSR | libc::S_IXUSR);
+
+            let builder = WalkBuilder::new(td.path());
+            assert_paths(td.path(), &builder, &["a"]);
+        }
+    }
```
BurntSushi: Hmm, yeah, tests should not fail depending on what user ran them. Maybe it would be good to check if the file is readable after chmodding it. If it is, then skip the test.

With that said, it would be nice if we could avoid bringing in `libc` and `unsafe` for this test. Is there some other way to test this? Nothing is coming to mind... Maybe using a pre-existing file that a normal user traditionally doesn't have access to? (And skipping the test if the file doesn't exist or is readable.)

BurntSushi: Please also make sure that lines are wrapped to 79 columns, inclusive.

jakubadamw: @BurntSushi, I added the proposed check, though, of course, it'd be better if the test harness allowed marking a test as skipped at runtime.

With regard to avoiding adding a dev-dependency on `libc` I don't have a good idea either. But now I made it a Linux-only dependency. Perhaps that's good enough?

If using `libc` is really an issue (even though it already is a regular dependency for some other ripgrep crates, though, in truth, not ignore itself), then I'll happily change this to try to walk /root and check if it is observing at least one valid entry.

BurntSushi: I probably care more about the extra `unsafe` issue to be honest. I'm mostly just trying to make sure things are as a tight as possible.

Just looking at my file system, maybe just trying look for `/etc/sudoers.d`? It's readable by `root`, but not by anyone else. So I think the logic might be something like this:

* Stat `/etc/sudoers.d` via `std::fs::metadata()`.
* If it errors, skip test.
* Read `/etc/sudoers.d`.
* If it succeeds, skip test.
* Try to traverse `/etc/sudoers.d` and run the test normally.

jakubadamw: @BurntSushi, alright, sounds like a plan! I pushed the proposed change. Without a doubt that made the test much shorter and simpler. Thanks for the review!

jakubadamw:
@BurntSushi, there's a button there for cancelling a workflow. Have you tried that? (You probably have and I'm asking a stupid question. 🙂)

BurntSushi:
> there's a button there for cancelling a workflow. Have you tried that? (You probably have and I'm asking a stupid question. slightly_smiling_face)

Hah, I wish. But yeah, that button does normally show up. In my case though, Actions seems to be stuck in some sort of initialization state where by that button doesn't show up. The jobs have been running for over 4 days too, which suggests timeout functionality doesn't work either.

I've sent mail to GitHub support but haven't heard back from them.

I've turned Travis and AppVeyor back on. Going to see if I can trigger it on this PR.

treere:
there is a subtle bug and a performance issue. give me some minutes and I will explain.

ignore/src/walk.rs
treere: I think that readdir shuold be handled after the next if (line 1376). just before the iteration on the folder content.

If the the maximum depth has been reached you should I risk to `self.quit_now()`?

ignore/src/walk.rs
tavianator: Maybe `/root` (or more generally `~root`) is more ubiquitous than `/etc/sudoers.d`?

ignore/src/walk.rs
```diff
@@ -1351,6 +1342,13 @@ impl Worker {
                 true
             };
 
+            // Try to read the directory first before we transfer ownership
+            // to the provided closure. Do not unwrap it immediately, though,
+            // as we may receive an `Err` value e.g. in the case when we do not
+            // have sufficient read permissions to list the directory.
+            // In that case we still want to provide the closure with a valid
+            // entry before passing the error value.
+            let readdir = work.read_dir();
```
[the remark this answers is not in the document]
BurntSushi: I'm not sure I follow your comment here. There is exactly one call to `read_dir` both before and after this PR.

PR #1372, opened 2019-09-11
Ensure intended builds run

jclem:
Since this workflow specifies `runs-on: ${{matrix.os}}`, the `os` parameter needs to be defined for every build run.

Even though `include` is defined just for `pinned-musl` and `stable`, the other `build`s will still run—`include` is just saying "when these parameters match existing jobs, include some additional ones".

I've also added `sudo` to your `apt-get` command here—when running steps directly on the VM (as opposed to Docker), you're not running as root, and so password-less sudo can be used where needed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
     branches:
     - master
   schedule:
-    cron: '00 01 * * *'
+    - cron: '00 01 * * *'
 jobs:
   test:
     name: test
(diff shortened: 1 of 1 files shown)


BurntSushi:
> Even though include is defined just for pinned-musl and stable, the other builds will still run—include is just saying "when these parameters match existing jobs, include some additional ones".

Ahhhh, I see now. I had commented out some of the builds in order to debug and build out the CI stuff incrementally. Thank you so much for the clarification and fix!

jclem:
You're welcome! I'm working on trying to get a fully passing workflow here—I think I need to install `zsh` as well.

BurntSushi:
@jclem Oh I wouldn't worry about a fully passing workflow too much. Just getting the CI running again is fantastic! There's quite a bit more work required to move ripgrep over to GA that I still have to do anyway. :-) Thank you though!

jclem:
Okay, sounds good! Obviously, feel free to just yank my changes here if that's easier than merging a PR.

PR #1390, opened 2019-09-26
Add --no-context-separator option

MoSal:
`--context-separator=""` still adds a new line separator, and changing
 that would be a breaking change.

 This new option has clear intent, and preserves backward compatibility.

### DIFF
--- a/complete/_rg
+++ b/complete/_rg
@@ -277,6 +277,7 @@ _rg() {
     ))'
     '*--colors=[specify color and style settings]: :->colorspec'
     '--context-separator=[specify string used to separate non-continuous context lines in output]:separator'
+    "--no-context-separator[Don't print context separators]"
     '--debug[show debug messages]'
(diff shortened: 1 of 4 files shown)


BurntSushi:
What is your use case for this?

If we can make `--context-separator ''` work, then I'd prefer that. Even if it's technically a breaking change, that's the kind of change I'm generally okay with. We can just introduce it in the ripgrep 12 release.

MoSal:
> What is your use case for this?

Let's take filtering M3U playlist files as an example. One can easily get the URL/file corresponding to the filtered playlist item with `-A1`. So basically you can `echo #EXTM3U` which is the header, and do the filtering, and you get a valid sub-playlist as output.

> If we can make --context-separator '' work, then I'd prefer that.

The issue with that solution is providing a way for users to get the old behavior back. Using `\n` as a separator would give two empty lines not one. If we special-case `\n`, then `\n` would print one new line and `\n\n` would print three!

If you still want to go that route, the patch is ready.

PR #1391, opened 2019-09-26
ignore/types: Add slim, slime, and skim template filetypes

paradox460:
[Slim](http://slim-lang.com/) is a popular templating language for ruby, and [Slime](https://github.com/slime-lang/slime) is its elixir counterpart. [Skim](https://github.com/appjudo/skim) is "rails javascript" variant of it.


BurntSushi:
Is it appropriate to group all of them under one file type? Do they have enough differences where it would make sense to give them each their own type?

paradox460:
Slime attempts to be identical to Slim, with a few minor elixir-specific things that wouldn't make sense in ruby.

Skim is designed to be pretty minimal, but still has identical syntax to main slim, just with features removed

PR #1393, opened 2019-09-29
Put ignore's parallel walk support behind a feature gate

nox:
We would like to use the ignore crate in [mozjs_sys](https://github.com/servo/mozjs) to tell cargo to rebuild
SpiderMonkey on filesystem changes in its own sources, but we don't
want to bring crossbeam-channel and whatnot in the build for such a trivial
operation.

### DIFF
--- a/ignore/Cargo.toml
+++ b/ignore/Cargo.toml
@@ -18,13 +18,13 @@ name = "ignore"
 bench = false
 
 [dependencies]
-crossbeam-channel = "0.3.6"
+crossbeam-channel = { version = "0.3.6", optional = true }
 globset = { version = "0.4.3", path = "../globset" }
 lazy_static = "1.1"
 log = "0.4.5"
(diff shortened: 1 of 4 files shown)


nox:
I don't understand the build failure.

BurntSushi:
Thanks @nox! I'm going to close this PR now for procedural reasons, but I'm not necessarily opposed to it:

1. Firstly, are you still interested in using `ignore`? Apologies for taking so long to respond to this, but I was on a bit of a hiatus.
2. Secondly, I'm a little confused about avoiding `crossbeam-channel` here. `crossbeam-channel` is one of the lighter weight dependencies of `ignore`, and itself does not bring in much else. `regex`, `globset` and friends are much heavier. So it seems a little weird to me to go through all this trouble just to cut out `crossbeam-channel`. Maybe you could say a bit more about this?

Thanks!

PR #1395, opened 2019-10-01
Feat: add ripgrep running in a container

mbologna:
Add the instructions to run `ripgrep` in a Docker container.


BurntSushi:
Thanks, but I'd rather keep the installation instructions limited to things that have a history and will fairly reliably be updated. Maybe check back in after a few releases and we can see about putting a link to your repo then.

PR #1396, opened 2019-10-01
Update bytecount to 0.6

Eijebong:


BurntSushi:
Thanks. I appreciate the PR. But I generally no longer accept PRs that bump dependency versions. Instead, I reserve that for myself so that I can spend the time reviewing the changes before upgrading.

PR #1397, opened 2019-10-02
walk: treat symbolic links to directories as directories

### LINKED ISSUE #1389: Combination of --no-ignore, --sort, and search path of a symbolic link to a directory causes panic

ripgrep 11.0.2
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)

`brew install ripgrep`

macOS Mojave 10.14.6 (18G95)

When I run the following command that searches a symbolic link, ripgrep panics (after displaying output):

```rg -n --hidden --no-ignore --color=always --sort path pattern ~/symbolic-link-to-directory```

Use the previous flags on a symbolic link to a directory.

```
$ rg --debug -n --hidden --no-ignore --color=always --sort path every baz
DEBUG|rg::config|src/config.rs:40: /Users/harrisonmccullough/.ripgreprc: arguments loaded from config file: ["--smart-case"]
DEBUG|rg::args|src/args.rs:560: final argv: ["rg", "--smart-case", "--debug", "-n", "--hidden", "--no-ignore", "--color=always", "--sort", "path", "every", "baz"]
DEBUG|grep_regex::literal|grep-regex/src/literal.rs:59: literal prefixes detected: Literals { lits: [Complete(EVERY), Complete(eVERY), Complete(EvERY), Complete(evERY), Complete(EVeRY), Complete(eVeRY), Complete(EveRY), Complete(eveRY), Complete(EVErY), Complete(eVErY), Complete(EvErY), Complete(evErY), Complete(EVerY), Complete(eVerY), Complete(EverY), Complete(everY), Complete(EVERy), Complete(eVERy), Complete(EvERy), Complete(evERy), Complete(EVeRy), Complete(eVeRy), Complete(EveRy), Complete(eveRy), Complete(EVEry), Complete(eVEry), Complete(EvEry), Complete(evEry), Complete(EVery), Complete(eVery), Complete(Every), Complete(every)], limit_size: 250, limit_class: 10 }
DEBUG|globset|globset/src/lib.rs:435: built glob set; 0 literals, 0 basenames, 11 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 0 regexes
baz/file2.txt
1:Hello everyone
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', src/libcore/option.rs:347:21
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
   1: std::sys_common::backtrace::print
   2: std::panicking::default_hook::{{closure}}
   3: std::panicking::default_hook
   4: std::panicking::rust_panic_with_hook
   5: std::panicking::continue_panic_fmt
   6: rust_begin_unwind
   7: core::panicking::panic_fmt
   8: core::panicking::panic
   9: rg::try_main
  10: rg::main
  11: std::rt::lang_start::{{closure}}
  12: std::panicking::try::do_call
  13: __rust_maybe_catch_panic
  14: std::rt::lang_start_internal
  15: main
```

Not panic.

nikofil:
Due to how walkdir works if symlinks are not followed, symlinks to
directories are seen as simple files by ripgrep. This caused a panic
in some cases due to receiving a WalkEvent::Exit event without a
corresponding WalkEvent::Dir event.

This is fixed by looking at the metadata of the file in the case of a
symlink to determine if it's a directory.

Fixes #1389

### DIFF
--- a/ignore/src/walk.rs
+++ b/ignore/src/walk.rs
@@ -1033,7 +1033,18 @@ impl Iterator for WalkEventIter {
             None => None,
             Some(Err(err)) => Some(Err(err)),
             Some(Ok(dent)) => {
-                if dent.file_type().is_dir() {
+                let is_dir = match (dent.file_type().is_dir(), dent.file_type().is_symlink()) {
+                    (true, _) => true,
+                    (false, true) => {
+                        if let Ok(metadata) = dent.path().metadata() {
+                            metadata.file_type().is_dir()
+                        } else {
+                            false
+                        }
+                    },
+                    _ => false,
+                };
+                if is_dir {
                     self.depth += 1;
                     Some(Ok(WalkEvent::Dir(dent)))
                 } else {


BurntSushi:
Thanks. It's going to take me a bit to review this, as this adds an additional syscall for every symlink that is seen, as far as I can tell. We've had performance problems in the past (especially on Windows) when doing that.

nikofil:
I see, that could indeed be too much overhead.
The only case that this PR fixes should be the case reported (when the given path is a symlink to a directory). Otherwise when the option to follow symlinks isn't given the depth never changes so there are no `WalkEvent::Exit` events caused by symlinks:
https://github.com/BurntSushi/ripgrep/blob/8892bf648cfec111e6e7ddd9f30e932b0371db68/ignore/src/walk.rs#L1026-L1030
Therefore checking only the given path, instead of any symlink encountered, should also fix this case. I can try fixing it that way if you agree.

Issue #1394, opened 2019-09-30
ignore: WalkBuilder.add_ignore() patterns are not rooted in the path being walked


dragonmaus:
`Alternative title: No way to specify "root path" for WalkBuilder.add_ignore()`

<not-ripgrep-snip>

#### Describe your question, feature request, or bug.

When building an instance of `Walk` using `ignore::WalkBuilder`, the patterns in files added by `.add_ignore()` are matched against the current working directory of the process, instead of the directory being `Walk`ed.

#### If this is a bug, what are the steps to reproduce the behavior?

After downloading the files in [this gist](https://gist.github.com/dragonmaus/65aec98cdc163693c162ac2884bf0cc5), perform the following steps:

```sh
mkdir src
mv main.rs src
cargo build --release
target/release/test -I ignore .              # this is "Test 1"
cd ..
test/target/release/test -I test/ignore test # this is "Test 2"
```

#### If this is a bug, what is the actual behavior?

Test 1 produces the expected output:
```
.
./Cargo.lock
./Cargo.toml
./ignore
./src
./src/main.rs
```

Test 2, however, also includes the contents of the target directory (full output [here](https://gist.github.com/dragonmaus/c7fcc6b5e8f631cfd19060e9dad82c95)).

#### If this is a bug, what is the expected behavior?

Test 2 should produce the following output:
```
test
test/Cargo.lock
test/Cargo.toml
test/ignore
test/src
test/src/main.rs
```

Changing the pattern in the file `ignore` to `/test/target/` produces the desired results, but is not what I would expect (or want) to have to do.

tommilligan:
I'd like to take this issue - I've reproduced it and should have a fix this week.

Minimal test case:

```diff
diff --git a/ignore/src/walk.rs b/ignore/src/walk.rs
index e00b29a..f765b7f 100644
--- a/ignore/src/walk.rs
+++ b/ignore/src/walk.rs
@@ -1919,6 +1919,22 @@ mod tests {
         );
     }
 
+    #[test]
+    fn explicit_ignore_absolute_path() {
+        let td = tmpdir();
+        let igpath = td.path().join(".not-an-ignore");
+        mkdirp(td.path().join("a"));
+        wfile(&igpath, "/a");
+        wfile(td.path().join("foo"), "");
+        wfile(td.path().join("a/foo"), "");
+        wfile(td.path().join("bar"), "");
+        wfile(td.path().join("a/bar"), "");
+
[...]

PR #1399, opened 2019-10-07
Walkbuilder add ignore root

tommilligan:
See #1394 for full description.

> When building an instance of Walk using ignore::WalkBuilder, the patterns in files added by .add_ignore() are matched against the current working directory of the process, instead of the root of the tree being Walked.

This PR changes the behaviour or `WalkBuilder.add_ignore()`. **This is a breaking change.** Absolute file paths are now resolved based on the directory the `WalkBuilder` was initialised with.

Example directory structure:
```log
// /home
//   /subdir
//     - ignorefile
//     - README.md
```
Contents of `/home/subdir/ignorefile`:
```log
/README.md
```

The below code will now always ignore `/home/subdir/README.md` rather than `./README.md` of wherever the executing process is.

```rust
// psuedococde
let mut builder = WalkBuilder::new("/home/subdir");
builder.add_ignore("/home/subdir/ignorefile);
```

### DIFF
--- a/ignore/src/walk.rs
+++ b/ignore/src/walk.rs
@@ -1920,6 +1920,22 @@ mod tests {
     }
 
     #[test]
+    fn explicit_ignore_absolute_path() {
+        let td = tmpdir();
+        let igpath = td.path().join(".not-an-ignore");
+        mkdirp(td.path().join("a"));
+        wfile(&igpath, "/a");
+        wfile(td.path().join("foo"), "");
+        wfile(td.path().join("a/foo"), "");
+        wfile(td.path().join("bar"), "");
+        wfile(td.path().join("a/bar"), "");
+
+        let mut builder = WalkBuilder::new(td.path());
+        assert!(builder.add_ignore(&igpath).is_none());
+        assert_paths(td.path(), &builder, &["foo", "bar"]);
+    }
+
+    #[test]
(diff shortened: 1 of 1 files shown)


tommilligan:
Test failure appears to be unrelated.

BurntSushi:
@tommilligan Thanks for investigating this! However, I don't think I can merge this PR in its current form, and I'm not quite sure what the fix _should_ be. As you point out, this is a breaking change. Unfortunately, this is not a _desirable_ breaking change. Namely, the current behavior is actually intended as far as I can tell. If you look at ripgrep's docs for the `--ignore-file` flag, then you'll see that all ignore files provided in this way (which use `WalkBuilder::add_ignore`, and the `--ignore-file` flag is why this method was added) are _supposed_ to be matched relative to the current working directory.

So you might say, "OK, that's fine, then we can add a new `add_ignore`-like method that matches relative to the root." While I would be okay with that, the problem here is that a `WalkBuilder` _doesn't actually have a single root_. You can add multiple additional paths via the `add` method. The ignore files processed by `add_ignore` would then be incorrectly matched on subsequent paths.

BurntSushi:
(I am going to close this PR for now for procedural reasons, but I will leave #1394 open for now I think.)

PR #1400, opened 2019-10-08
args: Do not attempt to build glob options when not specified.

### LINKED ISSUE #1291: doesn't work in a nonexistent dir

```
ripgrep 11.0.1
+SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

`cargo install ripgrep`

Ubuntu 19.04

When the CWD is a nonexistent directory, and ripgrep is invoked, it always fails even if target is a valid absolute path. GNU grep works in this case.

```
0 ➜ mkdir somepath
0 ➜ cd somepath/
0 ➜ rm -rf ../somepath/
0 ➜ rg something /etc/hostname 
No such file or directory (os error 2)
2 ➜ 
```

According to strace, ripgrep invokes getcwd() at some point and fails with it.

```
0 ➜ RIPGREP_CONFIG_PATH= rg --debug something /etc/hostname 
DEBUG|grep_regex::literal|/home/torkve/.cargo/registry/src/github.com-1ecc6299db9ec823/grep-regex-0.1.3/src/literal.rs:59: literal prefixes detected: Literals { lits: [Complete(something)], limit_size: 250, limit_class: 10 }
DEBUG|globset|/home/torkve/.cargo/registry/src/github.com-1ecc6299db9ec823/globset-0.4.3/src/lib.rs:435: built glob set; 0 literals, 0 basenames, 11 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 0 regexes
No such file or directory (os error 2)
```

I expect ripgrep to fail only in cases when any relative paths are used, while the only-absolute-paths command should be successfully executed.

JesseH-:
Closes #1291 

This is a first shot at addressing #1291, though I think some additional input is needed.

After reviewing and manually testing, I found that the source of the errors being raised was due to unconditionally building an `OverrideBuilder` from the globbing options. As this relies on the current working directory, it understandably fails when the cwd does not exist.

The immediate error can be suppressed by checking that the globbing options are actually required before attempting to handle them. This will still raise the same error when any of these options are present without an existing cwd, though perhaps a clearer error message could also be emitted as part of this change in that case.

On the testing front, I wasn't able to suitably recreate the original issue in a regression test. In my experiments, I tried to set the `current_dir` of the `TestCommand` to something non-existent before running, but this seemed to result in a panic before actually entering any ripgrep code when we go to execute the command.

Manual testing showed that we were able to execute non-globbing ripgrep commands without failure from a non-existent cwd with this change applied.

### DIFF
--- a/src/args.rs
+++ b/src/args.rs
@@ -1267,6 +1267,13 @@ impl ArgMatches {
 
     /// Builds the set of glob overrides from the command line flags.
     fn overrides(&self) -> Result<Override> {
+        if !(self.is_present("glob-case-insensitive")
+             || self.is_present("glob")
+             || self.is_present("iglob"))
+        {
+            return Ok(Override::empty());
+        }
+
         let mut builder = OverrideBuilder::new(env::current_dir()?);
         // Make all globs case insensitive with --glob-case-insensitive.
         if self.is_present("glob-case-insensitive") {
@@ -1479,6 +1486,10 @@ impl ArgMatches {
     /// flag. If no --pre-globs are available, then this always returns an
     /// empty set of globs.
     fn preprocessor_globs(&self) -> Result<Override> {
+        if !self.is_present("pre-glob") {
+            return Ok(Override::empty());
+        }
+
         let mut builder = OverrideBuilder::new(env::current_dir()?);
         for glob in self.values_of_lossy_vec("pre-glob") {
             builder.add(&glob)?;


BurntSushi:
Thank you very much for this PR! I ended up going in a different direction. In particular:

* I created a new `current_dir` helper function. If `std::env::curent_dir` fails, then ripgrep will try to read `PWD`. If that fails, then ripgrep will print a more reasonable error message.
* Similarish to what you did, if ripgrep does not need to build the glob matchers, then it just creates an empty `Override` matcher.

While I didn't end up using your PR, your diagnosis of the situation was super helpful. Likely saved me quite a bit of time! So thank you!

Issue #1370, opened 2019-09-08
A flag to make --count return 0


cebaa:
#### What version of ripgrep are you using?

ripgrep 11.0.1 (rev 973de50c9e)

#### How did you install ripgrep?

Github binary release.

#### What operating system are you using ripgrep on?

Linux.

#### Describe your question, feature request, or bug.

With `grep -cH needle app.log*`, you get something like:

```
app.log.2019-08-24:0
app.log.2019-08-25:3
app.log.2019-08-26:5
app.log.2019-08-27:0
```

which is visually useful for cases such as "interesting things happened on 25th and 26th, but not on 24th and 27th".

The difference is in the "not on 24th and 27th" part - without `rg` outputting zero counts it's not possible to distinguish between:

- No matches in those files
- Those files do not even exist
- You messed up a pattern and it's not matching files it should

It might be good to consider a flag to include zero counts as well. IMO it would help with sanity double-checks.

Issue #1317, opened 2019-07-02
--count does not return 0


OJFord:
#### What version of ripgrep are you using?

ripgrep 11.0.1
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)

#### How did you install ripgrep?

brew

#### What operating system are you using ripgrep on?

Darwin 18.6.0

#### Describe your question, feature request, or bug.

```
$ rg --count 'this does not match anything' f
$
$ rg --count 'this does match something' f
1
$
```

For scripting purposes, I'd have preferred `0` than no output.

I thought I might have to accept that it's done in pursuit of compatibility, but:
```
$ grep --count 'this does not match anything' f
0
$
```

The exit code is 1 too, which is also a break from grep, and not desirable IMO, or maybe it's just because there really was a hard error, which is also what stopped any output?

BurntSushi:
Interesting. I'm not sure this was intentional. Or at least, I can't think of any particular reason why I did this. However, if you run ripgrep over a _directly_ with the `--count` flag, then it will not print file paths with zero matches. grep does, but that divergence was intentional on my part because otherwise the output contains a lot of noise.

> The exit code is 1 too, which is also a break from grep, and not desirable IMO

An exit code of `1` for a non-match is correct and consistent with grep.

OJFord:
> An exit code of 1 for a non-match is correct and consistent with grep.

Not when `--count` is supplied, it exits `0` and with stdout '0'.

cebaa:
@BurntSushi Roger that, opened https://github.com/BurntSushi/ripgrep/issues/1370

Backgammon-:
If I need to write shell arithmetic involving a count which can be zero (not that it matters, but I am counting connected external monitors), then the fact that ripgrep gives nothing instead of `0` is unnecessarily inconvenient and inconsistent, both with grep and with itself (imagine if `wc -l` printed nothing when run on an empty file...).

To preserve your intent re: noise: rather than diverging from grep by making zero counts print nothing, I think it would be better to give `0` when the count is actually zero, make `--count` over a directory omit empty counts by default, and add a flag to not omit empty counts when running over directories. This is slightly different from @cebaa's request, which just asks for the flag.

BurntSushi:
[...]

PR #1405, opened 2019-10-17
Add --include-zero option

cstyles:
This PR addresses the suggestion in https://github.com/BurntSushi/ripgrep/issues/1370.

If `--count` or `--count-matches` is enabled, ripgrep currently only reports the number of matches if it's non-zero. [This is desired behavior](https://github.com/BurntSushi/ripgrep/issues/1317#issuecomment-507711847) because otherwise it can generate a lot of noise but sometimes users may want to print zeroes anyway. There's actually already a setting for outputting a zero if there were no matches but there's no way of turning it on from the command line interface. This PR adds the `--include-zero` option to fix that.

### DIFF
--- a/grep-printer/src/summary.rs
+++ b/grep-printer/src/summary.rs
@@ -25,7 +25,7 @@ struct Config {
     stats: bool,
     path: bool,
     max_matches: Option<u64>,
-    exclude_zero: bool,
+    include_zero: bool,
     separator_field: Arc<Vec<u8>>,
     separator_path: Option<u8>,
     path_terminator: Option<u8>,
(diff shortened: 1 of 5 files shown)


grep-printer/src/summary.rs
```diff
@@ -261,13 +261,13 @@ impl SummaryBuilder {
...
+    /// Include count-related summary results with no matches.
     ///
     /// When enabled and the mode is either `Count` or `CountMatches`, then
-    /// results are not printed if no matches were found. Otherwise, every
-    /// search prints a result with a possibly `0` number of matches.
-    pub fn exclude_zero(&mut self, yes: bool) -> &mut SummaryBuilder {
-        self.config.exclude_zero = yes;
+    /// results are printed even if no matches were found. Otherwise, every
+    /// search only prints a result if there were matches.
+    pub fn include_zero(&mut self, yes: bool) -> &mut SummaryBuilder {
```
BurntSushi: I'm not following why this change was made. It's a breaking API change and seems purely cosmetic.

PR #1408, opened 2019-10-25
Scarf install instructions

aviaviavi:
This change adds install instructions for the [Scarf package manager](https://scarf.sh).

I'd be happy to transfer package ownership, or add co-maintainers. Otherwise, I'll maintain the Scarf package myself, so no action necessarily required on anyone else's part.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,12 @@ $ nix-env --install ripgrep
 $ # (Or using the attribute name, which is also ripgrep.)
 ```
 
+If you're a **Scarf** user you can install ripgrep from the [official repo](https://scarf.sh/package/scarf/ripgrep):
+
+```
+scarf install ripgrep
+```
+
 If you're a **Debian** user (or a user of a Debian derivative like **Ubuntu**),
 then ripgrep can be installed using a binary `.deb` file provided in each
 [ripgrep release](https://github.com/BurntSushi/ripgrep/releases).


BurntSushi:
I hadn't heard of Scarf before, so I perused its homepage. I appreciate what you're trying to do in terms of funding FOSS, but I'm not sure I'm willing to get behind the data collection you're doing by default. I'm going to close this PR for now, but here are some possible avenues forward/ways I could be convinced:

* Is the data collection that Scarf is doing standard among other package managers? (Especially ones that are already in my README.) I don't think it is, but I confess that I am not intimately familiar with every package manager in my README.
* A _possible_ middle ground would be to include Scarf instructions in the README but plaster a giant warning that installation via Scarf will by default send data back to the Scarf package maintainers.

The reason why I am being a bit of a hardass about this is that I am very very deeply opposed to _ever_ adding any kind of telemetry to ripgrep, opt-in or otherwise. It would potentially be a huge boon to discovering usability problems, but the costs in trust and risks due to the networking code would be absolutely enormous.

Either way, if you want to pursue this, I'd recommend opening a new issue to discuss this. At that point, hopefully others can weigh in on this issue.

aviaviavi:
Thanks for the thoughtful response here @BurntSushi. 

With regard to telemetry in other package managers: No other managers I'm aware of go as far as to collect usage telemetry, but they _do_ still have telemetry. For example, Homebrew sends analytics to Google Analytics by default. The difference is that they do it to understand their user behavior, find broken packages and show download counts. Scarf does all of this too but is mainly trying to help package authors leverage their data to charge companies for their work.

In any case, I understand that telemetry, in general, is what you want to avoid. Since Scarf fundamentally is all about helping OSS developers, we're looking into other features we can provide package authors and users that could ship without analytics altogether. I'll open another PR when we have something more compelling.

Thanks for your time, and thanks for this excellent tool!

Issue #1414, opened 2019-10-29
Add switch to pay attention to .gitignore files when not in a git repo


JohnC32:
Could you add a switch e.g. --force-gitignore which would force usage of .gitignore files even when there's no .git directory?

Give a large repo, it's often convenient to archive versions of it and not save the .git directory. Currently ripgrep 11.0.2 does not pay attention to .gitignore files if there's no .git directory. This makes it impossible to use rg on such repo's. Another use case is when using Perforce. Perforce pays attention to .gitignore files and does not have a .git directory. A work around is to create an empty .git directory, but in our environment this causes problems because git based tools fail when there's an empty .git directory.

Thanks for ripgrep - it's really powerful!

BurntSushi:
What about adding an empty _file_ called `.git` instead?

You could also symlink `.gitignore` to `.ignore`.

I'm really not too inclined to add even more flags specifically for changing when `.gitignore` is interpreted.

> Another use case is when using Perforce. Perforce pays attention to .gitignore files and does not have a .git directory.

What is the standard way to detect a Perforce repo? If Perforce specifically recognizes gitignore files, then we could add the Perforce check in addition to the `.git` check. Although, this will likely have unacceptable performance overhead.

JohnC32:
An empty .git directory or empty .git file does cause git based tools to fail. Both external tools and internal tools. For example, "touch .git; git status" yields "fatal: Invalid gitfile format: .git".

In Perforce, you set in their config P4IGNORE to .gitignore. You can also place P4IGNORE in the environment. Thus, detecting a Perforce repo is going to require something else. You could look for the existence of a .perforce file, but even this isn't required.

We have a large repo with many hundreds of .gitignore files and developed on multiple platforms. Symlink'ing won't work because of Windows. Adding rules to generate .ignore from .gitignore files is challenging because of the variety of different build systems we use.

I assume you do the .git check once, could you update it check for another file, e.g. .ripgrep-force-gitignore?

JohnC32:
Out of curiosity, why does ripgrep require a .git directory to pay attention to the .gitignore files? In a quick scan of the code and doc, I couldn't find a reason.

BurntSushi:
See https://github.com/BurntSushi/ripgrep/issues/934 and https://github.com/BurntSushi/ripgrep/commit/e65ca21a6cebceb9ba79fcd164da24478cc01fb0

[...]

PR #1416, opened 2019-10-31
[RFC] ripgrep: expose custom ignore files via --ignore-file-name

akarle:
The ignore crate has the concept of "custom ignore" files. These files
take precedence over all others and can be used to declare arbitrary
file names as files which should be parsed for ignore patterns.

This commit adds the `--ignore-file-name` switch to `ripgrep`, exposing this
functionality to the user. For instance, if a user wanted to put their
ignores in many files called `.my_ignore` they could:

    rg --ignore-file-name .my_ignore

This can also be used as a hack to honor `.gitignore` files outside .git
repositories (#1414), which can be useful if a git repo has been copied
somewhere without the .git directory (i.e. downloaded as a zip from
GitHub) or for SCM systems like Perforce which allow `.gitignore` files
but are not valid git repositories. Note, however, that this tactic
gives `.gitignore` precedence over `.ignore` and `.rgignore`, so consider
using multiple flags:

    rg --ignore-file-name .ignore --ignore-file-name .gitignore

-------

Aside: this is discussed in #1414 .

@BurntSushi : I have another branch on my fork, [ignore_outside](https://github.com/akarle/ripgrep/commit/efbed3298e9a5c2383c91e084012e0cbbda9e2ff), that implements the other possible way we discussed of solving this (via a `--force-gitignore` flag, which I called `--ignore-outside-git`, but am happy to change to whatever we find appropriate).

I also came up with a **third** solution that would work for our use-case: creating a `--alt-git-root-file` switch which specifies a token that is determined to indicate the root of a git repo. This is the most intrusive implementation though (getting into the guts of the `ignore` crate), and I found my knowledge of asynchronous rust didn't cut it :)

I'm proposing we merge the `--ignore-file-name` switch, even though it is the most hacky solution to #1414 , because it is the least change to the code, and could have other benefits / be considered a feature. However, I'm happy to go with either the `--ignore-outside-git` or `--alt-git-repo-files` solutions if you deem them more appropriate for the project.

Rust is not my strongest language, so I appreciate any and all pointers / tips / corrections!

Thanks for your time and help!
Alex

### DIFF
--- a/complete/_rg
+++ b/complete/_rg
@@ -281,6 +281,7 @@ _rg() {
     '--dfa-size-limit=[specify upper size limit of generated DFA]:DFA size (bytes)'
     "(1 stats)--files[show each file that would be searched (but don't search)]"
     '*--ignore-file=[specify additional ignore file]:ignore file:_files'
+    '*--ignore-file-name=[specify custom ignore file]:ignore file name:_files'
     '(-v --invert-match)'{-v,--invert-match}'[invert matching]'
     '(-M --max-columns)'{-M+,--max-columns=}'[specify max length of lines to print]:number of bytes'
     '(-m --max-count)'{-m+,--max-count=}'[specify max number of matches per file]:number of matches'
--- a/src/app.rs
+++ b/src/app.rs
@@ -577,6 +577,7 @@ pub fn all_args_and_flags() -> Vec<RGArg> {
     flag_iglob(&mut args);
     flag_ignore_case(&mut args);
     flag_ignore_file(&mut args);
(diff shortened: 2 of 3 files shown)


BurntSushi:
Thanks for this PR! The idea was interesting, but I think a more explicit `--no-require-git` flag is probably better. I've added this flag in #1486, which will be in the next release.

akarle:
@BurntSushi -- thank you so much for taking the time to revisit this!

I agree, the `--no-require-git` flag is more explicit (and less likely to cause confusion). Thank you for implementing it! Looking forward to using it in the next release.

PR #1427, opened 2019-11-12
update crossbeam-channel to 0.4.0

### LINKED ISSUE #1423: ignore: updating crossbeam-channel dep to 0.4
Just wanted to see if there were any plans to update the crossbeam-channel dependency to v0.4.0 in `ignore`, and if so, if there's a ballpark timeline?

There was an issue with crossbeam-utils v0.6.6 (which crossbeam-channel depends on) that has subsequently been fixed on the crossbeam side and recently released (v0.7.0)
https://github.com/rust-lang/rust/issues/65424
https://github.com/crossbeam-rs/crossbeam/issues/435

Ignore is pinned to crossbeam-channel v0.3.6, and that version of crossbeam-channel has a ^0.6.x dependency on crossbeam-utils, preventing the ability to use the fixed version (v0.7.0) of crossbeam-utils

https://github.com/BurntSushi/ripgrep/blob/8892bf648cfec111e6e7ddd9f30e932b0371db68/ignore/Cargo.toml#L21

Folks have been able to work around that issue (for example by pinning utils to v0.6.5 in the lockfile) so it's not a terribly urgent item, but it would be nice to be able to avoid this.

ctaggart:
fix #1423

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,9 +13,14 @@ name = "atty"
 version = "0.2.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 dependencies = [
- "libc 0.2.62 (registry+https://github.com/rust-lang/crates.io-index)",
+ "libc 0.2.65 (registry+https://github.com/rust-lang/crates.io-index)",
  "winapi 0.3.8 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
+
+[[package]]
+name = "autocfg"
+version = "0.1.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


ctaggart:
I can use this branch to workaround the issue in my Cargo.toml like so:

```
[replace]
"ignore:0.4.10" = { git = "https://github.com/ctaggart/ripgrep", package = "ignore", branch = "crossbeam" }
```

https://doc.rust-lang.org/cargo/reference/manifest.html#the-replace-section

Cargo didn't like the dash, so I created another identical branch named `crossbeam` without it.

PR #1430, opened 2019-11-16
RFC feat(ignore): Allow post-processing at end-of-thread

### LINKED ISSUE #469: How to finalize walk in ignore::WalkParallel
Hi there,

I have a use case for `ignore::WalkParallel` where I want to accumulate results into a local data structure in each thread, and then at the end of the walk merge the local results into a single global data structure. I could implement this by wrapping a single global copy of the data structure in a lock or similar, but that just seems wasteful, since there isn't any need for the threads to interact until the very end. Keeping the partial results local is a classic optimization in parallel programming.

As best I can tell, I can pass a local data structure to each thread by building it in the `mkf` function and passing it into the closure I produce. However, I can't figure out any way to get the local data structures back out, in part because I don't see that the threads receive any notice that the walk is ending. The threads just keep getting `DirEntry`s until they don't, and at that point it's already too late to do anything.

Can you suggest any way to do what I want with `ignore::WalkParallel`? Or if this isn't possible right now, could you consider it as a feature request?

Thanks.

epage:
On top of the parallel-walk's closures, this provides a Visitor API.
- Clarifies the role of the two different closures.
- Allows implementing of `Drop` for post-processing at thread
  completion.

The closure API is maintained not just for compatibility but also
convinience for simple cases.

Fixes #469

### DIFF
--- a/ignore/src/walk.rs
+++ b/ignore/src/walk.rs
@@ -1068,6 +1068,54 @@ impl WalkState {
     }
 }
 
+/// Create per-thread `ParallelVisitor`s for `WalkParallel`.
+pub trait ParallelVisitorBuilder {
+    /// Create per-thread `ParallelVisitor`s for `WalkParallel`.
+    fn build(&mut self) -> Box<dyn ParallelVisitor>;
+}
+
+/// Receives files and directories for the current thread/
+///
+/// - Setup can be implemented in `ParallelVisitorBuilder::build`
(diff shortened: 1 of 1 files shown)


epage:
I labeled this "RFC" because I have an open question on the API over at the original issue, https://github.com/BurntSushi/ripgrep/issues/469#issuecomment-546652231 .

PR #1444, opened 2019-12-10
Fix the current GitHub Actions workflow

nickelc:
zsh auto completions test was missing the `TARGET` env var.
`cross` is required for the musl build

see: https://github.com/nickelc/ripgrep/runs/341954437


BurntSushi:
Thanks for attempting this. GA was in an intermediate state that I never quite finished. I switched back to Travis/AppVeyor temporarily, but I'm hoping to take another crack at this soon.

PR #1446, opened 2019-12-11
ignore: use git commondir for sourcing .git/info/exclude

### LINKED ISSUE #1445: rg does not ignore files listed in .git/info/exclude in git linked worktrees

ripgrep 11.0.2 (same in master I think)

https://www.archlinux.org/packages/community/x86_64/ripgrep/

Arch Linux

If I exclude a file privately in `.git/info/exclude`, ripgrep ignores that file as expected - but only when searching inside the main worktree. In linked worktrees, the file is not ignored.

```sh
#!/bin/sh
rm -rf repro repro-worktree

git init repro
cd repro
touch committed
git add committed
git commit -m 'initial commit'
echo ignored > .git/info/exclude
git worktree add ../repro-worktree
cd ../repro-worktree
touch ignored

echo
echo '$ rg --files'
rg --files
echo '$ git ls-files; git ls-files --exclude-standard --others'
git ls-files; git ls-files --exclude-standard --others
```

```
$ rg --files
ignored
committed
```

I think it should behave like `git ls-files; git ls-files --exclude-standard --others`:

```
$ rg --files
committed
```

krobelus:
Git looks for this file in GIT_COMMON_DIR, which is usually the same
as GIT_DIR (.git). However, when searching inside a linked worktree,
.git is usually a file that contains the path of the actual git dir,
which in turn contains a file "commondir" which references the directory
where info/exclude may reside, alongside other configuration shared across
all worktrees. This directory is usually the git dir of the main worktree.

Unlike git this does *not* read environment variables GIT_DIR and
GIT_COMMON_DIR, because it is not clear how to interpret them when
searching multiple repositories.

Fixes #1445

### DIFF
--- a/ignore/src/dir.rs
+++ b/ignore/src/dir.rs
@@ -15,6 +15,8 @@
 
 use std::collections::HashMap;
 use std::ffi::{OsStr, OsString};
+use std::fs::File;
+use std::io::{self, BufRead};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, RwLock};
 
@@ -226,6 +228,7 @@ impl Ignore {
         } else {
             let (m, err) = create_gitignore(
                 &dir,
+                &dir,
                 &self.0.custom_ignore_filenames,
                 self.0.opts.ignore_case_insensitive,
             );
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks! I'll be merging this with some minorish changes in #1486. The main changes were:

* Fix the style of the code. Namely, the lines were quite long. It's best to stick to the style of the surrounding code. I'll be switching to rustfmt soon, so this hopefully shouldn't be a problem too much longer.
* I added an integration level regression test.
* I slightly refactored the code such that it doesn't introduce an extra stat call for every directory. These can add up and wind up being surprisingly expensive. Since we are already checking for the existence of `.git` in each directory, I just reused the result of that stat call.

PR #1483, opened 2020-02-14
Add *.svelte to js types

zarkone:
filetype of https://svelte.dev/

### DIFF
--- a/ignore/src/types.rs
+++ b/ignore/src/types.rs
@@ -162,7 +162,7 @@ const DEFAULT_TYPES: &'static [(&'static str, &'static [&'static str])] = &[
     ("java", &["*.java", "*.jsp", "*.jspx", "*.properties"]),
     ("jinja", &["*.j2", "*.jinja", "*.jinja2"]),
     ("js", &[
-        "*.js", "*.jsx", "*.vue",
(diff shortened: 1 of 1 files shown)


BurntSushi:
Hmmm, I'm not so sure about this. I don't know anything about svelte, but looking at your link, svelte files look more like HTML files to me. In particular, the `js` type in ripgrep does _not_ contain HTML files, even though they can contain Javascript inside them.

Would others care to weigh in on this?

zarkone:
@BurntSushi well, svelte files looks pretty much the same as `*.vue` files, which mask is in `js` category

zarkone:
for me personally it is more convenient to have both `vue` and `svelte` in js, but it is arguable of course

BurntSushi:
@zarkone I think it'd be a useful exercise to evaluate this addition on its own. It could have been a mistake to add vue to js, I don't know. Are svelte (or vue) files _principally_ Javascript?

zarkone:
principally no, i think. This files contains html,css and js -- all mixed

BurntSushi:
All right, so if it's not principally Javascript, then I'm not quite sure I feel comfortable putting it under the `js` type.

I'd be happy to be overruled by this, but I think I'd want to see some consensus from relevant stakeholders.

PR #1491, opened 2020-02-18
ci: switch to GitHub Actions

BurntSushi:


tests/util.rs
```diff
@@ -171,11 +171,19 @@ impl Dir {
...
 
     /// Returns the path to the ripgrep executable.
-    pub fn bin(&self) -> PathBuf {
-        if cfg!(windows) {
+    pub fn bin(&self) -> process::Command {
+        let rg = if cfg!(windows) {
             self.root.join("../rg.exe")
         } else {
             self.root.join("../rg")
+        };
```
LPGhatguy: You can use [`std::env::consts::EXE_SUFFIX`](https://doc.rust-lang.org/stable/std/env/consts/constant.EXE_SUFFIX.html) here instead of branching on Windows if you want to save a `cfg`!

BurntSushi: Nice catch, thanks! Fixed!

PR #1492, opened 2020-02-19
globset: Implement serde::{Serialize, Deserialize} for Glob

### LINKED ISSUE #1455: globset: Implement serde Serialize/Deserialize
My project currently uses a wrapper type around globset's `GlobMatcher` in order to implement `Serialize` and `Deserialize`. The implementation [is pretty trivial](https://github.com/rojo-rbx/rojo/blob/c82bc8a3583a65279d7aa28cde4b53977cd78fc4/src/glob.rs#L38-L50)!

Does it make sense for the globset crate to implement `Serialize`/`Deserialize`? I'd like to have them on both the `Glob` and `GlobMatcher` types for myself, but it's possible other types make sense. This would probably want to be behind a feature flag.

If this sounds okay, I'll file a PR! I think this change is similar to the spirit of #1447, but that PR hasn't been reviewed yet as of time of this writing.

LPGhatguy:
Closes #1455.

This PR adds an optional dependency to serde, which implies a matching feature named `serde`. When enabled, `Serialize` and `Deserialize` are implemented for `Glob`, using its string form.

Implementing serde traits for `GlobSet` might also be a good idea, but looked like it might be a much more invasive change.

I used serde_json for the included test. There might be other edge cases that should be tested.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,6 +135,8 @@ dependencies = [
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 1.3.4 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.45 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
 [[package]]
--- a/crates/globset/Cargo.toml
+++ b/crates/globset/Cargo.toml
@@ -24,10 +24,13 @@ bstr = { version = "0.2.0", default-features = false, features = ["std"] }
 fnv = "1.0.6"
 log = "0.4.5"
 regex = "1.1.5"
+serde = { version = "1.0.104", optional = true }
 
 [dev-dependencies]
 glob = "0.3.0"
 lazy_static = "1"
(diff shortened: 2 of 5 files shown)


Cargo.lock
```diff
@@ -438,7 +440,7 @@ dependencies = [
 
 [[package]]
 name = "serde_json"
-version = "1.0.45"
+version = "1.0.48"
```
BurntSushi: Is this version bump necessary? If not, please don't do it. I try to make sure to update all dependencies locally so that I can audit changes carefully.

crates/globset/Cargo.toml
```diff
@@ -24,10 +24,12 @@ bstr = { version = "0.2.0", default-features = false, features = ["std"] }
 fnv = "1.0.6"
 log = "0.4.5"
 regex = "1.1.5"
+serde = { version = "1.0.104", optional = true }
```
BurntSushi: Could you please create a `serde1` feature instead of tying the feature name to the name of a crate? This is useful if e.g., serde support for some reason requires an additional dependency.

Cargo.lock
LPGhatguy: Sorry, I added the latest `serde_json` because it was a new dev dependency for this crate, I can reduce its minimum version req and undo the lockfile changes.

LPGhatguy:
Thanks for taking a look at my change!

I believe I've made the requested changes.

Issue #1488, opened 2020-02-16
make it easier to maintain fork with additional regex engines


pierreN:
I needed a CLI tool to parse a massive amount of regexps (and improve my rust at the same time) so I made a crate implementing the `Matcher` trait : https://git.sr.ht/~pierrenn/grep-hyperscan

From my (sporadic) tests, its starts to be useful when having at least 1000 regexps to parse more than 10GB of data. I had 4.5k on 150GB soo...  Since the data comes from disk reads, using `hyperscan` basically limits your speed to your disk speed.

Plus there is a limit to the size of the compiled expressions in `ripgrep`, so using `hyperscan` allows to bypass that.

Ideally it would be cool if it can be integrated in `ripgrep`. I prefered to open this issue before doing a PR to talk about it and gauge interest. Details are below.

## Implementation

It's just an implementation of `find_at` since `hyperscan` doesn't support groups (`new_captures` is implemented using `NoCaptures`).

I thought of 3 possible ways to implement `find_at`:
- `hyperscan` has a `HS_FLAG_SINGLEMATCH` which would be great for `find_at`. However this is incompatible with the flag `HS_FLAG_SOM_LEFTMOST` which is required to get the `from` of the `Match`.
- you can force hyperscan to stop scanning a haystack after the first match by returning not 0 in the callback provided to `hyperscan`. However, this means a call to hyperstack each time we have a match. An (outdated...) implementation of this idea is available in the branch `ideas/single_match`
- everytime we get a new haystack sent to `find_at`, scan it in one go with `hyperscan`, remember the matches into a `VecDeque`, and consume the deque at each new call. I guessed first that when we return `Ok(None)` from `find_at`, the next haystack will be a new one. However (and this is weird?) sometimes `find_at` get sent a new haystack while the "current" is not termined. From testing it seems to only be the same haystack with EOL added at the end, or the right-most part of the original haystack. Thus, we also start a new `hyperscan` run when we see a new haystack length. Avoiding the successive calls to `hyperscan` allows according to my (sporadic...) benchmarks to speed up the overall match by 10-20%. Ideally, it would be great if we could require `ripgrep` to only send once each haystack (or the minimal amount of data), but no idea how to do that.

## Things to do for integration

I think the following tasks should be done for integration:
[...]

PR #1499, opened 2020-02-27
add engine option to easily switch between regexp engines

pierreN:
This is the commits corresponding to https://github.com/BurntSushi/ripgrep/issues/1488 

I tried to not change any default behavior of the options.

If I've done things correctly, the only thing that should change is when using the "auto-hybrid-regex" option without pcre2 feature enabled, an additional error message stating that pcre2 is not enabled should appear.

Thanks !

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -86,8 +86,6 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v1
-      with:
-        fetch-depth: 1
 
     - name: Install packages (Ubuntu)
       if: matrix.os == 'ubuntu-18.04'
@@ -185,8 +183,6 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v1
-      with:
-        fetch-depth: 1
     - name: Install Rust
       uses: actions-rs/toolchain@v1
       with:
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -508,6 +508,13 @@ impl RGArg {
         self
     }
 
+    /// Sets the default value of this argument when not specified at
+    /// runtime.
+    fn default_value(mut self, value: &'static str) -> RGArg {
+        self.claparg = self.claparg.default_value(value);
+        self
+    }
+
     /// Sets the default value of this argument if and only if the argument
     /// given is present.
     fn default_value_if(
@@ -566,6 +573,7 @@ pub fn all_args_and_flags() -> Vec<RGArg> {
     flag_debug(&mut args);
     flag_dfa_size_limit(&mut args);
     flag_encoding(&mut args);
+    flag_engine(&mut args);
     flag_file(&mut args);
     flag_files(&mut args);
     flag_files_with_matches(&mut args);
@@ -1182,6 +1190,36 @@ This flag can be disabled with --no-encoding.
     args.push(arg);
 
     let arg = RGArg::switch("no-encoding").hidden().overrides("encoding");
+    args.push(arg);
+}
+
+fn flag_engine(args: &mut Vec<RGArg>) {
+    const SHORT: &str = "Specify which regexp engine to use: default|pcre2|auto-hybrid.";
+    const LONG: &str = long!(
+        "\
+Specify which regular expression engine to use. When you choose a
+regex engine, it applies that choice for every regex provided to ripgrep (e.g.,
+via multiple -e/--regexp or -f/--file flags).
+
+Accepted values are `default|pcre2|auto-hybrid`.
+
+The default value is default, which is the fastest and should be good for most
+use-cases. PCRE2 engine is generally useful when you want to use features such
+as look-around backreferences. auto-hybrid will dynamically choose between
+supported regex engines depending on the features used in a pattern.
+
+Please have a look at the custom flags '--pcre2' and '--auto-hybrid-regexp'
+which explain their respective meaning in more depth. Those flags override
+any value provided to the `engine` option.
+
+Using '--pcre2' or '--auto-hybrid-regexp' overrides the value of this flag.
+"
+    );
+    let arg = RGArg::flag("engine", "ENGINE")
+        .help(SHORT)
+        .long_help(LONG)
+        .possible_values(&["default", "pcre2", "auto-hybrid"])
+        .default_value("default");
     args.push(arg);
 }
 
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -1194,7 +1194,8 @@ This flag can be disabled with --no-encoding.
 }
 
 fn flag_engine(args: &mut Vec<RGArg>) {
-    const SHORT: &str = "Specify which regexp engine to use: default|pcre2|auto-hybrid.";
+    const SHORT: &str =
+        "Specify which regexp engine to use: default|pcre2|auto-hybrid.";
     const LONG: &str = long!(
         "\
 Specify which regular expression engine to use. When you choose a
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -1209,11 +1209,10 @@ use-cases. PCRE2 engine is generally useful when you want to use features such
 as look-around backreferences. auto-hybrid will dynamically choose between
 supported regex engines depending on the features used in a pattern.
 
-Please have a look at the custom flags '--pcre2' and '--auto-hybrid-regexp'
-which explain their respective meaning in more depth. Those flags override
-any value provided to the `engine` option.
+Please have a look at the custom flags '--pcre2' and '--auto-hybrid-regex'
+which explain their respective meaning in more depth.
 
-Using '--pcre2' or '--auto-hybrid-regexp' overrides the value of this flag.
+Using '--pcre2' or '--auto-hybrid-regex' overrides the value of this flag.
 "
     );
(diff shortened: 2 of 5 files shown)


pierreN:
Tests should be fine now ? Unsure if I'm missing something or.. ?

crates/core/app.rs
```diff
@@ -1185,6 +1193,36 @@ This flag can be disabled with --no-encoding.
...
 
+fn flag_engine(args: &mut Vec<RGArg>) {
+    const SHORT: &str = "Specify which regexp engine to use: default|pcre2|auto-hybrid.";
+    const LONG: &str = long!(
+        "\
+Specify which regular expression engine to use. When you choose a
+regex engine, it applies that choice for every regex provided to ripgrep (e.g.,
+via multiple -e/--regexp or -f/--file flags).
+
+Accepted values are `default|pcre2|auto-hybrid`.
```
BurntSushi: Could this be written in prose? i.e.,

> Accepted values are 'default', 'pcre2' or 'auto-hybrid'.

crates/core/app.rs
```diff
@@ -1185,6 +1193,36 @@ This flag can be disabled with --no-encoding.
...
+
+The default value is default, which is the fastest and should be good for most
+use-cases. PCRE2 engine is generally useful when you want to use features such
+as look-around backreferences. auto-hybrid will dynamically choose between
```
BurntSushi: "as look-around or backreferences"

crates/core/app.rs
```diff
@@ -1185,6 +1193,36 @@ This flag can be disabled with --no-encoding.
...
+supported regex engines depending on the features used in a pattern.
+
+Please have a look at the custom flags '--pcre2' and '--auto-hybrid-regexp'
```
BurntSushi: Should be `'--auto-hybrid-regex'`.

crates/core/app.rs
```diff
@@ -1185,6 +1193,36 @@ This flag can be disabled with --no-encoding.
...
+which explain their respective meaning in more depth. Those flags override
+any value provided to the `engine` option.
+
+Using '--pcre2' or '--auto-hybrid-regexp' overrides the value of this flag.
```
BurntSushi: Also should be `'--auto-hybrid-regex'`.

crates/core/app.rs
```diff
@@ -1185,6 +1193,36 @@ This flag can be disabled with --no-encoding.
...
+"
+    );
+    let arg = RGArg::flag("engine", "ENGINE")
+        .help(SHORT)
+        .long_help(LONG)
+        .possible_values(&["default", "pcre2", "auto-hybrid"])
+        .default_value("default");
```
BurntSushi: Hmm, so I think this is not quite right. I think we want this to explicitly say that it overrides the `--pcre2` and `--auto-hybrid-regex` flags. (And vice versa for them.) Otherwise, something like `--pcre2 --engine default` would still use PCRE2 instead of what it should do: use the default engine.

BurntSushi:
Not sure what's going on with CI. Could you try deleting

https://github.com/BurntSushi/ripgrep/blob/ecec6147d117da638c933cf103a3ea14ce87102d/.github/workflows/ci.yml#L89-L90

and

https://github.com/BurntSushi/ripgrep/blob/ecec6147d117da638c933cf103a3ea14ce87102d/.github/workflows/ci.yml#L188-L189

And see if that helps.

crates/core/args.rs
```diff
@@ -576,60 +576,87 @@ impl ArgMatches {
...
-                Ok(matcher) => matcher,
-                Err(err) => {
-                    return Err(From::from(suggest_pcre2(err.to_string())));
-                }
-            };
-            Ok(PatternMatcher::RustRegex(matcher))
+            let engine = self.value_of_lossy("engine")
+                             .ok_or(clap::Error::with_description("Please \
+                                     provide a valid value to the engine \
+                                     option.", clap::ErrorKind::InvalidValue))?;
```
BurntSushi: Could you create errors in a way that is consistent with the surrounding code? e.g., `From::from(format!("..."))`.

crates/core/args.rs
```diff
@@ -576,60 +576,87 @@ impl ArgMatches {
...
+            #[cfg(feature = "pcre2")]
+            "pcre2" => {
+                let matcher = self.matcher_pcre2(patterns)?;
+                Ok(PatternMatcher::PCRE2(matcher))
+            },
+
+            #[cfg(not(feature = "pcre2"))]
+            "pcre2" => {
+                Err(From::from("PCRE2 is not available in \
+                                this build of ripgrep"))
```
BurntSushi: Looks like this will fit on one line?

crates/core/args.rs
```diff
@@ -648,15 +651,16 @@ impl ArgMatches {
...
                 )))
-            },
+            }
 
-            _ => Err(From::from(format!("Engine '{}' not found. Refer to the \
-                                         help to know supported values.",
-                                        engine)))
+            _ => Err(From::from(format!(
+                "Engine '{}' not found. Refer to the help to know supported \
+                 values.",
```
BurntSushi: It looks like this should fit on one line?

crates/core/args.rs
BurntSushi: Actually, now that I think about it, you should just be able to call `unwrap()` here since `engine` has a default value.

crates/core/args.rs
```diff
@@ -648,15 +651,16 @@ impl ArgMatches {
...
+                engine
+            ))),
```
BurntSushi: Could you please squash this commit back into the pertinent commits? I try to keep ripgrep's commit history clean, readable and linear, with tests passing on all commits. I don't always succeed, but I'd like to try. There isn't too much code here, so I think it's not too big of a pain to split these changes and squash them back into appropriate commits.

Although, I guess now that I say this, the commit hygiene in this PR is a bit off. It probably would have been better to start by refactoring how matchers are built and have that in one commit. And then a separate commit that actually adds the new `--engine` flag, which would include all relevant changes (including completions and docs and tests).

crates/core/app.rs
```diff
@@ -1209,11 +1209,10 @@ use-cases. PCRE2 engine is generally useful when you want to use features such
...
 supported regex engines depending on the features used in a pattern.
 
-Please have a look at the custom flags '--pcre2' and '--auto-hybrid-regexp'
-which explain their respective meaning in more depth. Those flags override
-any value provided to the `engine` option.
+Please have a look at the custom flags '--pcre2' and '--auto-hybrid-regex'
+which explain their respective meaning in more depth.
 
-Using '--pcre2' or '--auto-hybrid-regexp' overrides the value of this flag.
+Using '--pcre2' or '--auto-hybrid-regex' overrides the value of this flag.
```
BurntSushi: Changes like this should always be folded back into their corresponding commits.

.github/workflows/ci.yml
```diff
@@ -185,8 +183,6 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@v1
-      with:
-        fetch-depth: 1
```
BurntSushi: It looks like this fixed CI. Not sure why, but fine to leave this as is. (And this should be in a separate commit.)

crates/core/args.rs
pierreN: Ok yes thanks, I see what you meant by "commit hygiene", seeing modifications that way does make more sense.

Seeing all the changes needed to be done, is it OK if we just close/throw away this PR and make a new one ? Thanks.

pierreN: Maybe throw it away, do a first PR for the CI build, then do a second PR for the refactor+engine option ? 

I'm actually impressed by the speed of your answers ! I will wait for your confirmation before doing that though :)

BurntSushi: Up to you. I don't have a strong opinion. I'm happy to have you force push stuff here, but if a separate PR is easier for you, then go for it.

As for the CI fix, sure, a separate PR seems fine for that. Thanks. :-)

BurntSushi: > I'm actually impressed by the speed of your answers 

Thanks. I'm not always this fast though. You just happened to catch me at a good time. :)

pierreN:
Thanks, then yes I'll do this as a separate PR.

crates/core/app.rs
pierreN: OK, fair point. 

While testing now I was unsure to add the `engine` override for the `no-pcre2` and `no-auto-hybrid-regexp` since intuitively the option set `--engine=pcre2 --no-auto-hybrid-regex` should enable pcre2.

But since `--pcre2 --no-auto-hybrid-regex` doesn't enable `pcre2` neither, I guess it's just better to keep backward compatibility with default options and add overrides everywhere.

pierreN: OK, fair point. 

While testing now I was unsure to add the `engine` override for the `no-pcre2` and `no-auto-hybrid-regexp` since intuitively the option set `--engine=pcre2 --no-auto-hybrid-regex` should enable pcre2.

But since `--pcre2 --no-auto-hybrid-regex` doesn't enable `pcre2` neither, I guess it's just better to keep backward compatibility with default options and add overrides everywhere.

BurntSushi: Yeah good point. Defer to backward compatibility for now. I'll take another pass over this before the next release to make sure it all makes sense.

PR #1502, opened 2020-02-27
add engine option to easily switch between regexp engines

pierreN:
This is the commits corresponding to #1488

I tried to not change any default behavior of the options.

If I've done things correctly, the only thing that should change is when using the "auto-hybrid-regex" option without pcre2 feature enabled, an additional error message stating that pcre2 is not enabled should appear.

This is a follow up on : https://github.com/BurntSushi/ripgrep/pull/1499

Thanks !


pierreN:
Just noticed that `suggest_pcre` when `pcre2` is not enabled was not next to its enabled counterpart, so amended the first commit to reflect that.

pierreN:
Ping, any update on this by any chance ? Thanks.

PR #1505, opened 2020-02-29
update globset and ignore's cargo manifest urls

chippers:
The homepage and repo urls for `ignore` and `globset` manifest keys weren't updated when the project structure changed, leading links from https://docs.rs and https://crates.io to 404s.

Sorry about the PR on the new globset repo, I happened to find it first before this crates directory.


chippers:
These two were the only ones that were not pointing to ripgrep root, I'll update the other ones from the root to their directory

BurntSushi:
Ah that would be great, thanks!

PR #1510, opened 2020-03-07
ignore crate: adding a min_modified() method to skip old files

P-E-Meunier:
In some cases we want to skip older files when walking through a directory. This PR does just that.

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -633,7 +633,10 @@ impl WalkBuilder {
     }
 
     /// Whether to ignore files older than the specified time.
-    pub fn min_modified(&mut self, modified: Option<std::time::SystemTime>) -> &mut WalkBuilder {
+    pub fn min_modified(
+        &mut self,
+        modified: Option<std::time::SystemTime>,
+    ) -> &mut WalkBuilder {
         self.min_modified_time = modified;
         self
     }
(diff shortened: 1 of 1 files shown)


BurntSushi:
I'm not sure this is a good fit. It's adding a lot of complexity and seems fairly niche. Moreover, this seems like the first of a long line of additional options: `max_modified`, `min_created`, `max_created`, `min_accessed`, `max_accessed`, and so on. I'd rather we have a more cohesive strategy to address those.

In the future, I strongly recommend opening issues before filing PRs, especially when the PRs are non-trivial.

P-E-Meunier:
Alright, sorry if I didn't follow the normal contribution process.

BurntSushi:
I don't really have a normal documented contribution process. I'm just trying to save you time for next time. Trivial PRs are fine because if they don't end up being merged, then you didn't spend a lot of time on it. Meatier PRs are best to discuss first to make sure the feature is actually desirable.

P-E-Meunier:
Oh, alright. So, I really need this feature, which is why I wrote it. Do you think it could be merged with max_filesize, for instance by providing arbitrary filters?
Let's say:

```
builder.filter(|path| std::fs::metadata(m).unwrap().filesize() < 200_000)
```

A perhaps cleaner design would allow stacking up these filters (like std's iterators), but then the parallel walker might be harder to do, since you would need a trait to tell that two stacked filters are a valid walker.

PR #1554, opened 2020-04-18
crates/ignore: switch to depth first traversal

### LINKED ISSUE #1550: high peak memory usage with parallel breadth first directory traversal

ripgrep 12.0.0
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)

`cargo install ripgrep` on `rustc 1.42.0 (b8cedc004 2020-03-09)`

Linux, derived from Ubuntu 18.04

rg seems to leak memory when searching for an expression in latest versions of all crates from crates.io. The memory usage gradually increases and peaks at around 1Gb.

```bash
# get all crates from crates.io
git clone https://github.com/the-lean-crate/criner && cd criner && cargo run --release -- mine
# get latest version of every crate
for dir in ./criner.db/assets/*/*/*/; do echo -n "$dir"; ls "$dir" | grep '\.crate' | sort -Vr | head -n 1; done > ./latest-versions
mkdir crates
cd crates
# decompress the latest version of every crate
cat ../latest-versions | parallel ../decompress-one-crate.sh
```
where decompress-one-crate.sh is
```bash
#!/bin/sh
cat "$1" | tar --gunzip -xf -
```

Sample `rg` invocation exhibiting the behavior is:

```
rg --iglob '*.rs' -iF '#[allow('
```

Memory usage increases as the search progresses, reaching as high as 1Gb. It increases gradually and accumulates.

Memory usage should not accumulate

BurntSushi:
This replaces the use of channels in the parallel directory traversal
with a simple stack. The primary motivation for this change is to reduce
peak memory usage. In particular, when using a channel (which is a
queue), we wind up visiting files in a breadth first fashion. Using a
stack switches us to a depth first traversal. While there are no real
intrinsic differences, depth first traversal generally tends to use less
memory because directory trees are more commonly wide than they are
deep.

In particular, the queue/stack size itself is not the only concern. In
one recent case documented in #1550, a user wanted to search all Rust
crates. The directory structure was shallow but extremely wide, with a
single directory containing all crates. This in turn results is in
descending into each of those directories and building a gitignore
matcher for each (since most crates have `.gitignore` files) before ever
searching a single file. This means that ripgrep has all such matchers
in memory simultaneously, which winds up using quite a bit of memory.

In a depth first traversal, peak memory usage is much lower because
gitignore matches are built and discarded more quickly. In the case of
searching all crates, the peak memory usage decrease is dramatic. On my
system, it shrinks by an order magnitude, from almost 1GB to 50MB. The
decline in peak memory usage is consistent across other use cases as
well, but is typically more modest. For example, searching the Linux
repo has a 50% decrease in peak memory usage and searching the Chromium
repo has a 25% decrease in peak memory usage.

Search times generally remain unchanged, although some ad hoc benchmarks
that I typically run have gotten a bit slower. As far as I can tell,
this appears to be result of scheduling changes. Namely, the depth first
traversal seems to result in searching some very large files towards the
end of the search, which reduces the effectiveness of parallelism and
makes the overall search take longer. This seems to suggest that a stack
isn't optimal. It would instead perhaps be better to prioritize
searching larger files first, but it's not quite clear how to do this
without introducing more overhead (getting the file size for each file
requires a stat call).

Fixes #1550 

cc @bmalehorn who attempted a similar change many moons ago!

### DIFF
--- a/crates/ignore/Cargo.toml
+++ b/crates/ignore/Cargo.toml
@@ -18,7 +18,6 @@ name = "ignore"
 bench = false
 
 [dependencies]
-crossbeam-channel = "0.4.0"
 crossbeam-utils = "0.7.0"
 globset = { version = "0.4.3", path = "../globset" }
 lazy_static = "1.1"
@@ -31,6 +30,9 @@ walkdir = "2.2.7"
 
 [target.'cfg(windows)'.dependencies.winapi-util]
 version = "0.1.2"
+
+[dev-dependencies]
+crossbeam-channel = "0.4.0"
 
 [features]
 simd-accel = ["globset/simd-accel"]
--- a/crates/ignore/src/lib.rs
+++ b/crates/ignore/src/lib.rs
@@ -46,7 +46,6 @@ See the documentation for `WalkBuilder` for many other options.
 
 #![deny(missing_docs)]
 
-extern crate crossbeam_channel as channel;
 extern crate globset;
 #[macro_use]
 extern crate lazy_static;
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -5,10 +5,11 @@ use std::fs::{self, FileType, Metadata};
 use std::io;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
+use std::thread;
+use std::time::Duration;
 use std::vec;
 
-use channel::{self, TryRecvError};
 use same_file::Handle;
 use walkdir::{self, WalkDir};
 
@@ -364,7 +365,8 @@ impl DirEntryRaw {
         })
     }
 
-    // Placeholder implementation to allow compiling on non-standard platforms (e.g. wasm32).
+    // Placeholder implementation to allow compiling on non-standard platforms
+    // (e.g. wasm32).
     #[cfg(not(any(windows, unix)))]
     fn from_entry_os(
         depth: usize,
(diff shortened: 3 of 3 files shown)


sharkdp:
A few days ago, I released a new version of `fd` and `cargo update`d all dependencies without noticing (shame on me) that there was a new patch version of the `ignore` crate, which includes this change. The `ignore` crate is really the heart of the `fd` application, so I usually follow `ignore` releases very closely, but somehow I missed this.

We have a two open tickets in `fd` that concern high peak memory usage (https://github.com/sharkdp/fd/issues/378, https://github.com/sharkdp/fd/issues/471), so I'm really excited to see some work on this front (thanks!). I haven't checked if those issues would be resolved by this change, but it could very well be the case.

I'm a bit "concerned" about the change to a depth-first traversal though. Sure, if users of `rg` or `fd` perform a full search, it doesn't really matter (except for the output order, which is non-deterministic anyway - especially if the search is not limited to a single thread). But how about the case where someone is just waiting for "that one" result in a longer search? Wouldn't we expect the breadth first search to result in a better experience?

I found this PR after reading the report of a `fd` user who uses `fd` in combination with `fzf` to interactively search for files. With the previous breadth first search behavior, file system entries close to the start directory would show up very quickly. With the new depth first search, some of these entries appear at the very end of the (long running) search. I know that a lot of `ripgrep` users also use it in combination with `fzf`. Could this change also result in worse behavior for them? How about those users who perform a search for "that single keyword that only appears once somewhere in my home folder"? Isn't that file likely to be located at shallower depths?

Sorry for this long post - I'd be glad to hear your thoughts!

BurntSushi:
@sharkdp Sure, this change is invariably going to make some cases worse. Not really sure if there's much to do about it though. The peak memory usage of breadth first traversal was really not acceptable, and was also significantly slower in some cases.

tmccombs:
> The same user also reports significant performance regressions in the latest fd version that includes this change.

I wonder if this has to do with the CPU cache. Specifically, if with breadth-first traversal, the entries for a directory were more likely to be in a nearby cache-line, but with depth-first traversal, the directory data might be flushed from the cache and has to be read from main memory on each iteration.

Issue #1555, opened 2020-04-20
Ignore: Skip files on MacOS hidden by attribute


casey:
@Celeo is working on integrating `ignore` into a project, when before it was using `walkdir`, in order to benefit from `.gitignore` and `.ignore` processing. (Thanks for both of these crates!)

There's one piece of functionality that isn't easy to replicate with `ignore`, which is skipping files and directories on MacOS if they're hidden by attribute. I had written my own function to check if files were hidden, and then applied it with the `filter_entry` iterator combinator from `walkdir`.

It looks like `ignore` checks for the hidden attribute on Windows, but not MacOS, and it doesn't expose `filter_entry` from walkdir.

What do you think about adding support for checking the macos hidden attribute? I'd be happy to contribute this as a PR. Or possibly is there something in the API that I've missed that allows filtering entries?

Unfortunately, my implementation of checking the hidden attribute involves unsafe code, and I don't think there's a safe alternative. Here's [the code](https://github.com/casey/intermodal/blob/b67a2f1885c9445c08411457809f8893ebfa2045/src/platform.rs#L17) for reference.

BurntSushi:
There is currently no "filter entry" API in the `ignore` crate.

My understanding is that using a `.` prefix on macOS is fairly standard for indicating hidden files. I didn't even know of the existence of a separate attribute.

I'm not inclined to support this in `ignore` directly. Windows support for this exists principally because the `.` prefix convention isn't really used there, _and_ it's cheap. That is, it requires no additional stat calls, but it looks like macOS would. That's a pretty expensive proposition.

I think the right path forward here is probably to add some kind of filtering functionality to the walker, possibly by adding a new method on `WalkBuilder` (instead of how the `walkdir` API works). Then callers could add their own filtering like this if they wanted.

I'd be okay accepting a PR for that if it's a relatively simple change.

casey:
Whoops, I totally missed something, it looks like `st_flags` is [actually available](https://github.com/rust-lang/rust/blob/d8bdb3fdcbd88eb16e1a6669236122c41ed2aed3/src/libstd/os/macos/fs.rs#L66) via `MetadataExt` on MacOS.

[...]

PR #1557, opened 2020-04-21
crates/ignore: Allow filtering with predicate

casey:
Adds `WalkBuilder::filter_entry` that takes a predicate to be applied to
all entries. If the predicate returns `false` on a given entry, that
entry and all children will be skipped.

Fixes: #1555


casey:
You're most welcome, thanks for merging!

Issue #1542, opened 2020-04-05
make jemalloc an optional Cargo feature to make musl builds easier


moderation:
#### What version of ripgrep are you using?

non-musl version built from master

```
rg --version
ripgrep 12.0.1 (rev 49de7b119c)
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

#### How did you install ripgrep?

Compile from source 

#### What operating system are you using ripgrep on?

`Linux moderation-x1 5.3.0-40-generic #32-Ubuntu SMP Fri Jan 31 20:24:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`

```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=19.10
DISTRIB_CODENAME=eoan
DISTRIB_DESCRIPTION="Ubuntu 19.10"
```

#### Describe your question, feature request, or bug.

Trying to build a musl binary from source following instructions at https://github.com/BurntSushi/ripgrep#building

#### If this is a bug, what are the steps to reproduce the behavior?

`cargo build --release --verbose --target x86_64-unknown-linux-musl`

I _think_ have the required Rust targets an OS packages

```
rustup target list | rg installed
x86_64-unknown-linux-gnu (installed)
x86_64-unknown-linux-musl (installed)
```

```
dpkg --get-selections | rg musl
musl:amd64                                      install
musl-dev:amd64                                  install
musl-tools                                      install
```

```
rustc --version
rustc 1.42.0 (b8cedc004 2020-03-09)
```

#### If this is a bug, what is the actual behavior?

```
[...]

PR #1569, opened 2020-05-06
ci: puts jemalloc allocator behind a cargo feature flag

jonstites:
I have to say, thank you for making such incredibly nice, well-designed tools. I went through `ripgrep` and `xsv` a lot when I was first wrapping my head around Rust.

I saw an issue that I thought I could handle, so in an effort to be helpful, here it is.

This puts the jemalloc allocator behind a cargo feature as suggested here:
https://github.com/BurntSushi/ripgrep/issues/1542

The GitHub action for building releases appears to work:
https://github.com/jonstites/ripgrep/actions/runs/97156617

I did my best to make sensible choices, but I could be off the mark. Some thoughts:
- It might be confusing that there is a feature called "jemalloc" that doesn't actually use jemalloc unless on x86 musl. Cargo does not yet fully support platform-specific features. The feature could be called "jemalloc-musl-x86" or it could actually try to use "jemalloc" for any target. But jemallocator doesn't fully support all that many targets...
- This changes the default allocator for x86_64-unknown-linux-musl. This might lead to a big performance hit for someone. The feature could instead be for disabling jemalloc.
- I didn't touch the build-deb file, partly because I don't fully understand how .deb packaging works and partly because as best I can tell, jemalloc is disabled: https://salsa.debian.org/rust-team/debcargo-conf/-/blob/85f65b218cf86dbded7b66705617d3a5eb0b6a25/src/ripgrep/debian/patches/disable-jemallocator.diff

PS I am not someone who would take it personally if you requested big changes or decided that this issue isn't important enough for the added complexity. It's your project and your call on the code!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -153,6 +153,12 @@ jobs:
       if: matrix.target != ''
       run: ${{ env.CARGO }} test --verbose --all ${{ env.TARGET_FLAGS }}
 
+    - name: Run tests with jemalloc (Musl)
+      # We only use the jemalloc allocator when building with musl.
+      # The system allocator is good enough for other platforms.
+      if: matrix.os == 'nightly-musl'
+      run: ${{ env.CARGO }} test --verbose --all --features jemalloc ${{ env.TARGET_FLAGS }}
+
     - name: Test for existence of build artifacts (Windows)
       if: matrix.os == 'windows-2019'
       shell: bash
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -162,7 +162,13 @@ jobs:
         echo "::set-env name=RELEASE_VERSION::$release_version"
         echo "release version: $RELEASE_VERSION"
 
-    - name: Build release binary
+    - name: Build release binary (linux)
+      if: matrix.build == 'linux'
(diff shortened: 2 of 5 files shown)


BurntSushi:
Hmmm... Thanks so much for putting in the work for this! Actually seeing it done kind of makes me think it's not worth doing. At minimum, I would probably either want to rename the `jemalloc` feature to `jemalloc-x86_64-musl` _or_ cause the `jemalloc` feature to enable jemalloc regardless of the platform (even if it isn't supported), like you suggested. But this does indeed lead to a case where building musl by default will now result in a slower ripgrep. I'm not quite sure how to mitigate that unfortunately. I think I would rather not have jemalloc configurable as a feature. :-( If disabling it ends up really being required, then I think just patching ripgrep is probably the best bet and is easy to do.

So thanks again for this, but I'm going to close this out for now.

hboetes:
I just applied this patch to my git repo and it works like a charm. Finally can build static binaries. Please apply to master.

jameshilliard:
> But this does indeed lead to a case where building musl by default will now result in a slower ripgrep.

I think you can just set it as optional but enabled by default like [this](https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature) to fix that.

PR #1613, opened 2020-06-09
On Windows MSVC, statically link the C runtime

AustinWise:
Before this change, rg.exe depended on vcruntime140.dll, which does not exist on a fresh install of Windows. The missing DLL prevents rg.exe from starting.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,7 @@
+
+# On Windows MSVC, statically link the C runtime so that the resulting EXE does
+# not depend on the vcruntime DLL.
+[target.x86_64-pc-windows-msvc]
+rustflags = ["-C", "target-feature=+crt-static"]
+[target.i686-pc-windows-msvc]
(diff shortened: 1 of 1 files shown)


AustinWise:
There is presence for using this option. The rg.exe shipped with VS Code is [compiled with this option](https://github.com/microsoft/ripgrep-prebuilt/blob/677ffbed25585367af2d7642650f976496091a23/build/windows.yml#L35). rustc.exe is [compiled with this option](https://github.com/rust-lang/rust/pull/39837) too.

AustinWise:
The main downside to this change is it grows the binary size. For x64, it increases 2% from 4.14 MB to 4.24 MB.

wdscxsj:
I suppose 0.1 MB is negligible here... An immediately usable binary is much better than having to find and download the venerable Visual C++ Redistributable in emergency.

BurntSushi (approved):
I love this, thank you! I'll also update the README which IIRC mentions installing vcruntime or similar.

PR #1615, opened 2020-06-11
Add nightly flag to SIMD cargo build command.

catleeball:
A tiny doc update that adds the `+nightly` flag to the example command to build ripgrep with simd extensions. Thought it might be nice in case folks have the stable rustc set to default and forget to add the flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,7 +370,7 @@ If you have a Rust nightly compiler and a recent Intel CPU, then you can enable
 additional optional SIMD acceleration like so:
 
 ```
-RUSTFLAGS="-C target-cpu=native" cargo build --release --features 'simd-accel'
+RUSTFLAGS="-C target-cpu=native" cargo +nightly build --release --features 'simd-accel'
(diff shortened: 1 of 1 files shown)


BurntSushi:
I'm not sure I want to do this. I think it assumes you're using rustup. Probably not a terrible assumption, but probably not universally true either. The README does mention that night is required.

To be honest, I would just assume remove this section entirely. The additional optimizations gained by this are incredible niche.

catleeball:
@BurntSushi That's a valid point! Instead of completely removing the section, would it make sense to instead move it to FAQ.md possibly? That way the info is still there without having to discover it in the source.

catleeball:
I'll close this PR, let me know if you'd like me to move the docs around any though. Thanks for making a lot of cool rust crates! :)  FYI, I've been using ripgrep at work for directories with bunches of >100mb configs and logs and it's been phenomenally faster than grep. 👍

PR #1636, opened 2020-07-07
cli: add colored output

pickfire:


BurntSushi:
This doesn't work, because [ripgrep disables clap's color feature](https://github.com/BurntSushi/ripgrep/blob/ffd4c9ccba0ffc74270a8d3ae75f11a7ba7a1a64/Cargo.toml#L55-L58) which avoids bringing in a [second dependency for handling colors in ripgrep's dependency tree](https://github.com/clap-rs/clap/tree/v2.33.1#features-enabled-by-default).

We could reconsider this once ripgrep moves to clap 3, which IIRC, uses `termcolor` for dealing with colors instead of `ansi_term`, which is the same crate that ripgrep uses.

PR #1660, opened 2020-08-20
Ensure that trivial skipping is done prior to filesystem calls.

rkhoury:
This seems like an obvious optimization but becomes critical when
filesystem operations even as simple as stat can result in
significant overheads; an example of this was a bespoke filesystem
layer in Windows that hosted files remotely and would download them
on-demand when particular filesystem operations occurred. Users of
this system who ensured correct file-type fileters were being used
could still get unnecessary file access resulting in large downloads.

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -934,14 +934,22 @@ impl Walk {
         if ent.depth() == 0 {
             return Ok(false);
         }
-
+        // We ensure that trivial skipping is done before any other potentially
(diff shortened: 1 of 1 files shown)


rkhoury:
This is based on the discussion: https://github.com/BurntSushi/ripgrep/discussions/1657
I've only rearranged the code that affected my particular scenario, so this is not intended to be an exhaustive optimization pass.  Let me know if the code comment is what you had imagined or is too verbose.

PR #1671, opened 2020-08-29
crates/globset: (experimental) Add support for matching parent directories of a path

andrewhickman:
Hi,

I mostly wrote this code to suit my own need of an efficient way to visit all the paths matched by a glob. As such it might not be suitable for inclusion in this crate, so feel free to close this. However I thought it might be useful :slightly_smiling_face:

This adds `Glob::compile_parent_matcher`, which returns a regex that matches not just the globs paths, but all parent paths as well. For example the regex for the glob `a/b/*` would look something like `^a(?:b/(?:/.*)?)?$`. By using this regex as a filter in `walkdir` or similar, it will avoid traversing directories unnecessarily.

### DIFF
--- a/crates/globset/src/glob.rs
+++ b/crates/globset/src/glob.rs
@@ -120,7 +120,33 @@ pub struct GlobMatcher {
     re: Regex,
 }
 
+/// A parent matcher for a single pattern.
+#[derive(Clone, Debug)]
+pub struct GlobParentMatcher {
+    /// The underlying pattern.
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks! I think this is an interesting idea, but I think I'd want to consider the feature a bit more holistically before adding public APIs. I might very well circle back around to this at some point (likely not soon though). But for now, I'm going to close this.

PR #1675, opened 2020-09-02
Simplify release workflow

murlakatamenka:
Remove unnecessary steps in the current release workflow to make it simpler and less hacky.

Good reference:
- https://eugene-babichenko.github.io/blog/2020/05/09/github-actions-cross-platform-auto-releases/

[Runs successfully](https://github.com/murlakatamenka/ripgrep/actions/runs/235810016) across all the platforms.

---

Also relevant thing that I learned alongside:

> `::set-env name={name}::{value}`
>
> The action that creates or updates the environment variable does not have access to the new value, but all subsequent actions in a job will have access.

[source](https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable)

Isn't intuitive to me.

---

Workflow file can be additionally formatted for better indentation, I didn't do it to make reviewing easier.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,14 +1,17 @@
-# The way this works is a little weird. But basically, the create-release job
-# runs purely to initialize the GitHub release itself. Once done, the upload
-# URL of the release is saved as an artifact.
+# The way this works is the following:
 #
-# The build-release job runs only once create-release is finished. It gets
-# the release upload URL by downloading the corresponding artifact (which was
-# uploaded by create-release). It then builds the release executables for each
-# supported platform and attaches them as release assets to the previously
-# created release.
+# - create-release job runs purely to initialize the GitHub release itself
+# and to output upload_url for the following job.
+#
+# - build-release job runs only once create-release is finished. It gets
+# the release upload URL from create-release job outputs, then builds
+# the release executables for each supported platform and attaches them
+# as release assets to the previously created release.
 #
 # The key here is that we create the release only once.
+#
+# Reference:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -17,58 +20,33 @@ on:
...
-      - name: Save version number to artifact
-        run: echo "${{ env.RG_VERSION }}" > artifacts/release-version
-
-      - name: Upload artifacts
-        uses: actions/upload-artifact@v1
-        with:
-          name: artifacts
-          path: artifacts
+          tag_name: ${{ github.ref }}
+          release_name: ${{ github.ref }}
```
BurntSushi, line 38: Can you say why this works?

Oh, I see, you moved the logic down below.

BurntSushi (approved):
I think this LGTM. I'll merge and tweak this is a bit once I'm ready to do the next release. Thanks!

.github/workflows/release.yml
murlakatamenka: It's right from the examples of [`@actions/create-release`](https://www.github.com/actions/create-release) and [`@actions/upload-release-asset`](https://www.github.com/actions/upload-release-asset). They're smart enough to strip the unnecessary part and leave only tag name. Cannot find the resource where it was mentioned, but it works as we can see :man_shrugging:

murlakatamenka:
Good to have CI / CD improved since it often serves as a reference. I've seen

> Apparently, this is the right way to get a tag name. Really?

in many other Rust projects 😄 

Thanks for the good work, as always 👍

PR #1692, opened 2020-09-29
A typo?

qm3ster:
A cursory glance at TYPO3 documentation suggests they use `.typoscript` files.
however, there is `"*.vue"` in the `"js"` category, perhaps `.ts` should go there as well?


BurntSushi:
TypoScript was added in https://github.com/BurntSushi/ripgrep/pull/1477, and as far as I can tell, it's correct.

Typescript already has a type in ripgrep, but it's called `ts`.

qm3ster:
Ah, I see it now. And having multiple types claim an extension should be fine too, there are three instances of `"*.v"` for example.
Sorry.

PR #1693, opened 2020-09-30
deps: use fs-err instead of std::fs in walkdir

hoodie:
Hi there, thanks so much for ripgrep, I use it every day!
I just stumbled upon this little crate called [fs-err](https://github.com/andrewhickman/fs-err) which produces nicer fs errors but is basically a drop in replacement for `std::err`.

I thought I'd give it a try in ripgrep et voilá

before:
```bash
~  /usr/bin/rg password /root
/root: Permission denied (os error 13)
```

after:
```bash
~  rg password /root 
/root: failed to read directory `/root`
```

thank you for your consideration

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -135,6 +135,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3f9eec918d3f24069decb9af1554cad7c880e2da24a9afd88aca000531ab82c1"
 
 [[package]]
+name = "fs-err"
+version = "2.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


okdana:
Is it nicer? It seems like it provides less useful information

hoodie:
@okdana actually you might be right 😀

BurntSushi:
@hoodie I had the same reaction as @okdana. And ripgrep already tries pretty hard to add file paths to error messages, which appears to be the primary benefit of something like `fs-err`.

hoodie:
It was mainly a ln exercise for me, I was so surprised that it was so easy to integrate that I didn't validate the results thoroughly enough

PR #1706, opened 2020-10-16
doc: elaborate on the function of -u/--unrestricted

### LINKED ISSUE #1703: Short Alias for --no-ignore
`--no-ignore` is, in my experience, one of the most common flags one would want to pass to ripgrep, but unfortunately it doesn't have a single character alias for its option. Obviously you can just add it by creating an alias under a different name, but it's annoying to do that on every system. It would be great if ripgrep supported such an option out of the box.

okdana:
The added wording reads a little redundant in the man page since the *AUTOMATIC FILTERING* section says basically the same thing. Not sure if that's a problem. But this should clarify the `--help` output at least?

Fixes #1703

### DIFF
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -1945,6 +1945,10 @@ Don't respect ignore files (.gitignore, .ignore, etc.). This implies
 This does *not* imply --no-ignore-files, since --ignore-file is specified
 explicitly as a command line argument.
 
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
This looks great, thanks! I think the redundancy is okay. The problem is that some folks might only look at one particular place and not the other.

Issue #1710, opened 2020-10-19
Add a type for minified files


duongdominhchau:
When searching, usually the matches from minified files are unwanted because each match is extremely long and the match can be found in original file as well. Because matches from minified files are rarely desired, and currently `rg --type-list` has no type for minified files, I suggest adding a `minified` type so people can ignore minified files while searching by using `--type-not minified`. This new type should be printed as `minified: *.min.js, *.min.css, *.min.html` when we run `rg --type-list`.

BurntSushi:
I think this seems reasonable to me. I don't usually track open issues for file types. Just submitting a PR is fine.

Also, you might be interested in the `-M/--max-columns` flag. It is particularly helpful with things like minified files since it will suppress extremely long lines.

PR #1711, opened 2020-10-19
Add a type for minified files

duongdominhchau:
Well, I don't know Rust and the directory structure looks weird. Thought it would be hard to make the change, but `rg` told me I was wrong, `rg '\*.toml'` shows me exactly what I want. Thank you for this awesome tool!

This PR is for issue #1710


BurntSushi (approved):
Great, thank you!

PR #1712, opened 2020-10-19
Remove regex unicode dependency from globset

ajeetdsouza:
I don't think globset requires Unicode support in regex for path matching (correct me if I'm wrong). I removed the dependency to make the crate lighter.

This won't affect the size of ripgrep, since it depends on Unicode in regex elsewhere. However, I've been meaning to use globset in [zoxide](https://github.com/ajeetdsouza/zoxide), and I'd like to keep the binary size as small as possible.

I tested a trivial program before and after the change, and the executable size went from 2544 KB to 2084 KB.

Thanks!

### DIFF
--- a/crates/globset/Cargo.toml
+++ b/crates/globset/Cargo.toml
@@ -23,7 +23,7 @@ aho-corasick = "0.7.3"
 bstr = { version = "0.2.0", default-features = false, features = ["std"] }
 fnv = "1.0.6"
 log = "0.4.5"
-regex = "1.1.5"
+regex = { version = "1.1.5", default-features = false, features = ["perf", "std"] }
 serde = { version = "1.0.104", optional = true }
 
 [dev-dependencies]


BurntSushi (approved):
Hmmm, yes, I _believe_ this is indeed correct. Good catch. In particular, the glob-to-regex transform explicitly disables Unicode support in the regex, so I think that means it's impossible for the regex to make use of any of the Unicode features.

ajeetdsouza:
Yup, just saw that too. Thanks!

BurntSushi:
Aye, yeah, passing the test suite is good enough for me. I expect we'll find out if this was an unwise choice after the next release! Hah. But I think it should be fine.

BurntSushi:
Whoops, I forgot to put out a new release of globset with this change. This PR is now in `globset 0.4.6` on crates.io.

ajeetdsouza:
Awesome, thank you!

PR #1716, opened 2020-10-23
Generalize bazel type to include *.bazel

emidln:
Bazel supports `BUILD.bazel` as well as `WORKSPACE.bazel`. In
addition, it is common to ship BUILD/WORKSPACE templates for
external repositories suffixed with .bazel for easier tool
recognition.


BurntSushi (approved):
SGTM!

PR #1720, opened 2020-10-31
Add Futhark recognition to ignore crate

vmchale:
This adds recognition of [Futhark](https://futhark-lang.org/) files.

Thanks! :)


BurntSushi (approved):
Thanks!

PR #1728, opened 2020-11-09
ci: install cross from crates.io

taiki-e:
Now that a new version of `cross` that includes the fix of https://github.com/rust-embedded/cross/issues/357 has been released (https://github.com/rust-embedded/cross/issues/359), we can resolve this FIXME.


BurntSushi (approved):
Lovely, thank you!

PR #1738, opened 2020-11-21
README: installation: fix link to .deb

tleb:
Small fix, updating the `.deb` link in the install commands. Also, a `.deb` can be installed using `apt install ./foo.deb`, should this be recommended instead?


BurntSushi (approved):
Thanks!

> Also, a .deb can be installed using apt install ./foo.deb, should this be recommended instead?

I don't know..? I don't change things for the sake of change. If you have a compelling reason to change it, then please open a new PR and we can consider it.

PR #1740, opened 2020-11-22
ignore: Provide IO Error

epage:
`ignore::Error` wraps `std::io::Error` with additional information (as
well as expose non-IO errors).  For people wanting to inspect what the
error is, they have to recursively match the Enum.  Like with `is_io()`,
this provides `io_error` and `into_io_error` helpers to do this for
the user.  These are written to be consistent with `walkdir::Error`.

### DIFF
--- a/crates/ignore/src/lib.rs
+++ b/crates/ignore/src/lib.rs
@@ -196,6 +196,71 @@ impl Error {
         }
     }
 
+    /// Inspect the original [`io::Error`] if there is one.
+    ///
+    /// [`None`] is returned if the [`Error`] doesn't correspond to an
+    /// [`io::Error`]. This might happen, for example, when the error was
+    /// produced because a cycle was found in the directory tree while
+    /// following symbolic links.
+    ///
+    /// This method returns a borrowed value that is bound to the lifetime of the [`Error`]. To
+    /// obtain an owned value, the [`into_io_error`] can be used instead.
+    ///
+    /// > This is the original [`io::Error`] and is _not_ the same as
+    /// > [`impl From<Error> for std::io::Error`][impl] which contains additional context about the
+    /// error.
+    ///
+    /// [`None`]: https://doc.rust-lang.org/stable/std/option/enum.Option.html#variant.None
+    /// [`io::Error`]: https://doc.rust-lang.org/stable/std/io/struct.Error.html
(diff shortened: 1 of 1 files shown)


crates/ignore/src/lib.rs
```diff
@@ -196,6 +196,27 @@ impl Error {
...
+            Error::WithLineNumber { ref err, .. } => err.io_kind(),
+            Error::WithPath { ref err, .. } => err.io_kind(),
+            Error::WithDepth { ref err, .. } => err.io_kind(),
+            Error::Loop { .. } => None,
+            Error::Io(ref err) => Some(err.kind()),
+            Error::Glob { .. } => None,
+            Error::UnrecognizedFileType(_) => None,
+            Error::InvalidDefinition => None,
+        }
+    }
```
BurntSushi, line 218: I think it would probably be better to just expose a `&std::io::Error`, no? It gives strictly more information and also provides access to a `std::io::ErrorKind`. Might as well provide an `into_io_error` as well.

This is how walkdir does it: https://docs.rs/walkdir/2.3.1/walkdir/struct.Error.html#method.io_error

BurntSushi (changes requested):

crates/ignore/src/lib.rs
epage: Good to know there is a baseline to be consistent with.  I copied things over from there.

When writing this, I was split on which direction to go.  I saw the `ErrorKind` being useful for programmatic use and `Error` being useful for error propagation but wasn't as clear what the value of a `&Error` would be.  However, there is value in not being prescriptive and letting the user decide what has value.  And then comes along consistency :)

epage:
Random asides from before I saw `WalkDir::Error`:
- I considered naming `io_error` as `io_error_ref` though from the API guidelines `as_io_error_ref` would probably be the most "correct" but at that point its a bit verbose.
- I considered `into_io_error` returning a `Result<std::io::Error, Self>`, kind of like what some String operations do, to avoid any data loss but the `Partial` case made this messier.  I figured someone could take the non-IO case and `to_string` it and put it into an `std::io::Error`.

BurntSushi (approved):
Nice, LGTM!

BurntSushi:
This PR is on crates.io in `ignore 0.4.17`.

PR #1742, opened 2020-11-23
Fix stdin detection when using Powershell in Unix (fix #1741)

### LINKED ISSUE #1741: Standard input detection doesn't work under PowerShell on macOS

```
ripgrep 12.1.1
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

`brew install ripgrep`

macOS Big Sur 11.0.1 (20B29)

#### Describe your bug.

Under PowerShell, when piping output to ripgrep it seems to search in the current directory instead of in the piped input.

In a directory containing a single file `test.txt` containing the line "infile", try searching input using `rg`.

```
$ echo "input" | rg input

$ echo "input" | rg input -
input

$ echo "input" | rg infile
test.txt
1:infile
```

The input should be searched instead of current directory.

r-darwish:
It seems that Powershell uses sockets instead of FIFOs to redirect the output between commands

### DIFF
--- a/crates/cli/src/lib.rs
+++ b/crates/cli/src/lib.rs
@@ -210,7 +210,7 @@ pub fn is_readable_stdin() -> bool {
             Err(_) => return false,
             Ok(md) => md.file_type(),
         };
-        ft.is_file() || ft.is_fifo()
+        ft.is_file() || ft.is_fifo() || ft.is_socket()
     }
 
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thank you so much! I really appreciate you digging into this and fixing it. :-)

PR #1756, opened 2020-12-08
globset: fix recursive suffix over matching

zertosh:
The syntax documentation correctly states how recursive suffixes should
behave:

> `**` recursively matches directories [..] `foo/**` matches `foo/a` and
> `foo/a/b`, but not `foo`.

Previously `**` would actually match `foo`. This patch makes it so that
it doesn't. Also adds support to the "prefix" strategic matcher so that
it handles common cases like `foo/**`.

### DIFF
--- a/crates/globset/src/glob.rs
+++ b/crates/globset/src/glob.rs
@@ -403,7 +403,7 @@ impl Glob {
         if self.opts.case_insensitive {
             return None;
         }
-        let end = match self.tokens.last() {
+        let (end, need_sep) = match self.tokens.last() {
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
Makes sense to me, thank you!

passcod:
Even though I agree this is a fix as per documentation, perhaps it should have been a semver bump? It changed behaviour (and broke my tests)...

BurntSushi:
Perhaps. But in this case, this is bringing the implementation in line with documented behavior.

Issue #1404, opened 2019-10-15
Display a warning when searching within an ignored directory


Manishearth:
#### What version of ripgrep are you using?
```
$ rg --version
ripgrep 11.0.2
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

#### How did you install ripgrep?
Cargo

#### What operating system are you using ripgrep on?
Ubuntu

#### Describe your question, feature request, or bug.

Sometimes you end up inside a directory that's actually gitignored, and search for stuff. Of course, no file will get searched, and you will be confused.

If ripgrep ends up not searching any files and it happens to be in a situation where there _are_ files to search that it cannot because of the ignore file, it should probably display a warning saying "some files were skipped because of the ignore file, try `--no-ignore-vcs` to include". 

I don't know if ripgrep ever shows warnings, but this would be a nice place to do so.

BurntSushi:
This feature actually existed at one point. Unfortunately, it looks like it has evaporated, probably intentionally during a refactor. Getting this back in is a good idea.

coderkalyan:
Hi, I'd like to try to implement this for #hacktoberfest. I understand the feature request, but does someone have a simple example I could use to test this feature? In addition, what's the layout of this codebase - where would this feature be implemented?

genivia-inc:
Removing UI usage confusion is always a good thing. I am sure I'm preaching to the choir here, but I'd suggest either with by issuing warnings, or even better by [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) design.

Following POLA, files and directories specified as arguments to `rg` (i.e. to be explicitly searched) should never be ignored. It's utterly confusing when that happens. So I came up [with an alternative](https://github.com/Genivia/ugrep#ignore) that both respects `.gitignore` and files and directories specified as arguments to be searched. Furthermore, I though it would be a good idea to have option `--stats` extended to show the selection criteria applied to the search results and the locations of each `.gitignore` found when applied to the results.

For example, searching the working directory that has a `.gitignore`, while also searching the `Makefile` even though it is excluded by a `.gitignore` rule, now produces predictable results:
~~~
$ ugrep -rl --ignore-files --stats 'c++' . Makefile
...
Makefile
...
Searched 43 files with 8 threads in 1 directory: 42 matching
The following pathname selections and restrictions were applied:
--ignore-files='.gitignore'
  ./.gitignore exclusions were applied to .
~~~
[...]

PR #1759, opened 2020-12-10
Display a warning when nothing is searched

goto-engineering:
Relates to: https://github.com/BurntSushi/ripgrep/issues/1404

I brought back the "nothing searched message" when no files were matched.

Unsure how/where to add unit tests for this, I didn't find any tests for `main`.


crates/core/main.rs
```diff
@@ -68,11 +68,20 @@ fn try_main(args: Args) -> Result<()> {
         process::exit(0)
     } else if messages::errored() {
         process::exit(2)
+    } else if !matched {
```
goto-engineering, line 71: I wasn't sure if the last `else` branch can be reached in any case that isn't the "nothing searched" case, so I made a new `else if`.

BurntSushi: Yes, the `else` branch is only reached when `matched` is false.

crates/core/main.rs
```diff
@@ -68,11 +68,22 @@ fn try_main(args: Args) -> Result<()> {
...
+        eprint_nothing_searched();
+        process::exit(1)
```
BurntSushi, line 73: Hmmm, this is definitely not correct. :-( This is printing a message when nothing is _matched_, but what we want to do is print a message when nothing is _searched_.

BurntSushi (changes requested):
Thanks! Unfortunately, this is not quite correct. I left a comment describing why below.

> Unsure how/where to add unit tests for this, I didn't find any tests for main.

`main` is probably the most tested routine in all of ripgrep. Tests for this probably want to go in `tests/feature.rs`.

crates/core/main.rs
goto-engineering: Oh, I see. I can't seem to find the logic where the error message used to be shown, i.e. where we can conclude that there are no search candidates. Do you happen to know the area so I can hone in on it? I'm mostly looking in `walk.rs` right now.

BurntSushi: I would keep trying in `main.rs`. In particular, I believe the logic will need to be in both the `search` and `search_parallel` routines. For the former, I think it needs to check if `searcher.search()` is ever executed. Similarly for the latter.

PR #1762, opened 2020-12-15
Add warning if nothing was searched

goto-engineering:
Ok, another attempt at fixing https://github.com/BurntSushi/ripgrep/issues/1404.

I somehow messed up the other PR in the Github UI, not sure how. You can delete it.

Let me know if this is on the right track, and I'll look into tests.

### DIFF
--- a/crates/core/main.rs
+++ b/crates/core/main.rs
@@ -83,12 +83,14 @@ fn search(args: &Args) -> Result<bool> {
     let mut stats = args.stats()?;
     let mut searcher = args.search_worker(args.stdout())?;
     let mut matched = false;
+    let mut searched = false;
 
     for result in args.walker()? {
         let subject = match subject_builder.build_from_result(result) {
             Some(subject) => subject,
             None => continue,
         };
+        searched = true;
         let search_result = match searcher.search(&subject) {
             Ok(search_result) => search_result,
             Err(err) => {
@@ -107,6 +109,9 @@ fn search(args: &Args) -> Result<bool> {
         if matched && quit_after_match {
             break;
         }
+    }
+    if !searched {
+        eprint_nothing_searched();
     }
     if let Some(ref stats) = stats {
         let elapsed = Instant::now().duration_since(started_at);
@@ -129,11 +134,13 @@ fn search_parallel(args: &Args) -> Result<bool> {
     let bufwtr = args.buffer_writer()?;
     let stats = args.stats()?.map(Mutex::new);
     let matched = AtomicBool::new(false);
+    let searched = AtomicBool::new(false);
     let mut searcher_err = None;
     args.walker_parallel()?.run(|| {
         let bufwtr = &bufwtr;
         let stats = &stats;
         let matched = &matched;
+        let searched = &searched;
         let subject_builder = &subject_builder;
         let mut searcher = match args.search_worker(bufwtr.buffer()) {
             Ok(searcher) => searcher,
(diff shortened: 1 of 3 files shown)


BurntSushi:
> I somehow messed up the other PR in the Github UI, not sure how. You can delete it.

Please don't delete PRs or issues. I'm really surprised GitHub gives you the ability to delete issues/PRs filed on another repo. Please don't do that. Because you deleted it, we've now lost some historical context and some helpful info on a false start. Sure, the PR might have been wrong, but we don't just sweep wrong things under the rug. We learn from them.

Aside from that, you shouldn't have needed to open a new PR. For a small change like this, it's enough to keep it inside a single commit. When you make a change, just amend the commit. (`git commit -a --amend` for example.) And since the history is altered, you'll need to force push it, e.g., `git push <remote> <branch> --force`.

BurntSushi (changes requested):
I think this looks correct to me! Have you tried it? Does it work?

I think adding some tests would be a good idea. They should go in `tests/feature.rs`. I think you'll find `assert_non_empty_stderr()` helpful in this case. Here's an example of another test that uses it: https://github.com/BurntSushi/ripgrep/blob/a6d05475fb353c756e88f605fd5366a67943e591/tests/feature.rs#L44-L47

Take a look at other tests in that file to see how to setup directory trees and what not.

goto-engineering:
> > I somehow messed up the other PR in the Github UI, not sure how. You can delete it.
> 
> Please don't delete PRs or issues. I'm really surprised GitHub gives you the ability to delete issues/PRs filed on another repo. Please don't do that. Because you deleted it, we've now lost some historical context and some helpful info on a false start. Sure, the PR might have been wrong, but we don't just sweep wrong things under the rug. We learn from them.
> 
> Aside from that, you shouldn't have needed to open a new PR. For a small change like this, it's enough to keep it inside a single commit. When you make a change, just amend the commit. (`git commit -a --amend` for example.) And since the history is altered, you'll need to force push it, e.g., `git push <remote> <branch> --force`.

Sorry, I got confused with the whole fork thing on github. Since there was only 1 commit, I deleted the branch on my fork, not realizing that would close the PR. I had intended to just force push my new commit to the same branch name.

goto-engineering:
> I think this looks correct to me! Have you tried it? Does it work?
> 
> I think adding some tests would be a good idea. They should go in `tests/feature.rs`. I think you'll find `assert_non_empty_stderr()` helpful in this case. Here's an example of another test that uses it:
> 
> https://github.com/BurntSushi/ripgrep/blob/a6d05475fb353c756e88f605fd5366a67943e591/tests/feature.rs#L44-L47
> 
> Take a look at other tests in that file to see how to setup directory trees and what not.

Yes, I've tried it with this setup of files:
```
.ignore - (ignored-dir/**)
ignored-dir/file.txt ("needle")
other-file.txt (doesn't contain needle)
```
Then I do `rg needle`. If `other-file.txt` is there, it doesn't show the warning. If only `ignored-dir` is there, it does show up.

Thanks for the hints, I'll look into the tests later!

goto-engineering:
I added a test case for the warning and it passes. But another test (`regression::r428_unrecognized_style`) fails because its setup also triggers the warning.

I think it's because the case being tested there throws an exception and therefore never triggers my `searched = true`. I'm not really sure how to handle that. Since it's about an unrecognized style, is this happening in the builder itself?

goto-engineering:
@BurntSushi Ping just in case you hadn't seen this

PR #1786, opened 2021-01-17
Updates all dependencies; replaces memmap with memmap2

### LINKED ISSUE #1785: RUSTSEC-2020-0077: memmap: memmap is unmaintained
When `cargo-audit` is called on any crate that directly or indirectly depends on `grep-searcher`, the following error will appear:

```
Crate:         memmap
Version:       0.7.0
Warning:       unmaintained
Title:         memmap is unmaintained
Date:          2020-12-02
ID:            RUSTSEC-2020-0077
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0077
Dependency tree:
memmap 0.7.0
└── grep-searcher 0.1.7
```

It can be easily remedied by switching from memmap to memmap2 or or mapr (more info at https://rustsec.org/advisories/RUSTSEC-2020-0077, https://github.com/danburkert/memmap-rs/issues/90)

I would be willing to attempt to implement this and submit a PR if it would be accepted.

brightly-salty:
Fixes #1785 

Originally replaced memmap with memmap2 per `cargo-audit`;
Also upgraded all dependencies

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,9 +2,9 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.14"
+version = "0.7.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b476ce7103678b0c6d3d395dbbae31d48ff910bd28be979ba5d48c6351131d0d"
+checksum = "7404febffaa47dac81aa44dba71523c9d069b1bdc50a77db41195149e17f68e5"
 dependencies = [
  "memchr",
 ]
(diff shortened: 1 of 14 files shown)


BurntSushi:
Thanks, I've done this myself. While I don't document this anywhere, I _always_ do dependency updates myself. That's so I can carefully review the changes and inspect any new dependencies that may sneak in and understand why they are needed. Accepting a PR like this that not only updates every dependency but _also_ bumps the minimal version required for each dependency would make it very very easy to miss if a new dependency ended up sneaking in.

I say all this in an effort to help others understand and appreciate that dependencies should be carefully scrutinized because they are a liability.

Thanks for bringing the RUSTSEC advisory to my attention and submitting a patch though. I appreciate it!

BurntSushi:
Note though that I did not bump the minimal version of each dependency. I don't see a strong reason to do this, and if I were consistent about it, it would lead to an annoyingly high rate of churn.

brightly-salty:
> Note though that I did not bump the minimal version of each dependency. I don't see a strong reason to do this, and if I were consistent about it, it would lead to an annoyingly high rate of churn.

All I did was run `cargo-upgrade` (part of `cargo-edit`) which automatically did this. I completely understand and appreciate your quick response.

PR #1789, opened 2021-01-26
Add *.BUILD and *.bazelrc patterns to the default bazel type

vors:
It's common to use `foo.BUILD` for the external workspaces `@foo` build files in bazel.
Example from pytorch https://github.com/pytorch/pytorch/blob/master/third_party/cpuinfo.BUILD
See more files like that in the same folder.

Also, add `*.bazelrc` pattern while we are here.
https://docs.bazel.build/versions/master/guide.html#bazelrc-the-bazel-configuration-file

The default one used by bazel is just plain `.bazelrc`, but it's common to use `user.bazelrc` and other named groupings and import them from the top-level one.


BurntSushi (approved):
Thanks!

PR #1801, opened 2021-02-12
Fix/Ignore HTML output of -Z timings

hbina:


BurntSushi:
Why is the diff for this PR larger than it needs to be?

hbina:
Hmm let me try again, I was using Windows, might have to do with CRLF.

BurntSushi (approved):
Thanks!

PR #1802, opened 2021-02-12
These structs are extremely large!

hbina:
The smallest of them is 600 bytes. Is there a benchmark somewhere?

### DIFF
--- a/crates/regex/src/matcher.rs
+++ b/crates/regex/src/matcher.rs
@@ -95,7 +95,7 @@ impl RegexMatcherBuilder {
         }
 
         let matcher = MultiLiteralMatcher::new(&slices)?;
-        let imp = RegexMatcherImpl::MultiLiteral(matcher);
+        let imp = RegexMatcherImpl::MultiLiteral(Box::new(matcher));
         Ok(RegexMatcher {
             config: self.config.clone(),
             matcher: imp,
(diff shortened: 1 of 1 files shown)


BurntSushi:
I don't understand this PR. Could you please motivate it? "The structs are large" isn't good enough.

hbina:
Sorry about that, from the commit
```
The smallest of them is 600 bytes.
From my (admittedly minimal) knowledge, this causes trashing.

Since these structs are so large, objects before and after this object
won't exist in the same page. Assuming that these values are being used together,
they require 2 pages to actually work instead of (possibly) just 1.

Was the point to prevent a syscall?

```

BurntSushi:
If you can provide a meaningful benchmark motivating this change, then we can think abiut merging. But without that, I don't know that this is a problem, even if your theory is correct.

marcospb19:
@hbina If 600 bytes don't fit in a page, moving the same 600 bytes to the `heap`, they still wouldn't fit in a page, right?

Cause the page size is the same for the stack and the heap, or maybe I'm missing something.

Also, I think that these changes add some problems:
- Unnecessary increase of code complexity

And because we are talking about tiny optimizations:
- Waiting for allocator to allocate (and deallocate) the resources
- One more level of indirection, adding a branch and a possible cache miss

And AFAIK, each one of the above would make code even slower.

BurntSushi:
While it is plausible that big structs cause perf problems---I've certainly seen and fixed a few of those in the past---primarily by spending more time copying data around, I'm not convinced that that is the case here. I'd need to see a real benchmark to be sure. So with that, I'm going to close it.

Issue #1765, opened 2020-12-21
Panic with --crlf due to integer underflow


whatisaphone:
#### What version of ripgrep are you using?

ripgrep 12.1.1 (rev 7cb211378a)
-SIMD -AVX (compiled)
+SIMD -AVX (runtime)

#### How did you install ripgrep?

scoop install ripgrep

#### What operating system are you using ripgrep on?

Windows 10 Build 19041

#### Describe your bug.

With some regexes, combining --crlf with input that contains unix line endings causes a panic.

#### What are the steps to reproduce the behavior?

Minimal repro:

```sh
echo '' | rg --crlf x?
```

Don't forget that `echo ''` writes a line terminator:

```
echo '' | xxd
00000000: 0a                                       .
```

#### What is the actual behavior?

```
echo '' | RUST_BACKTRACE=full rg --debug --crlf x?
```

```
DEBUG|globset|crates\globset\src\lib.rs:431: built glob set; 0 literals, 0 basenames, 12 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 0 regexes
DEBUG|globset|crates\globset\src\lib.rs:431: built glob set; 0 literals, 3 basenames, 1 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 0 regexes
thread 'main' panicked at 'index out of bounds: the len is 1 but the index is 18446744073709551615', D:\a\ripgrep\ripgrep\crates\printer\src\standard.rs:1476:38
```

<details>
<summary>You'll be disappointed with the stack trace, but I'm including it anyway, as requested</summary>

```
stack backtrace:
   0:     0x7ff61f25899e - <unknown>
   1:     0x7ff61f276f5c - <unknown>
   2:     0x7ff61f2548ac - <unknown>
   3:     0x7ff61f25bcbb - <unknown>
   4:     0x7ff61f25b908 - <unknown>
   5:     0x7ff61f25c496 - <unknown>
   6:     0x7ff61f25c01f - <unknown>
   7:     0x7ff61f275820 - <unknown>
   8:     0x7ff61f2757e7 - <unknown>
   9:     0x7ff61f030748 - <unknown>
  10:     0x7ff61f01f2c3 - <unknown>
  11:     0x7ff61f03db67 - <unknown>
  12:     0x7ff61f048c3c - <unknown>
  13:     0x7ff61f0109c5 - <unknown>
  14:     0x7ff61efad5b2 - <unknown>
[...]

PR #1803, opened 2021-02-12
Prevent crash when user provides bad input.

hbina:
Running "echo '' | rg --crlf x?" should no longer crash.
I probably introduced performance regressions haha

Related issue: https://github.com/BurntSushi/ripgrep/issues/1765

aww shucks, sublice pattern is unstable. If you really want this, I could reimplement it using more backward compat ways.


BurntSushi:
Thanks! But I believe it is intentional that `is_suffix` is defined this way. I ended up going in a different direction. (See linked commit.)

PR #1805, opened 2021-02-16
Compile under WASI

jcaesar:
Hi, and thank you for this awesome tool.

I'm currently playing around with porting some [Rust Alternative Command Line Tools](https://zaiste.net/posts/shell-commands-rust/) to wasm32-wasi. For most tools, this requires some rather unsavoury code changes, but rg is a pretty exception. So I decided to see if you don't want to add support for wasi as a compilation target.

I did say "pretty", but there are some points…
* memmap2 does not want to compile under anything except windows and unix ([conscious decision](https://github.com/RazrFalcon/memmap2-rs/pull/10#issuecomment-778788586)), so it requires some macros
* I did default `is_readable_stdin` to false on unsupported platforms (because same_file returns errors on unsupported platforms, and imp translates errors to false), but I'm not sure that is correct. Would it be better to panic, or leave the function unimplemented on unsupported platforms?
* PCRE2 won't build since it requires(?) pthreads
* Cross can't run tests on wasm32-wasi

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,6 +83,10 @@ jobs:
         - build: win-gnu
           os: windows-2019
           rust: nightly-x86_64-gnu
+        - build: stable-wasi
+          os: ubuntu-18.04
+          rust: stable
+          target: wasm32-wasi
     steps:
     - name: Checkout repository
       uses: actions/checkout@v2
@@ -103,6 +107,7 @@ jobs:
         toolchain: ${{ matrix.rust }}
         profile: minimal
         override: true
+        target: ${{ matrix.target }}
 
     - name: Use Cross
       if: matrix.target != ''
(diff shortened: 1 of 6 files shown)


BurntSushi:
Thanks for the PR. Could you say why you want to do this?

jcaesar:
Why **I** want to do this? I suppose I'm intrigued by the idea of system-independent binaries.

But if you're asking what the benefits of merging this are… admittedly rather thin. To run wasm/wasi binaries, you do need a runtime like wasmtime, wasmer, or lucet, and those aren't (yet?) easily available through distros or installed on your typical windows machine. So for now, this would be a rather niche way of using ripgrep.

BurntSushi:
All righty. I would like to hold off then until there is a more concrete use case for running ripgrep in this way.

jcaesar:
(It seems that mmap2 will now (v0.2.3) compile under WASI. In case this gets reconsidered, the only remaining necessary change is to `crates/cli/src/lib.rs `.)

PR #1831, opened 2021-03-20
ci: fix deb build script in clean checkout

aswild:
If ripgrep hasn't been built yet (i.e. target/debug/ doesn't exist),
then cargo-out-dir can't find OUT_DIR and the copy commands fail. Fix by
running cargo build before finding OUT_DIR.

Also add a check to fail early with a sensible error message when
asciidoctor isn't installed, rather than failing because of a missing
rg.1 file after the build.

### DIFF
--- a/ci/build-deb
+++ b/ci/build-deb
@@ -17,16 +17,21 @@ if ! command -V cargo-deb > /dev/null 2>&1; then
     exit 1
 fi
 
+if ! command -V asciidoctor > /dev/null 2>&1; then
+    echo "asciidoctor command missing" >&2
+    exit 1
+fi
+
 # 'cargo deb' does not seem to provide a way to specify an asset that is
 # created at build time, such as ripgrep's man page. To work around this,
 # we force a debug build, copy out the man page (and shell completions)
 # produced from that build, put it into a predictable location and then build
 # the deb, which knows where to look.
+cargo build
 
 DEPLOY_DIR=deployment/deb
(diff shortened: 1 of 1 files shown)


aswild:
Example output when running the current version of build-deb from a clean checkout:

```
% ci/build-deb                                                                                                                                                 
find: ‘target/debug/’: No such file or directory                                                                                                               
   Compiling memchr v2.3.4                                                                                                                                     
---snip---                                                                                    
    Finished dev [unoptimized + debuginfo] target(s) in 12.03s                                                                                                 
cp: cannot stat './rg.1': No such file or directory                                                                                                            
cp: cannot stat './rg.bash': No such file or directory                                                                                                         
cp: cannot stat './rg.fish': No such file or directory                                                
```

Also the asciidoctor check could be expanded to allow `a2x` as well since build.rs can fall back to that, whichever you prefer.

ci/build-deb
```diff
@@ -17,16 +17,21 @@ if ! command -V cargo-deb > /dev/null 2>&1; then
     exit 1
 fi
 
+if ! command -V asciidoctor > /dev/null 2>&1; then
+    echo "asciidoctor command missing" >&2
+    exit 1
+fi
```
BurntSushi, line 23: This LGTM. The `a2x` support in the build script is legacy I think, so it's likely to be removed.

BurntSushi (approved):
Lovely, nice catch. Thank you!

PR #1839, opened 2021-03-31
Update outdated comment in line_buffer.rs

aricha1940:
Looks like this was accidentally left set to 8 in commit 46fb77c.


BurntSushi (approved):
Ah right, thank you!

PR #1849, opened 2021-04-15
Update --hidden help section to define hidden

### LINKED ISSUE #1847: Help text for --hidden should clarify what "hidden" means

```
ripgrep 12.1.1 (rev 7cb211378a)
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

Issue repros with both `scoop install ripgrep` and `cargo install ripgrep` versions

Windows 10 Version 20H2 (OS Build 19042.867)

#### Describe your bug.

On Windows, ripgrep (via `ignore`) considers files which have the "hidden" attribute OR whose basename starts with a `.` to be "hidden" for the purposes of the `--hidden` flag

https://github.com/BurntSushi/ripgrep/blob/64ac2ebe0f2fe1c8967e7ec550bc32466bf40a07/crates/ignore/src/pathutil.rs#L22-L46

I can understand why ripgrep might want to ignore `.dotfiles` on Windows for consistency (and that changing it now would be breaking), but as a Windows user, I was surprised by this behavior because it uses a meaning of the term "hidden" which is different from the OS's without documenting that it's going to do so. It would be nice if `rg --help`'s `--hidden` section could clarify what ripgrep considers to be a "hidden" file, similar to the comment in `pathutil.rs`.

```
> rg --help | rg -A 7 -e --hidden
```

```
        --hidden
            Search hidden files and directories. By default, hidden files and directories
            are skipped. Note that if a hidden file or a directory is whitelisted in an
            ignore file, then it will be searched even if this flag isn't provided.

            This flag can be disabled with --no-hidden.

        --iglob <GLOB>...
```

Something along the lines of

```
        --hidden
            Search hidden files and directories. By default, hidden files and directories
            are skipped. Note that if a hidden file or a directory is whitelisted in an
            ignore file, then it will be searched even if this flag isn't provided.

            A file or directory is considered hidden if its base name starts with a dot
            character ('.'). On operating systems which support a `hidden` file attribute,
            like Windows, files with this attribute are also considered hidden.

            This flag can be disabled with --no-hidden.
```

My initial attempts at searching through the ripgrep documentation to understand why it was skipping `.directories` by default involved `rg --help | rg -e dot`, which pointed me to the `--no-ignore-dot` flag (which sounds promising but isn't actually relevant to ignoring dotfiles generally). If that feels like a mistake that other users might reasonably be expected to make, it might be worth additionally updating the `--no-ignore-dot` help section with a pointer along the lines of `This does **not** affect whether ripgrep will ignore files and directories whose name begins with a dot. For that, see --hidden and --no-hidden.`

I also noticed that the `--no-hidden` flag the `--hidden` section mentions doesn't have a top-level help entry; not sure if that's intentional.

dbjorge:
Fixes #1847

If you'd prefer to leave out the `--no-ignore-dot` part, I'm happy to update the PR accordingly

Thanks!

### DIFF
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -1975,7 +1975,7 @@ fn flag_no_ignore_dot(args: &mut Vec<RGArg>) {
         "\
 Don't respect .ignore files.
 
-This does **not** affect whether ripgrep will ignore files and directories
+This does *not* affect whether ripgrep will ignore files and directories
 whose names begin with a dot. For that, see --hidden.
 
 This flag can be disabled with the --ignore-dot flag.
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -1433,6 +1433,10 @@ Search hidden files and directories. By default, hidden files and directories
 are skipped. Note that if a hidden file or a directory is whitelisted in an
 ignore file, then it will be searched even if this flag isn't provided.
 
+A file or directory is considered hidden if its base name starts with a dot
+character ('.'). On operating systems which support a `hidden` file attribute,
+like Windows, files with this attribute are also considered hidden.
+
 This flag can be disabled with --no-hidden.
 "
     );
@@ -1970,6 +1974,9 @@ fn flag_no_ignore_dot(args: &mut Vec<RGArg>) {
     const LONG: &str = long!(
         "\
 Don't respect .ignore files.
+
+This does **not** affect whether ripgrep will ignore files and directories
+whose names begin with a dot. For that, see --hidden.
 
 This flag can be disabled with the --ignore-dot flag.
 "


crates/core/app.rs
```diff
@@ -1971,6 +1975,9 @@ fn flag_no_ignore_dot(args: &mut Vec<RGArg>) {
         "\
 Don't respect .ignore files.
 
+This does **not** affect whether ripgrep will ignore files and directories
```
BurntSushi, line 1978: Just looking at other text, I think you just want `*not*` here instead of `**not**`. (This unfortunately isn't Markdown. It's AsciiDoc that gets converted into roff.)

dbjorge, line 1978: This is technically inaccurate in that a .ignore file with negation patterns could affect whether ripgrep ignores dotfiles, but it felt overly verbose to clarify that as part of the note. If you have a better wording proposal that is more accurate, I'd be happy to update accordingly

dbjorge: Oops, fixed!

BurntSushi (approved):

dbjorge:
Thanks very much for the quick review!

Issue #103, opened 2016-09-26
Add filename-only option


ddrcoder:
I quite frequently use `grep -l` to simply identify files which contained any match of the given pattern.  Since only existence checking is required, you could terminate at the first match. The list of files can then be fed to `xargs` or the like. This would go hand-in-hand with the `-0, --null` option.

BurntSushi:
This is already done. It's in `0.2.0`.

Flaque:
To be clear for folks who stumbled upon this like I did, the way to only show filenames in `rg` is the same as in `grep`:

```
$ rg -l 
```

watsoncj:
Here's the relevant help section from ripgrep 12.1.0:
```
    -l, --files-with-matches
            Only print the paths with at least one match.

            This overrides --files-without-match.
```

iburunat:
Maybe helpful to know that adding the flag `-c/--counter` will show the number of lines matching the searched term. The flag `--count-matches` will count the number of matches instead (see [man page](https://www.mankier.com/1/rg#--count)).

NetMage:
I think the help could be a little clearer - it doesn't clearly imply that the matches won't be printed.

SoftwareApe:
I agree with @NetMage I looked through all the file related options in the help and this one didn't seem to apply.

>Only print the paths with at least one match.

To me this says that it will only print those files that have at least one match. Which sounds like the default behavior. It doesn't say that it *doesn't* print the match, which is what most people are likely looking for.

I think it would be more clear if the help read

>Only print the path to the files with matches, not the matches.

NightMachinary:
> To be clear for folks who stumbled upon this like I did, the way to only show filenames in `rg` is the same as in `grep`:
> 
> ```
> $ rg -l 
> ```

`rg -l .`, to just list everything.

RichardBronosky:
Thank you for this! I used it to make `vg`, the perfect companion to `rg`.

```sh
function vg() { $EDITOR $(rg -l "$@"); }
```
Then I can follow up "find all files containing *pattern*" with "edit them".

PR #1869, opened 2021-05-25
fix: Clarify the help of --files-with-matches and --files-without-match

lf-:
See
https://github.com/BurntSushi/ripgrep/issues/103#issuecomment-763083510

I was confused as to the meaning of these options and tried to clarify their descriptions.

### DIFF
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -1266,10 +1266,10 @@ This is useful to determine whether a particular file is being searched or not.
 }
 
 fn flag_files_with_matches(args: &mut Vec<RGArg>) {
-    const SHORT: &str = "Only print the paths with at least one match.";
+    const SHORT: &str = "Print the paths with at least one match.";
     const LONG: &str = long!(
         "\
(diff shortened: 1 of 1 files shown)


crates/core/app.rs
```diff
@@ -1283,10 +1283,10 @@ This overrides --files-without-match.
 }
 
 fn flag_files_without_match(args: &mut Vec<RGArg>) {
-    const SHORT: &str = "Only print the paths that contain zero matches.";
+    const SHORT: &str = "Print the paths that contain zero matches.";
     const LONG: &str = long!(
         "\
-Only print the paths that contain zero matches. This inverts/negates the
+Print the paths that contain zero matches and suppress match contents. This inverts/negates the
```
BurntSushi, line 1289: Could you please wrap this to 79 columns inclusive?

BurntSushi (changes requested):
The wording LGTM, thanks! Just one style nit.

BurntSushi (approved):
Thanks!

PR #1871, opened 2021-05-26
core: Add field-match-separator and field-context-separator

### LINKED ISSUE #1842: Add option to customize the separator character betwen paths, lines, columns and matches
As of version 12.1.1, if the `--no-heading` option is used then a `:` will be used to separate paths, lines, columns (if the `--column` option is used) and matches. Is it possible to customize those characters individually?

For example I'd like to use a `@` between paths and lines, the default `:` between lines and columns and a space between columns and matches.

I'm new to ripgrep, I've read the manual but I haven't found anything about it so I assume this behaviour is not currently supported?

Anthuang:
This allows for the customization of the match and context separators. The match separator defaults to `:`, while the context separator defaults to `-`. Both are limited to a single byte.

Closes #1842

### DIFF
--- a/complete/_rg
+++ b/complete/_rg
@@ -314,6 +314,8 @@ _rg() {
     "--no-config[don't load configuration files]"
     '(-0 --null)'{-0,--null}'[print NUL byte after file names]'
     '--path-separator=[specify path separator to use when printing file names]:separator'
+    '--field-match-separator=[specify field match separator to use when printing file names]:separator'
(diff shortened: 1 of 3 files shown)


BurntSushi:
Thanks! I'm bringing this in via my rollup branch. I added tests and also removed the one byte restriction. The printer itself supports any number of bytes (including zero), so we might as well allow that.

PR #1881, opened 2021-05-30
Some sentences improve.

varikvalefor:
An unnecessary personal pronoun is removed.
A somewhat clunky clause is re-written.

### DIFF
--- a/doc/rg.1.txt.tpl
+++ b/doc/rg.1.txt.tpl
@@ -27,7 +27,7 @@ Synopsis
 
 DESCRIPTION
 -----------
-ripgrep (rg) recursively searches your current directory for a regex pattern.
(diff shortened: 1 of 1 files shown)


doc/rg.1.txt.tpl
```diff
@@ -82,7 +82,7 @@ _PATH_::
 
 OPTIONS
 -------
-Note that for many options, there exist flags to disable them. In some cases,
+Note that many options can be disabled through flags. In some cases,
```
iago-lito, line 85: `with`/`via` flags?

varikvalefor: "[W]ith flags" and "via flags" are also acceptable.  "[V]ia flags" is probably the best choice.

BurntSushi (approved):

PR #1882, opened 2021-05-31
Expand docs for GlobBuilder::literal_separator

sourcefrog:


BurntSushi (approved):

sourcefrog:
Actually, I'm sorry, I think I had the logic backwards: the default is that `*` does match `/` (which was confusing to me.) PTAL?

PR #1883, opened 2021-05-31
impl Default for GlobSet

sourcefrog:
This makes it easier to use in structs that derive Default

The empty value seems like a reasonable default "zero" value?


sourcefrog:
Fixed rustfmt, sorry.

PR #1886, opened 2021-06-01
Add config.ru and *.rbw to --type ruby

BuonOmo:
- https://stackoverflow.com/questions/55072891/what-is-this-file-config-ru-and-what-is-it-for
- https://fileinfo.com/extension/rbw

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -173,7 +173,12 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("red", &["*.r", "*.red", "*.reds"]),
     ("robot", &["*.robot"]),
     ("rst", &["*.rst"]),
-    ("ruby", &["Gemfile", "*.gemspec", ".irbrc", "Rakefile", "*.rb"]),
+    ("ruby", &[
+        // Idiomatic files
+        "config.ru", "Gemfile", ".irbrc", "Rakefile",
+        // Extensions
+        "*.gemspec", "*.rb", "*.rbw"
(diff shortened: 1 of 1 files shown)


crates/ignore/src/default_types.rs
```diff
@@ -173,7 +173,7 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("red", &["*.r", "*.red", "*.reds"]),
     ("robot", &["*.robot"]),
     ("rst", &["*.rst"]),
-    ("ruby", &["Gemfile", "*.gemspec", ".irbrc", "Rakefile", "*.rb"]),
+    ("ruby", &["config.ru", "Gemfile", "*.gemspec", ".irbrc", "Rakefile", "*.rb", "*.rbw"]),
```
BurntSushi, line 176: Thanks! Could you wrap this to 79 columns (inclusive) like other long lines in this file?

BurntSushi (changes requested):

crates/ignore/src/default_types.rs
BuonOmo: @BurntSushi of course :)

Thanks for being so reactive!

BurntSushi (approved):
Thanks!

BuonOmo:
how soon should I expect a new version ?

BurntSushi:
https://github.com/BurntSushi/ripgrep/blob/master/FAQ.md#release

It does just so happen that I am in the middle of getting out a release now. But I do not put out releases for things like this. Especially when you can add your own file types.

PR #1887, opened 2021-06-02
edition: migrate to Rust 2018

BurntSushi:
This was long overdue. Essentially, we apply `cargo fix --edition`, then add `edition = "2018"` to every `Cargo.toml`, then run `cargo fix --edition --edition-idioms` and then finally do manual touch-ups (mostly to remove uses of `extern crate`).


iago-lito:
Great ! Now soon 2021 ^ ^"

BurntSushi:
Yeah, hopefully more quickly than Rust 2018. I didn't wait this long intentionally or anything, just wasn't particularly motivated to do it. It's likely that the move to Rust 2021 will be much less eventful though.

PR #1895, opened 2021-06-15
ignore/types: name aliases for file types

### LINKED ISSUE #1857: python as alias for py file type
While I know type aliases would maybe best be avoided, maybe they can smooth out the user interface.

Currently there is a bit of inconsistency:

- `-t rust` for *.rs files
- `-t py` for *.py files

Naturally I'd reach for `-t python` if I didn't double check the type list - allowing both python and py to mean the same file type would smooth this out. I'm not sure how to avoid an explosion of aliases, but allowing the "full name" like python seems like a good idea on its own.

kotborealis:
Closes #1857

* Changed `ignore/types/default_types.rs` to allow array of names, which allows aliases — like `py==python` — without specifying globs twice.
* Updated `ignore/types/types.rs` to use name aliases
* Updated tests in `ignore/types/types.rs`


matkoniecz (approved):
note: I lack Rust knowledge and have not run this code. But from looking at diff it looks good and wanted to add thank for implementing this.

Having `ruby` type and `py` type without aliases is very confusing for someone new.

kotborealis:
Well, almost year and a half later I've managed to remember about this PR.
I have rebased my work, solved conflicts and tested it once more. Should we finally merge this one? 😅

PR #1903, opened 2021-06-18
Fix compilation with serde feature enabled

glebpom:
The bug was introduced during the migration to the 2018 edition


BurntSushi:
This is on crates.io in `globset 0.4.8`.

PR #1904, opened 2021-06-20
Correct types for crystal

kostya:
Instead of Projectfile used shard.yml (since some years). Also add ecr, crystal template type.


BurntSushi:
@kostya I would like to be careful about removing things from a list like this. For example, maybe there are still some people using `Projectfile`. And in particular, it doesn't seem like too much trouble to just leave it in the list. What do you think?

kostya:
updated

BurntSushi (approved):
Great, thanks!

Issue #1909, opened 2021-06-23
`ignore`: gitignore relative paths are not treated relatively


SergioBenitez:
In short, gitignore rules that should be relative to a specific gitgnore path are instead reinterpreted as `**/rule`. Here's the failing test case:

```rust
use std::path::PathBuf;
use ignore::gitignore::GitignoreBuilder;

fn main() {
    let mut builder = GitignoreBuilder::new("/root");
    builder.add_line(Some(PathBuf::from("/root/scratch/.gitignore")), "!foo.html").unwrap();
    let matcher = dbg!(builder.build().unwrap());

    assert!(matcher.matched("/root/scratch/foo.html", false).is_whitelist());
    assert!(dbg!(matcher.matched("/root/unknown/foo.html", false)).is_none());
}
```

The relevant output:

```
[src/main.rs:7] builder.build().unwrap() = Gitignore {
    root: "/root",
    globs: [
        Glob {
            from: Some(
                "/root/scratch/.gitignore",
            ),
            original: "!foo.html",
            actual: "**/foo.html",
            is_whitelist: true,
            is_only_dir: false,
        },
    ],
    num_ignores: 0,
    num_whitelists: 1,
}
[src/main.rs:10] matcher.matched("/root/unknown/foo.html", false) = Whitelist(
    Glob {
        from: Some(
            "/root/scratch/.gitignore",
        ),
        original: "!foo.html",
        actual: "**/foo.html",
        is_whitelist: true,
        is_only_dir: false,
    },
)
thread 'main' panicked at 'assertion failed: dbg!(matcher.matched(\"/root/unknown/foo.html\", false)).is_none()', src/main.rs:10:5
```

BurntSushi:
I believe these are the relevant gitignore docs:

> If there is a separator at the beginning or middle (or both) of the pattern, then the pattern is relative to the directory level of the particular .gitignore file itself. Otherwise the pattern may also match at any level below the .gitignore level.

In particular, the "Otherwise ..." language applies here since there is no `/` in your pattern. The issue here is that it should _only_ apply to any level _below_ that `.gitignore` file, but here, it is also being applied (incorrectly) to a sibling directory.

[...]

PR #1910, opened 2021-06-24
globset: make 'log' an optional dep

SergioBenitez:
As in the title. Would be nice to avoid the dependency (and the transitive dependency on `cfg-if`) when `log` isn't being used. Made it a default so logs remain when updating the crate.

As an aside, does the `simd-accel` feature do anything in `globset`?

### DIFF
--- a/crates/globset/src/lib.rs
+++ b/crates/globset/src/lib.rs
@@ -125,6 +125,16 @@ mod pathutil;
 #[cfg(feature = "serde1")]
 mod serde_impl;
 
+#[cfg(feature = "log")]
+macro_rules! debug {
+    ($($token:tt)*) => (::log::debug!($($token)*);)
+}
+
+#[cfg(not(feature = "log"))]
+macro_rules! debug {
+    ($($token:tt)*) => {};
+}
+
 /// Represents an error that can occur when parsing a glob pattern.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct Error {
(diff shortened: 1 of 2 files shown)


BurntSushi:
I'm not so sure about this. `globset` isn't exactly a dependency-slim crate itself. `log` is itself small and ubiquitous. Where is `globset` being used that `log` isn't?

As a downside, this infects every logging statement to the point of requiring conditional compilation. We could define our own logging macros to paper over this, but the juice really doesn't seem worth the squeeze given the number of other deps that `globset` already has.

> As an aside, does the `simd-accel` feature do anything in `globset`?

No. It used to. I probably should have added a note saying that it was left behind for backwards compatibility. `globset` hasn't had a semver bump in quite some time.

SergioBenitez:
> I'm not so sure about this. `globset` isn't exactly a dependency-slim crate itself. `log` is itself small and ubiquitous. Where is `globset` being used that `log` isn't?

In an application/library (the latter to resolve #1909) that I'm working on, of course.

> As a downside, this infects every logging statement to the point of requiring conditional compilation. We could define our own logging macros to paper over this, but the juice really doesn't seem worth the squeeze given the number of other deps that `globset` already has.

It's true, but it doesn't seem like log statements have seen much action for a while, so I can't imagine this will be much of a maintenance burden. Happy to add a macro that hides the conditional compilation, however.

SergioBenitez:
> Happy to add a macro that hides the conditional compilation, however.

Just did this. I really want to see this merged. I see zero downside with a 25% reduction in the number of deps.

crates/globset/src/lib.rs
```diff
@@ -125,6 +125,16 @@ mod pathutil;
 #[cfg(feature = "serde1")]
 mod serde_impl;
 
+#[cfg(feature = "log")]
+macro_rules! debug_log {
```
BurntSushi, line 129: Can we just call this `debug!`?

BurntSushi (changes requested):
LGTM after a possible name change and a `rustfmt` run.

SergioBenitez:
Done!

BurntSushi (approved):

BurntSushi:
This PR is in `globset 0.4.9` on crates.io.

PR #1919, opened 2021-06-30
Add CUDA file extensions to default types

### LINKED ISSUE #1918: Add support for cuda files
CUDA programming is usually done within .cu and .cuh files, though it can also include the usual c/c++ file extensions.  Please add "cuda" as another one of the ripgrep types.

eyal0:
This fixes #1918


BurntSushi (approved):
Thanks!

In the future, there's no need to create a separate issue for file types.

eyal0:
> Thanks!
> 
> In the future, there's no need to create a separate issue for file types.

I like to be organized!

Issue #1790, opened 2021-01-29
Stop on first non-match after positive match


sveniu:
### Feature request summary

Stop search on first non-matching line, but only after having seen matching lines.

### Motivation

On sorted or semi-sorted files, it is often useful to extract consecutive lines that match a given pattern. In some cases involving large files, time and IO can be saved by stopping the search after having matched such a run of consecutive lines.

I'm fairly convinced this feature can save countless CPU cycles and IO in big-ish data situations, particularly where timestamp matching is involved.

### Example

Input:
```
2021-01-29T07:08:09+0100 foo
2021-01-29T08:08:09+0100 bar
2021-01-29T09:08:09+0100 baz
2021-01-29T10:08:09+0100 boz
```

Behaviour when searching for pattern `^2021-01-29T08` with this option enabled:

1. First line does not match; continue to next.
2. Second line matches; print it and continue to next.
3. Third line does not match; break and exit; line 4 is never scanned.

### Pseudo code

```
have_seen_matches = null
for each line:
  if line matches:
    print line
    have_seen_matches = true
  else:
    if have_seen_matches == true:
      break
```

### Suggested documentation
```
--stop-on-nonmatch
    Stop reading a file on the first non-matching line that follows a matching
    line. At least one line must have matched for this to take effect.
```

BurntSushi:
Why doesn't the `--max-count` flag achieve what you need here?

sveniu:
> Why doesn't the `--max-count` flag achieve what you need here?

Because the number of matching lines is not known when starting the search. I might be misunderstanding what you actually mean, or my description or use case was unclear.

PR #1930, opened 2021-07-08
New option: Stop on non-match after a match

edoardopirovano:
Closes https://github.com/BurntSushi/ripgrep/issues/1790

This implements the option suggested in the linked issue. In particular, it adds a `--stop-on-nonmatch` flag that will stop reading a file once a non-matching line is found after a matching one. This is useful if, for example, a sorted file is being searched and the pattern being matched on is at the start of the line (so it is expected all matches will be adjacent).

### DIFF
--- a/complete/_rg
+++ b/complete/_rg
@@ -319,6 +319,7 @@ _rg() {
     '(-q --quiet)'{-q,--quiet}'[suppress normal output]'
     '--regex-size-limit=[specify upper size limit of compiled regex]:regex size (bytes)'
     '*'{-u,--unrestricted}'[reduce level of "smart" searching]'
+    '--stop-on-nonmatch[stop on first non-matching line after a matching one]'
 
     + operand # Operands
     '(--files --type-list file regexp)1: :_guard "^-*" pattern'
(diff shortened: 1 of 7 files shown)


edoardopirovano:
Hi @BurntSushi! No particular rush, but just checking you've seen this PR? Let me know if there's any changes you'd like me to make 🙂

BurntSushi:
No, I haven't looked at it yet. Unfortunately, my time is very limited and I tend to do PR reviews in batches in order to minimize context switching. It might be a while.

complete/_rg
```diff
@@ -319,6 +319,7 @@ _rg() {
     '(-q --quiet)'{-q,--quiet}'[suppress normal output]'
     '--regex-size-limit=[specify upper size limit of compiled regex]:regex size (bytes)'
     '*'{-u,--unrestricted}'[reduce level of "smart" searching]'
+    '--stop-on-nonmatch[stop on first non-matching line after a matching one]'
```
lespea, line 322: Isn't this missing the `=` sign like the rest of them?

BurntSushi: No. The `=` sign is only used for flags that accept an argument.

edoardopirovano: I think this shouldn't have an `=` because the option does not take any arguments. See for example the `--debug` option further up which also does not take any arguments.

PR #1940, opened 2021-07-19
Reduce `.crate` file size from 272 KB to 168 KB

Svetlitski:
The `ripgrep` crate currently includes test data, benchmark results, continuous integration configuration, and a few other miscellaneous items in its `.crate` file, bloating its size significantly ([crates.io](https://crates.io/crates/ripgrep) shows it presently weighs in at 272 KB). The changes in this PR ensure that these files are excluded from the `.crate` file going forward.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,14 @@ repository = "https://github.com/BurntSushi/ripgrep"
 keywords = ["regex", "grep", "egrep", "search", "pattern"]
 categories = ["command-line-utilities", "text-processing"]
 license = "Unlicense OR MIT"
-exclude = ["HomebrewFormula"]
+exclude = [
+  "HomebrewFormula",
+  "/.github/",
+  "/ci/",
+  "/pkg/",
+  "/benchsuite/",
+  "/scripts/",
+]
 build = "build.rs"
 autotests = false
 edition = "2018"


Cargo.toml
```diff
@@ -13,7 +13,7 @@ repository = "https://github.com/BurntSushi/ripgrep"
 keywords = ["regex", "grep", "egrep", "search", "pattern"]
 categories = ["command-line-utilities", "text-processing"]
 license = "Unlicense OR MIT"
-exclude = ["HomebrewFormula"]
+exclude = ["HomebrewFormula", "/.github/", "/ci/", "/pkg/", "/tests/data/", "/benchsuite/", "/scripts/"]
```
BurntSushi, line 16: Could you make sure we don't exceed 79 columns here? e.g.,

```toml
exclude = [
    "foo",
    "bar",
    ...
]
```

Also, please drop `/tests/data` from this list. Dropping that means you can't run tests using just the crate download.

Ignoring the rest looks sensible though. Thanks!

BurntSushi (changes requested):

Svetlitski:
Requested changes have been made. The `.crate` file is now 168 KB if you were curious.

Svetlitski:
@BurntSushi Bump on this. I believe I made all the requested changes. Would appreciate a review when you find the time.

BurntSushi (approved):
Thanks!

PR #1942, opened 2021-07-21
Update installation instructions using binary for 13.0.0

orfeasa:


BurntSushi:
Thanks. Looks like this was fixed elsewhere at some point.

Issue #607, opened 2017-09-19
Add option to disable regex search and search only for exact matches


hauleth:
This option could also accept `--smart-case` and `--ignore-case` options with would allow to case independent searches.

BurntSushi:
@hauleth Why doesn't the `-F/--fixed-literals` flag, which is named after the corresponding flag in `grep`, work for your use case?

hauleth:
Oh, I've must missed that option. Sorry for this fuss.

cedricvanrompay:
I would suggest to add this to the FAQ or to add a section for it in the user guide because it's pretty hard to find in the user guide right now when you don't know the name of the option already.

viggy28:
I didn't find `--fixed-literals`, rather `--fixed-strings` is available.

BurntSushi:
Yes, I misspoke. I corrected my comment.

guanghechen:
Hi, I just wonder if there is any workaround to search strings contain `\n` with the --fixed-strings option specified like the vscode did?

BurntSushi:
@guanghechen Please ask a question via Discussions: https://github.com/BurntSushi/ripgrep/discussions

Please show an example of what you want. Please show what you've tried and the output of what happened when you did.

PR #1945, opened 2021-07-22
Add -F/--fixed-strings to the User Guide's Common options

LordFlashmeow:
#607 is the top result for the search "ripgrep disable regex". I think it makes sense to add it to the user guide, since it's a very useful flag.


BurntSushi (approved):
I agree! Thank you. :-)

PR #1960, opened 2021-08-02
Add dependabot config for cargo updates

dbast:
Creates update PRs similar to https://github.com/dbast/ripgrep/pulls with history etc for review.

PRs can be merged if CI is green or closed to ignore the dependency regarding auto updates.


BurntSushi:
Thanks for the thought, but I do not want this. For a variety of reasons, but the principle one being that it would drive me crazy.

PR #1968, opened 2021-08-10
Lock stdout before printing an error message to stderr

### LINKED ISSUE #1941: result and error output mixing up

```
% rg --version
ripgrep 13.0.0
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

Arch Linux

Arch Linux

#### Describe your bug.

Searching in a folder that contains files/folders that I don't have permissions to causes the errors to mix up with the results.

You need to search in a folder with some files that you can't access (like /etc), that also contains some files that match your search.

```
% rg pcp /etc 
/etc/shadow: Permission denied (os error 13)
/etc/gshadow: Permission denied (os error /etc/passwd
39:pcp:x:959:959:Performance Co-Pilot:/var/lib/pcp:/usr/bin/nologin
13)

/etc/shadow-/etc/group
72:pcp:x:959:
: Permission denied (os error 13)
/etc/gshadow-: Permission denied (os error 13)
/etc/named.conf: Permission denied (os error 13)
/etc/libaudit.conf: Permission denied (os error 13)
/etc/sudoers.pacnew: Permission denied (os error 13)
/etc/shadow.pacnew: Permission denied (os error 13)
/etc/crypttab: Permission denied (os error 13)
```

errors should likely be sent to the main thread for printing so they are in-order:

```
% rg pcp /etc 
/etc/shadow: Permission denied (os error 13)
/etc/gshadow: Permission denied (os error 13)
/etc/passwd
39:pcp:x:959:959:Performance Co-Pilot:/var/lib/pcp:/usr/bin/nologin

/etc/group
72:pcp:x:959:
/etc/shadow-: Permission denied (os error 13)
/etc/gshadow-: Permission denied (os error 13)
/etc/named.conf: Permission denied (os error 13)
/etc/libaudit.conf: Permission denied (os error 13)
/etc/sudoers.pacnew: Permission denied (os error 13)
/etc/shadow.pacnew: Permission denied (os error 13)
/etc/crypttab: Permission denied (os error 13)
```

film42:
This avoids interleaving error messages messages when search is writing to stdout.

Closes #1941

There are probably a few other places we'd want to make a similar patch.

### DIFF
--- a/crates/core/logger.rs
+++ b/crates/core/logger.rs
@@ -33,7 +33,7 @@ impl Log for Logger {
     fn log(&self, record: &log::Record<'_>) {
         match (record.file(), record.line()) {
             (Some(file), Some(line)) => {
-                eprintln!(
+                eprintln_locked!(
                     "{}|{}|{}:{}: {}",
                     record.level(),
                     record.target(),
@@ -43,7 +43,7 @@ impl Log for Logger {
                 );
             }
             (Some(file), None) => {
-                eprintln!(
+                eprintln_locked!(
                     "{}|{}|{}: {}",
                     record.level(),
                     record.target(),
@@ -52,7 +52,7 @@ impl Log for Logger {
                 );
             }
             _ => {
-                eprintln!(
+                eprintln_locked!(
                     "{}|{}: {}",
                     record.level(),
                     record.target(),
@@ -63,6 +63,6 @@ impl Log for Logger {
     }
 
     fn flush(&self) {
-        // We use eprintln! which is flushed on every call.
+        // We use eprintln_locked! which is flushed on every call.
     }
 }
--- a/crates/core/main.rs
+++ b/crates/core/main.rs
@@ -47,7 +47,7 @@ type Result<T> = ::std::result::Result<T, Box<dyn error::Error>>;
 
 fn main() {
     if let Err(err) = Args::parse().and_then(try_main) {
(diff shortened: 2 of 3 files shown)


BurntSushi:
> There are probably a few other places we'd want to make a similar patch.

I believe the only other place is in the logger? Maybe it would make sense to write our own `eprintln_locked` macro that does this, and then just replace every call to `eprintln!` in `crates/core` with that new macro.

I think this patch is right, but it's an unfortunate abstraction violation. This only works because the buffer printer in `search_parallel` will acquire the same stdout lock, which is done inside of `termcolor`. To avoid the abstraction violation, I think we'd need _another_ mutex in `crates/core` itself, but I'm not so sure it's worth it.

To mitigate the abstraction violation, could you make the comment added here mention and acknowledge the abstraction violation? Basically what I said above. And if we put this in its own macro, we only need to write this comment once.

film42:
Adding a new macro certainly cleans things up a lot. I added a comment in the code (but not as part of the public documentation) since a comment about an abstraction violation may not be useful to the user outside of the code. Happy to make that part of the public docs though.

In my first cut, I added a lazy static rwlock and mapped the `read` to `STDOUT` with `write` mapping to `STDERR` to keep logging decently quick even with a mutex, but ultimately, locking `STDOUT` was a lot cleaner and intuitive. It also accomplishes the same thing.

PR #1969, opened 2021-08-11
doc: improve documentation for --quiet flag

### LINKED ISSUE #1962: Documentation for '--quiet' is misleading
The documentation for the `--quiet` flag is misleading with the `--files` flag.

The current documentation states:

```
       -q, --quiet
           Do not print anything to stdout. If a match is found in a file,
           then ripgrep will stop searching. This is useful when ripgrep is
           used only for its exit code (which will be an error if no matches
           are found).

           When --files is used, then ripgrep will stop finding files after
           finding the first file that matches all ignore rules.
```

The very last part, "the first file that matches all ignore rules" seems like it's worded backwards.  I get what it means, it's just oddly worded.  It might be better to state instead "the first file that does not match any ignore rules".

Just an opinion, feel free to close if you disagree.

```
rg --version
ripgrep 13.0.0
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

Vinegret43:
Closes #1962

Fix oddly worded sentence in the last paragraph

### DIFF
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -2584,7 +2584,7 @@ will stop searching. This is useful when ripgrep is used only for its exit
 code (which will be an error if no matches are found).
 
 When --files is used, then ripgrep will stop finding files after finding the
-first file that matches all ignore rules.
+first file that does not match any ignore rules.
 "
     );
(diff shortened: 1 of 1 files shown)


BurntSushi:
Looks like this was superseded by #2462.

PR #1971, opened 2021-08-12
Add vimrc to vim and vimscript types

keith:
This adds `*vimrc` to the vim and vimscript file types. Often this will
match `~/.vimrc` but given how unambiguous this is I think matching
more potential options like `~/.ideavimrc` (Intellij's vim plugin
configuration file) and others makes sense.


keith:
https://github.com/BurntSushi/ripgrep/commit/ba535fb5a34f6d703fc652e8dfc23c0fed78fc74

PR #1972, opened 2021-08-14
Minor cleanups

a1346054:


BurntSushi:
This looks like a weird PR. There might be one or two things here that seem okay, but most of it looks either wrong or just changing things for the sake of changing them.

a1346054:
I don't remember what tooling I used 2 years ago, but these days `shellcheck` detects some of the script problems (they always need to be reviewed by a human) and `shfmt` gives guidance on unified coding style for the scripts.

The trailing whitespace removal should be self-explanatory.

Issue #442, opened 2017-04-09
rg --help | less assumes the terminal has 100 columns


bluss:
If we pipe rg's output to something, it assumes the terminal has 100 columns.

For example rg --help, which has output that is long and begs for pagination.

To reproduce: Install rg. Adjust terminal to a width of for example 90 columns. Use `rg --help | less`  (with no shell aliases or anything in between).

(Sorry! rg --help is now fixed, so the bug report now had its schedule unblocked)

BurntSushi:
ripgrep doesn't do anything explicitly smart with terminal size, so I assume this is a clap thing.

BurntSushi:
I could have sworn clap had some setting or knob for this, but I can't find it after a quick skim.

kbknapp:
There's [`App::max_term_width`](https://docs.rs/clap/2.23.1/clap/struct.App.html#method.max_term_width) and [`App::set_term_width`](https://docs.rs/clap/2.23.1/clap/struct.App.html#method.set_term_width)

`max_term_width` will allow freeform wrapping at whatever the user terminal width is, *up to* some maximum. This is what ripgrep uses currently (at 100).

`set_term_width` sets the virtual width and always wraps at that column width.

kbknapp:
[I hit comment too early]

Not using either will just wrap at whatever the user terminal width happens to be.

BurntSushi:
@kbknapp Oh interesting. Welp. I was clearly wrong about ripgrep not doing anything.

With that said, I can't seem to cause clap to do the right thing here. Namely, if `max_term_width` is set to `100`, then shouldn't it wrap text at `N < 100` after I shrink my terminal? (I also just tried `set_term_width`, which has the same behavior.)

If I remove the `max_term_width`/`set_term_width` setting completely, then the help text sprawls across my entire screen, which I suspect is why I set it in the first place.

BurntSushi:
I see. clap does do the right thing *unless* its output is piped to `less`. So presumably that interferes with `clap` figuring out the terminal width?

kbknapp:
Interesting, my initial thinking is `less` is being detected a terminal without a width (or rather `stdout` not being a terminal) which is causing the fallback to 100.

BurntSushi:
[...]

PR #1991, opened 2021-09-16
Reflow help text to fit within 80 columns

mattiase:
The `--help` text was unnecessarily hard to read in an 80 column terminal window (which is the default on many systems, for reasons good and bad). This change reflows the paragraphs so that they fit.

As the text is printed with 12 spaces indentation, the maximum paragraph line length is 68 characters. Some people prefer not using the last (80th) column because older terminals may have trouble showing such lines without wrapping; I didn't care about that but if you do, I'll readjust.

Similarly, there are good reasons for using even shorter text lines in prose; `man` leaves a few more columns empty at the right margin for reasons of legibility and aesthetics. That might be a good idea here, too.

I don't think full justification in monospaced text is, well, justified but again some people like it -- do tell.

### DIFF
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -651,15 +651,17 @@ fn arg_pattern(args: &mut Vec<RGArg>) {
     const SHORT: &str = "A regular expression used for searching.";
     const LONG: &str = long!(
         "\
-A regular expression used for searching. To match a pattern beginning with a
-dash, use the -e/--regexp flag.
+A regular expression used for searching. To match a pattern
+beginning with a dash, use the -e/--regexp flag.
 
-For example, to search for the literal '-foo', you can use this flag:
+For example, to search for the literal '-foo', you can use
+this flag:
 
     rg -e -foo
 
-You can also use the special '--' delimiter to indicate that no more flags
-will be provided. Namely, the following is equivalent to the above:
+You can also use the special '--' delimiter to indicate that no more
+flags will be provided. Namely, the following is equivalent to
+the above:
 
     rg -- -foo
 "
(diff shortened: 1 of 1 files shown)


mattiase:
It appears that the `-h` help text would benefit from a similar treatment but before I extend the PR to handle that, it would be good to know that we agree on the basics to avoid wasting your time and mine.

BurntSushi:
I think clap is supposed to do this automatically.

Either way, I'd prefer we not strongly couple---where reasonable---the line length limits shown in the `--help` output with the line length limits in the source code. This, for example, is something I'd guess would be difficult to maintain. It's already hard to get people to adhere to 79 columns, even though the surrounding code/text clearly respects it. This is just going to be a nightmare to enforce and basically impossible to see clearly in PR review.

mattiase:
> I think clap is supposed to do this automatically.

Right, and I agree it's better than fixed formatting. But clap doesn't work properly without the `wrap_help` feature; see the bottom comments of issue #442.
I build ripgrep with `wrap_help` added to `Cargo.toml` locally and it works. Shall I open a new PR? It's only a one-liner, so perhaps you prefer doing it yourself.

BurntSushi:
Yes, but that adds a new dependency on `term_size` which appears deprecated.

I was thinking about this: https://github.com/BurntSushi/ripgrep/blob/f4d07b9cbdcf7d52e63f35fe76279e99a17492ef/crates/core/app.rs#L63

But it's set to 100.

I am currently thinking about migrating away from clap to lexopt, and if and when I do that, I'll reconsider this. I just know that the approach taken in this PR isn't going to work for me.

mattiase:
We could also add backslashes to the end of lines in doc strings, so that clap's attempt to word-wrap actually works. It's actually already done in places but not very consistently.

It's a bit of work but since I already did it in my tree, there's no extra cost. New PR?

BurntSushi:
That doesn't sound like a good idea to me at all. As I said, the strings in the code should be optimized for reading and writing in the code. Adding backslashes everywhere sounds bad for both reading and writing.

Whatever solution is used here needs to leave the doc strings in the code as they are.

mattiase:
Right. Then new text filling code is needed, that uses blank lines to detect end of paragraphs, perhaps along with some rudimentary mark-up to force line breaks.

There are a couple of places where an indented block is used (such as the warning paragraph in the doc string for `--pre`); some mark-up would be needed there too, to distinguish it from pre-formatted text. And there are some tables.

These could all be treated as preformatted to simplify the text filler, but then their strings have to be formatted by hand to fit some typical narrowest case (probably 80 columns).

All doable.

PR #2009, opened 2021-09-29
moving .ts from typoscript to TypeScript

Mellbourn:
The TypeScript language is in the top ten of the [most popular programming languages](https://pypl.github.io/PYPL.html).

Typoscript is a custom language for the TYPO3 CMS, that is far less used.

I think that the `.ts` extension should belong to the `typescript` type in ripgrep, not typoscript.


BurntSushi:
The `ts` file type already exists. That's for typescript.

File types don't own extensions. Multiple file types may have the same globs.

Mellbourn:
Right you are. It was a bit confusing that I could start to type `typ` in the shell and the completion system suggested `typoscript`, which I initially thought was a typo :D

matkoniecz:
So 

```
    ("typescript", &["*.ts"]),
    ("typoscript", &["*.ts", "*.typoscript"]),
```

would be better, right?

BurntSushi:
This is wrong. Firstly, if `ts` and `typescript` are referring to the same thing, then they should have the same globs. This PR does not have that. Secondly, this PR removes a glob from `typoscript` without justifying it.

PR #2049, opened 2021-11-04
pkg: embed windows manifest to support long file paths if enabled

### LINKED ISSUE #364: long file name support by default?
Seems long file names may not work in the Windows x64 build of ripgrep 0.4.0.

In cmd on windows 10 (Version 1607 OS Build 14393.693) do the following:

```batch 
cd c:\
mkdir ar
cd ar
echo "foobar" > abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrs
```

Execute rg:
```batch
rg "foobar"
```

Result:
`
./abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrs: The system cannot find the path specified. (os error 3)
`

Try again:
```batch
rg "foobar" \\?\%cd%
```

Result:
`
\\?\c:\ar\abcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrstuvwxyzabcdefghijklmnopqrs
1:"foobar"
`

It would be nice to have long file name support by default without having to add `\\?\%cd%` .

thargor:
fixes #364 

Implementation is based on https://gal.hagever.com/posts/windows-long-paths-in-rust/ .

This still requires long file paths to be enabled in the windows registry.


thargor:
this is also what https://github.com/nabijaczleweli/cargo-update is using

SimplyDanny:
Works fine for me with the registry key `LongPathsEnabled` set to 1.

build.rs
```diff
@@ -1,12 +1,12 @@
...
 use std::env;
 use std::fs::{self, File};
 use std::io::{self, Read, Write};
 use std::path::Path;
 use std::process;
 
-use clap::Shell;
-
 use app::{RGArg, RGArgKind};
+extern crate embed_resource;
```
SimplyDanny, line 9: This line is [no longer needed](https://doc.rust-lang.org/nightly/edition-guide/rust-2018/path-changes.html) since Rust Edition 2018.

build.rs
```diff
@@ -48,6 +48,9 @@ fn main() {
     if let Some(rev) = git_revision_hash() {
         println!("cargo:rustc-env=RIPGREP_BUILD_GIT_HASH={}", rev);
     }
+
+    // embed windows application manifest file
```
SimplyDanny, line 52: From what I see browsing through the code base, comments are written as sentences, i.e. they start with a capital letter and end with a period.

build.rs
```diff
@@ -1,12 +1,12 @@
+use clap::Shell;
```
BurntSushi, line 1: Why was this moved? Please put it back. :)

build.rs
BurntSushi: Indeed, this should be removed.

build.rs
BurntSushi: Yeah, that's preferred. But what would be better here is more elaboration. Mostly, this comment nearly repeats what the code itself says, without adding much. I think this comment should:

* Say why this is being done.
* Include instruction on how to take advantage of it. If you're someone who is reading this `build.rs` trying to understand it, and we say, "this enables long file paths" without saying that it also requires a registry edit, that might produce some confusion. So let's shine a light where we can.

rg-manifest.rc
```diff
@@ -0,0 +1,2 @@
+#define RT_MANIFEST 24
+1 RT_MANIFEST "rg.exe.manifest"
```
BurntSushi, line 2: I can't comment on it, but where did the `rg.exe.manifest` file come from? How was it generated?

Also, I think it would be better to put these files into a `windows` sub-directory, instead of polluting the root of the repo even more than it is.

BurntSushi: Oh I see, GitHub falsely detects it as a binary file, but it's just plain text. Phew. That's good. Yeah, I think just move these into a new `windows` sub-directory. Thanks!

build.rs
```diff
@@ -48,6 +48,9 @@ fn main() {
     if let Some(rev) = git_revision_hash() {
         println!("cargo:rustc-env=RIPGREP_BUILD_GIT_HASH={}", rev);
     }
+
+    // embed windows application manifest file
+    embed_resource::compile("rg-manifest.rc");
```
BurntSushi, line 53: What does this do on non-Windows targets? Even if nothing, it might be nice to gate this on `#[cfg(windows)]` to make it a bit clearer to those reading the code.

Cargo.toml
```diff
@@ -61,6 +61,7 @@ version = "0.3.0"
 
 [build-dependencies]
 lazy_static = "1.1.0"
+embed-resource = "1.6"
```
BurntSushi, line 64: I don't think this is needed for non-Windows builds right? Could you please make this a target specific dependency for windows only? Thanks!

BurntSushi (changes requested):
I got through most of this PR review before realizing that it's using a dependency that I will not bring in. I'm sorry. Someone tried this before in #1898. Specifically, see [my comment on that PR](https://github.com/BurntSushi/ripgrep/pull/1898#issuecomment-862466605).

thargor:
@BurntSushi @SimplyDanny thank you for your feedback!

Would an implementation based on https://crates.io/crates/windres (which is based on https://crates.io/crates/find-winsdk ) be acceptable for ripgrep?

BurntSushi:
As long as there's nothing written by Jonathan Blow in the dependency tree, the licenses are compatible and the crates are decently maintained. Remember, once this gets added to ripgrep, it's forever mine to maintain, evev if the dependencies go defunct.

So in short: probably, but I'll still need to do a review if a PR comes in.

t-rapp:
For anyone interested in long path support: Stumbled over this recent [pull request for rustc](https://github.com/rust-lang/rust/pull/96737) which adds a similar manifest file to the rustc executable. It apparently works for the *-msvc target by using a build script without external libraries.

BurntSushi:
@ChrisDenton what are your thoughts on this PR? Is this the right way to support long paths on Windows?

ChrisDenton:
Yep. I'd be minded to append `.xml` to the file name to make github happy. And while an xml declaration (`<?xml ...`) isn't strictly necessary, it can make other tools happy. Also linking to [the docs](https://learn.microsoft.com/en-us/windows/win32/sbscs/application-manifests#longpathaware) would make sense.

My only quibble is that it's a lot of dependencies for what can be [a couple of lines](https://github.com/rust-lang/rust/blob/ce519c5945c90596cf429dc2a91c846daeb75cd1/compiler/rustc/build.rs#L23-L24) in a build script (plus a few more lines to apply it conditionally). Though unfortunately windows-gnu does not support this.

BurntSushi:
@ChrisDenton Oh yessss! Thank you for that link. I will do what you did for `rustc`. That looks much better.

PR #2052, opened 2021-11-07
If require_git is false, don't stat .git

joshtriplett:
I've confirmed via strace that this eliminates a pile of stat calls.


BurntSushi (approved):
LGTM! I think this just has to be rustfmt'd. :)

joshtriplett:
rustfmt'd. :)

PR #2061, opened 2021-11-12
globset: add escaping mechanism

### LINKED ISSUE #2060: [globset] Create an escaped pattern from a literal string
Basically I'm migrating some code from `glob` and looking for a functional equivalent of https://docs.rs/glob/0.3.0/glob/struct.Pattern.html#method.escape. Maybe put is as `GlobBuilder::new_escaped` or something.

piegamesde:
Closes #2060. The implementation is taken 1:1 from the glob crate. Function signatures may be subject to bike shedding.

I feel that there may be some optimization potential in the `escape` method that doesn't allocate if there are no special characters to be escaped (probably by returning a `Cow`), but I would have to benchmark it and *creating* globs is usually not on the critical path.

I've seen some `Strategy` foo while skimming over the code. Will the crate automatically pick a `Literal` strategy or do I need to do something to activate it?

### DIFF
--- a/crates/globset/src/glob.rs
+++ b/crates/globset/src/glob.rs
@@ -137,6 +137,27 @@ impl GlobMatcher {
     }
 }
 
+impl PartialEq for GlobMatcher {
+    fn eq(&self, other: &GlobMatcher) -> bool {
+        self.pat == other.pat
+    }
+}
+
+impl Eq for GlobMatcher {
+}
+
+impl hash::Hash for GlobMatcher {
+    fn hash<H: hash::Hasher>(&self, state: &mut H) {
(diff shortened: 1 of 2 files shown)


piegamesde:
@BurntSushi friendly ping

BurntSushi:
Thanks! I had to trim this PR down quite a bit as you added a bunch of stuff beyond `pub fn escape(..) -> String` that I do not want.

And yes, `Literal` should automatically be selected. And if it isn't, we should fix that. But nothing else should care about escaping.

BurntSushi:
And yeah, definitely no need to optimize a routine like this. Your signature and implementation are more than sufficient.

PR #2079, opened 2021-11-21
transcoding is done in addition to search

matkoniecz:
Even if transcoding would be faster than search it would still incur performance penalty

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -648,7 +648,7 @@ given, which is the default:
   they correspond to a UTF-16 BOM, then ripgrep will transcode the contents of
   the file from UTF-16 to UTF-8, and then execute the search on the transcoded
   version of the file. (This incurs a performance penalty since transcoding
-  is slower than regex searching.) If the file contains invalid UTF-16, then
+  is needed in addition to regex searching.) If the file contains invalid UTF-16, then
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
Thanks!

BurntSushi:
I wish there were a way to have CI bleat about 79 column violations, but it's not a rule I enforce 100% strictly. e.g., URLs and sometimes even code going over 79 columns is okay when that's the clearer path. But alas, such cases call for human judgment.

matkoniecz:
Maybe some bot that would comment on PR "you added new lines in XXX file longer than N characters, typically this is unwanted, see this contributing.md section for more info"?

matkoniecz:
I just checked and I see no CONTRIBUTING.md file at all (if that docs exist somewhere then making link-only CONTRIBUTING.md is still a good idea to make easy to find them)

For bot itself - I have seen some code coverage monitor bots posting, Visual Studio Code has some bot actions on issue tracker (automatically and triggered). See https://github.com/microsoft/vscode/issues/137641#issuecomment-975481667 https://github.com/microsoft/vscode-github-triage-actions

But I have no direct experience with maintaining, making or using something like that.

BurntSushi:
Aye thanks, I created #2080 to track your idea.

PR #2103, opened 2021-12-16
add yum-utils for Red Hat & CentOS

strouja:
I also suggest removing the `$` in front of each command so end users can simply copy and paste the line without having to edit and remove the `$`
If you do not like my suggestions I won't be offended.

`yum-config-manager` will fail unless `yum-utils` is installed as `yum-config-manager` is installed with `yum-utils`.  If `yum-utils` is already installed then the `yum install -y yum-utils` is harmless

Great job with the tool and I love it.  Thanks for everything

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,71 +200,72 @@ If you're a **macOS Homebrew** or a **Linuxbrew** user, then you can install
 ripgrep from homebrew-core:
 
 ```
-$ brew install ripgrep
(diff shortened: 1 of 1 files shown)


BurntSushi:
I ended up just adding this in myself because I didn't want to bring in the changes to every other command too. Thanks for the tip!

PR #2107, opened 2021-12-20
ignore/gitignore: Support quoted INI values

Enzime:
Fix parsing `excludesFile` values that are quoted to match Git and [other INI parsers](https://en.wikipedia.org/wiki/INI_file#Quoted_values).


Enzime:
@BurntSushi any thoughts on this PR?

BurntSushi:
I think I'm going to pass on this. It permits mixed quotes and doesn't handle escaping. Maybe it's still better than the status quo, but I'd rather just keep things simple instead of trying to hackily improve it. Probably we should spend more effort trying to parse the INI file, but it isn't a priority for me right now.

PR #2114, opened 2022-01-02
Add 3rd-party wrappers section to readme

ElectricRCAircraftGuy:
Add "3rd-Party Wrappers" section to readme, and add link to the interactive `rgf` RipGrep Fuzzyfinder search tool wrapper I wrote.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,7 @@ Please see the [CHANGELOG](CHANGELOG.md) for a release history.
 * [Shell completions](FAQ.md#complete)
 * [Building](#building)
 * [Translations](#translations)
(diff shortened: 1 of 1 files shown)


ElectricRCAircraftGuy:
Any problems with merging this small readme change?

BurntSushi:
I think I'm going to pass on this one. I'm not sure it's worth doing, and a list with a single link to someone's dotfiles probably doesn't justify a whole new section to an already long README.

If we were going to do this, I'd rather see an FAQ question and the answer could link to one or more tools that appropriately answers the question. This way, it ties to the tool to a specific use case.

PR #2141, opened 2022-02-06
Add embedded puppet templates to the puppet type

nbarrientos:
.epp files are getting more and more common in Puppet code bases so it
makes sense I think to include them as part of the "puppet" type.

https://puppet.com/docs/puppet/7/lang_template_epp.html


nbarrientos:
Any chance this could be please merged? Thanks.

PR #2146, opened 2022-02-18
core: update clap to version 3.1.0

budde25:
This updates clap to version 3.1.0 following this migration guide
https://github.com/clap-rs/clap/releases/tag/v3.0.0

As far as I can tell it still works the same as before and it passes all the tests.
but the tricky part is the subtle changes section and I am unsure if it still follows all the same expected behavior as before


BurntSushi:
Thanks. I'll try to take a look at this soon.

Could you please get rid of merge commits? Thanks!

budde25:
Thanks, sure thing!
How would I go about doing that? I tried to do `git rebase -i master` without much luck.

BurntSushi:
Yes, rebase is the answer. Unfortunately, I don't have the time to mentor someone through that. Perhaps someone else can help?

BurntSushi:
I'm going to pass on this and stick with clap 2.x for now. I didn't do the upgrade to clap 3 in the first place because I knew clap 4 was just around the corner. And I'm now considering switching to lexopt.

In general it's a good idea to get sign off from maintainers before doing big changes/refactors like this, because I think I probably could have save you the effort.

Thank you for the work though. I appreciate it.

PR #2147, opened 2022-02-19
Fixing broken build and making tiny improvements

ink-splatters:
Hi there,

being long-term `ripgrep` admirer and everyday user, I've come up with those ultra-tiny improvements, besides fixing blocker build issue caused by `packed_simd_2` which had used now removed feature.

Update, however, has been done using `cargo update` hence rip previous `Cargo.lock`

Another thing is enabling lto: I've not profiled your tool yet, but basing on my recent experience with custom build of Python (with which I observed perf improvements, at least, in some synthetic benchmarks)[1], I consider this potentially to be
beneficial.

New profile has been created as lto badly slows down the linking, which may not be desirable by default.

Finally - rip warnings :) ( related to code considered dead by rust )

---
[1] FYI Python build details:
```shell
    PYTHON_CONFIGURE_OPTS="--enable-framework" CONFIGURE_OPTS="--enable-optimizations --with-lto" \
    CPPFLAGS="-I$(brew --prefix openssl)/include -I$(brew --prefix libffi)/include" \
    LDFLAGS="-L$(brew --prefix openssl)/lib -L$(brew --prefix libffi)/lib" \
pyenv install 3.9.2
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -30,15 +30,15 @@ checksum = "904dfeac50f3cdaba28fc6f57fdcddb75f49ed61346676a78c4ffe55877802fd"
 
 [[package]]
 name = "bitflags"
-version = "1.2.1"
+version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "cf1de2fe8c75bc145a2f577add951f8134889b4795d47466a54a5c846d691693"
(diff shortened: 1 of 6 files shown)


BurntSushi:
Thanks, but I think I'm going to pass on this for now.

I also generally prefer to make all `Cargo.lock` changes myself manually. And putting such changes in PRs like this makes them a nightmare to merge in my workflow. (Which is that I have bursts of activity where PRs tend to sit for a while and get stale.)

PR #2149, opened 2022-02-23
Attempt at a unified diff output format for use with --replace

sonohgong:
This is a draft for what could be a unified diff output for ripgrep when using the `--replace` option.

The chance for acceptance is low considering the feature request #74 was closed,
however this is an implementation of @samuelcolvin suggestion to produce a diff output format
instead of actually writing to files. The job of changing files can then be handed to something like `patch` or `git apply`.

I wanted to at least investigate what this would look like and managed to get something working,
so I hope this can be a way to discuss around some actual code. @BurntSushi stated they were
actively looking to moving more functionality into a library, so I'm not sure if this was totally dismissed.

I ran into a few problems with the replacement buffer which seems to contain more than the original match
and I managed to reproduce this with ripgrep 13.0.0, notes are in the code. I managed to get it working by
modifying the replacement, not sure but it could be related to #2095.
There is also no support for additional context, as that would require keeping track of it before and after a match
to then produce the correct hunk header with context included, and I'm not very familiar with the code or rust
to come up with a nice way to do that.

The implementation is mostly taken from the JSON printer with necessary parts replaced.
It can handle multiline replacements that differ in the amount of lines before/after replacement.

I've tested mostly with running the following in the ripgrep repo:
```
rg -U '\.A\n' --replace 'BB'
rg -U '.A\nB' --replace 'BB' UNLICENSE
```
as these also produced the spurious output I found with rg 13.0.0.

Outstanding stuff to leave draft status:
 - [ ] unit tests
 - [ ] better flag handling (--diff should somehow require you to use --replace)

### DIFF
--- a/crates/printer/src/util.rs
+++ b/crates/printer/src/util.rs
@@ -63,9 +63,13 @@ impl<M: Matcher> Replacer<M> {
         // See the giant comment in 'find_iter_at_in_context' below for why we
         // do this dance.
         let is_multi_line = searcher.multi_line_with_matcher(&matcher);
+        let mut extension_len = 0;
         if is_multi_line {
             if subject[range.end..].len() >= MAX_LOOK_AHEAD {
+                extension_len = MAX_LOOK_AHEAD;
                 subject = &subject[..range.end + MAX_LOOK_AHEAD];
+            } else {
+                extension_len = subject.len() - range.end;
             }
         } else {
             // When searching a single line, we should remove the line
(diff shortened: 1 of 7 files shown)


crates/printer/src/util.rs
```diff
@@ -102,6 +106,18 @@ impl<M: Matcher> Replacer<M> {
...
+
+            if is_multi_line {
+                // Remove the subject buffer beyond the range.end.
+                // NOTE: could this be a bug with the current replace functionality?
+                // As an example, running `rg -U '\.\nA' --replace 'BB' -C 2` in this repo produces
+                // spurious extra lines of output in the replacement, as the extra bytes where
+                // not removed. This seems to fix that, but it's not pretty and might be wrong.
+                dst.truncate(dst.len() - extension_len);
+            } else {
+                // Restore the line terminator.
```
sonohgong, line 118: This is probably wrong, but I'm relying on the original line terminators instead of reproducing it in the printer. I'm not sure how to fix it properly.

sonohgong: At least all tests seem to pass.

BurntSushi:
Thanks for working on this. But as discussed in #74, this really isn't something I want to support. It's the kind of feature that begs more features. I don't want it in ripgrep. I think search-and-replace is a complicated enough task that some other tool should try to solve it. (And indeed, several have. But the UX is difficult to get right I think. Yet another reason I don't want to touch this particular problem.)

PR #2150, opened 2022-02-24
search *.type if filetype not exist

gcxfd:
search *.type if filetype not exist


BurntSushi:
I'm sorry, but I don't agree with this behavior change. I would rather see explicit failure modes than trying to guess at the user's intent.

PR #2151, opened 2022-02-24
Limit amount of matches during replacement

sonohgong:
The buffer passed to the replace function can be bigger than the original match
and cause extra matches to occur during replacement. This adds an extra replace
function that doesn't allow matches beyond the original range to get replaced.

I've added `replace_with_captures_in_range` to do this, but I'm not sure if this should be fixed within
`replace_with_captures_at` itself (as that one isn't always used with a range).

Fixes #2095

### DIFF
--- a/crates/matcher/src/lib.rs
+++ b/crates/matcher/src/lib.rs
@@ -38,9 +38,11 @@ implementations.
 
 #![deny(missing_docs)]
 
+use std::cmp::min;
 use std::fmt;
 use std::io;
 use std::ops;
+use std::ops::Range;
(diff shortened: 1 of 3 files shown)


sonohgong:
Resolved by #2209

BurntSushi:
Ah I missed this PR! Thanks. I did end up taking a slight different approach by writing the new function as a new internal helper routine instead of making it a new public API in `grep-matcher`. Probably `grep-matcher` should evolve its `*_at` APIs to accept ranges instead of just the starting location. But that's a battle for another day.

PR #2188, opened 2022-04-19
Add "asp" for default type asp

MatrixDai:
The `*.asp` was included in the type "asp" when it was added.
https://github.com/BurntSushi/ripgrep/pull/1134

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -16,7 +16,7 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("asciidoc", &["*.adoc", "*.asc", "*.asciidoc"]),
     ("asm", &["*.asm", "*.s", "*.S"]),
     ("asp", &[
-        "*.aspx", "*.aspx.cs", "*.aspx.vb", "*.ascx", "*.ascx.cs", "*.ascx.vb",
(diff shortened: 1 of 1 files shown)


crates/ignore/src/default_types.rs
```diff
@@ -16,7 +16,7 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("asciidoc", &["*.adoc", "*.asc", "*.asciidoc"]),
     ("asm", &["*.asm", "*.s", "*.S"]),
     ("asp", &[
-        "*.aspx", "*.aspx.cs", "*.aspx.vb", "*.ascx", "*.ascx.cs", "*.ascx.vb",
+        "*.aspx", "*.aspx.cs", "*.aspx.vb", "*.ascx", "*.ascx.cs", "*.ascx.vb", "*.asp"
```
BurntSushi, line 19: Please follow the style of the surrounding code and make sure lines are no longer than 79 columns.

(It's hard to tell if this is longer than 79 columns in the GitHub UI, but it looks like it.)

BurntSushi (changes requested):

crates/ignore/src/default_types.rs
MatrixDai: Thank you, I was commit the change from Github online editor directly due to simple change. Forgot to check the line length, it's far exceed 79 already.
Now, fixed.

BurntSushi (approved):

PR #2192, opened 2022-04-23
ignore/types: Add "*.hh" to headers

cemeyer:
Like .hpp, .hh is an occasionally used extension for C++ headers
(to distinguish them from C headers).

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -80,7 +80,7 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("gradle", &["*.gradle"]),
     ("groovy", &["*.groovy", "*.gradle"]),
     ("gzip", &["*.gz", "*.tgz"]),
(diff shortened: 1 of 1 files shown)


BurntSushi:
Can you show me something that uses `.hh` for this purpose?

cemeyer:
Sure.  There are a handful (<10) of these in FreeBSD's src tree -- e.g., https://github.com/freebsd/freebsd-src/blob/main/sbin/devd/devd.hh .  Others:

```
$ find . -name \*.hh
./sbin/devd/devd.hh
./tests/sys/fs/fusefs/utils.hh
./tests/sys/fs/fusefs/mockfs.hh
./usr.bin/dtc/checking.hh
./usr.bin/dtc/dtb.hh
./usr.bin/dtc/fdt.hh
./usr.bin/dtc/input_buffer.hh
./usr.bin/dtc/util.hh
```

BurntSushi:
Lovely, thank you! TIL. :-)

PR #2203, opened 2022-05-09
Detect MODULE.bazel and WORKSPACE.bazel

keith:
MODULE.bazel is a new file, and WORKSPACE.bazel was always supported
similar to BUILD.bazel vs BUILD.

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -23,7 +23,8 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("avro", &["*.avdl", "*.avpr", "*.avsc"]),
     ("awk", &["*.awk"]),
     ("bazel", &[
-        "*.bazel", "*.bzl", "*.BUILD", "*.bazelrc", "BUILD", "WORKSPACE",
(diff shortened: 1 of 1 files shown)


crates/ignore/src/default_types.rs
```diff
@@ -23,7 +23,8 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("avro", &["*.avdl", "*.avpr", "*.avsc"]),
     ("awk", &["*.awk"]),
     ("bazel", &[
-        "*.bazel", "*.bzl", "*.BUILD", "*.bazelrc", "BUILD", "WORKSPACE",
+        "*.bazel", "*.bzl", "*.BUILD", "*.bazelrc", "BUILD", "MODULE.bazel", "WORKSPACE",
+        "WORKSPACE.bazel",
```
BurntSushi, line 27: It's hard to tell in GitHub's UI, but could you please make sure lines are wrapped to 79 columns inclusive? Thanks.

BurntSushi (changes requested):

crates/ignore/src/default_types.rs
keith: Done!

keith:
Thanks!

PR #2216, opened 2022-05-19
Add MPR installation instructions

hwittenborn:
This adds installation instructions for the [makedeb Package Repository](https://mpr.makedeb.org), an AUR-like platform for Debian and Ubuntu based systems.

The MPR is a personal project of mine, but I thought it'd be helpful to include it here so people know its an option.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,14 @@ seem to work right and generate a number of very strange bug reports that I
 don't know how to fix and don't have the time to fix. Therefore, it is no
 longer a recommended installation option.)
 
+If you're a **makedeb** user, you can install makedeb from the [makedeb Package Repository](https://mpr.makedeb.org/packages/ripgrep):
+
+```
+$ git clone 'https://mpr.makedeb.org/ripgrep'
+$ cd ripgrep/
+$ makedeb -si
+```
+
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks for submitting this.

I'll note that your project page doesn't appear to load for me (note that I am behind a VPN):

```
$ curl 'https://mpr.makedeb.org'
curl: (35) OpenSSL SSL_connect: SSL_ERROR_SYSCALL in connection to mpr.makedeb.org:443
```

Separately from that, I personally haven't heard of the `makedeb` project before, which makes me pause before including it in the installation instructions. I'd prefer it reach a bit more of a critical mass before recommending it here. Apologies for being vague, but it is difficult not to be in such things.

hwittenborn:
> I'll note that your project page doesn't appear to load for me (note that I am behind a VPN)

I'm assuming it's something with your VPN, it loads just fine on my end. I'd try without it if you're still curious.

>  I personally haven't heard of the makedeb project before, which makes me pause before including it in the installation instructions

That's fair enough. The MPR has about 300 packages at the moment, but makedeb itself has been growing in popularity. It's definitely more of a niche thing, as the project was only started about a year ago.

Thanks regardless though! I thought I'd just throw this out and see what happened.

PR #2263, opened 2022-07-18
Extend `php` and markdown default types

mal-tee:


crates/ignore/src/default_types.rs
```diff
@@ -167,7 +167,10 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("pascal", &["*.pas", "*.dpr", "*.lpr", "*.pp", "*.inc"]),
     ("pdf", &["*.pdf"]),
     ("perl", &["*.perl", "*.pl", "*.PL", "*.plh", "*.plx", "*.pm", "*.t"]),
-    ("php", &["*.php", "*.php3", "*.php4", "*.php5", "*.phtml"]),
+    ("php", &[
+        "*.php", "*.php3", "*.php4", "*.php5", "*.php7", "*.php8",
```
BurntSushi, line 171: What about `*.php6`?

crates/ignore/src/default_types.rs
```diff
@@ -167,7 +167,10 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
...
+        "*.pht", "*.phtml"
```
BurntSushi, line 172: I don't think I've ever seen `pht` used for PHP before. Can you show me a project that uses it?

BurntSushi (commented):
Thanks! I left some questions since I'm a little unsure.

crates/ignore/src/default_types.rs
mal-tee: It quite obscure. But I've just noticed that it's also used by other language eco-systems. Maybe we shouldn't include it to prevent overlaps and false-positives.

crates/ignore/src/default_types.rs
mal-tee: PHP 6 doesn't exist [1], but we could include the extension as well. :) 

[1] https://wiki.php.net/rfc/php6

BurntSushi: Haha. Nice. OK, yes, leave it out. Could you add a small comment with a link to that page in the source here? I'm sure I won't be the only one to notice and ponder about why `.php6` is absent.

crates/ignore/src/default_types.rs
mal-tee: [Github's search ](https://github.com/search?q=extension%3Apht) reveals some repos using it, e.g. https://github.com/chrisbroski/lulaleasuretime/tree/master/inc and https://github.com/RundizBones/ModuleAdmin/tree/3599dfbe41708dd94cfb7306dad1a44045004bd4/Console/CreateModuleTemplate

BurntSushi: Fair enough...

crates/ignore/src/default_types.rs
```diff
@@ -167,7 +167,11 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("pascal", &["*.pas", "*.dpr", "*.lpr", "*.pp", "*.inc"]),
     ("pdf", &["*.pdf"]),
     ("perl", &["*.perl", "*.pl", "*.PL", "*.plh", "*.plx", "*.pm", "*.t"]),
-    ("php", &["*.php", "*.php3", "*.php4", "*.php5", "*.phtml"]),
+    ("php", &[
+        // note that PHP 6 doesn't exist
```
BurntSushi, line 171:
```suggestion
        // note that PHP 6 doesn't exist
        // See: https://wiki.php.net/rfc/php6
```

BurntSushi (changes requested):

BurntSushi (approved):
Thanks!

PR #2266, opened 2022-07-23
fix(ignore): do not panic if matched file is outside root dir

mauri870:


BurntSushi:
@mauri870 Does this correspond to an actual bug you're seeing in ripgrep? Otherwise, the behavior is consistent with the method's public docs.

Really need more context for stuff like this please.

mauri870:
@BurntSushi I guess this is expected behavior then, I will close the ticket.

PR #2274, opened 2022-08-06
Fix typos in byte-offset and colors

drebelsky:
The documentation for both of these flag had "the the"; this PR changed them to "the".


BurntSushi (approved):
Thanks!

PR #2275, opened 2022-08-07
Add `patch` printer

BatmanAoD:
Proof-of-concept for a grep-printer that works in conjunction with `--replace` to generate `patch` files, as suggested [here](https://github.com/BurntSushi/ripgrep/issues/74#issuecomment-251377251).

### DIFF
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -2383,11 +2383,11 @@ A patch file in the unified diff format (used by 'diff -u') will be generated;
 applying this patch file will replace every match with the replacement text
 specified by '--replace'. (Note that ripgrep will only generate the patch; it
 will not modify your files.)
+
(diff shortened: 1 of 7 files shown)


BurntSushi:
Could you please state your intent here?

Could you please also confirm that you've seen [this comment](https://github.com/BurntSushi/ripgrep/issues/74#issuecomment-1191336022)?

BatmanAoD:
@BurntSushi It's essentially context for the comment I just added to that thread. I wanted to see what it would actually take to generate patch output as suggested.

I think the links from the comment will still work if the PR is closed, so I'll go ahead and close it.

BurntSushi:
Ah gotya. I wasn't sure. Thanks for clarifying!

PR #2276, opened 2022-08-07
Expose utils

BatmanAoD:
Needed for https://github.com/BatmanAoD/RipPatch


BurntSushi:
I'm not comfortable exposing this as-is. It seems pretty out of place in the public API of this crate.

Looking at the code, it looks like you could probably copy/re-implement it without too much trouble.

Issue #2283, opened 2022-08-16
Add support for Solidity/*.sol file type


johnsaigle:
#### Describe your feature request

Add Solidity file type to the list of supported files and extensions for the --type flag. Should match *.sol extensions

Not supported as of this rg version:
ripgrep 13.0.0
-SIMD -AVX (compiled)

BurntSushi:
I don't track requests to add things to the default type definitions. Instead, just submit a PR please. Thanks.

For more obscure file types (I've never heard of Solidity), please include some link explaining what it is.

Thanks!

PR #2284, opened 2022-08-16
Add support for Solidity as a file extension

johnsaigle:
Adding support for Solidity as a language and the *.sol file extension corresponding to it. The idea is to allow for the use of the flag `--type solidity` in order to search only in Solidity files.

Solidity is a programming language for the Ethereum Virtual Machine and is one of the most popular languages for developing blockchain software.
https://soliditylang.org/about/
https://en.wikipedia.org/wiki/Solidity

Follow up to #2283. I'm not an experienced Rust dev and this is my first contribution to the repo so let me know if there are additional code locations to update or any other steps required for making a high-quality PR.


BurntSushi (approved):
LGTM! Thanks.

PR #2301, opened 2022-09-06
derive `Default` for `SubjectBuilder`

kraktus:
Hello,

I am reviewing potentially clippy false-positive/false-negative after a change to `clippy::derivable_impls`.
In this case the lint seems correct unless the manual implementation is to emphasise the `strip_dot_prefix: false`. In this case feel free to close

### DIFF
--- a/crates/core/subject.rs
+++ b/crates/core/subject.rs
@@ -4,15 +4,9 @@ use ignore::{self, DirEntry};
 use log;
 
 /// A configuration for describing how subjects should be built.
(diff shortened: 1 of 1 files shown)


BurntSushi:
I think it's just a matter of code evolution. If one adds another boolean that one wants to default to `true`, the `derive` will silently do the wrong thing.

It's not a huge deal. I sometimes derive `Default` for types like this, sometimes not. In this case I think I'd just prefer to leave it as-is. Thanks!

PR #2302, opened 2022-09-09
Add support for v language

edam:
V (http://vlang.io) uses '.v' files.


svercl:
Doesn't Verilog also use the `.v` extension?

edam:
Yes, it does.

Also Coq (also supported by rg) uses .v files.

But if this is just about file selection, does it matter? Can't rg just support all three? 🤔

PR #2312, opened 2022-09-23
bump bstr to 1.0

### LINKED ISSUE #2310: bump bstr to 1.0
Bstr version 1.0 was released with a new MSRV of 1.60.
I understand from the README that ripgrep has a MSRV of 1.34 so upgrading isn't directly an option..
However many rust project that depend on components of ripgrephave an MSRV of 1.60 (or above).
It would be nice if a new version of the ripgrep crates were released that bump bstr 1.0.

Currently any crate that depends on ripgrep components can not update bstr (or any crate that depends on it) without compiling two versions of bstr.

This problem occurred in https://github.com/helix-editor/helix/pull/3890.
Here we wanted to update gitoxide to 0.24, which depends on bstr 1.0 but chose to postpone because globset still depends on bstr 0.2.

I originally wanted to open a PR, however as ripgrep is developed in a workspace bumping bstr would mean bumping the MSRV of ripgrep aswell. Might it be possible to add an optional feature to switch to bstr 1.0?

pascalkuthe:
updates bstr to 1.0 and closes #2310
This allows downstream crates to use bstr 1.0 (or other crates that depend on it) without compiling it twice.

This raises the MSRV to 1.60.
When I first filed #2310 I assumed that raising the MSRV was not an option because it was set to 1.34.
However I discovered that ripgrep already has an MSRV of 1.52 since 9f924ee187d4c62aa6ebe4903d0cfc6507a5adb5.
To avoid future confusions like this I also updated the README with the new MSRV.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,7 +43,7 @@ jobs:
         include:
         - build: pinned
           os: ubuntu-22.04
-          rust: 1.52.1
+          rust: 1.60.0
         - build: stable
           os: ubuntu-22.04
           rust: stable
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,13 +36,14 @@ checksum = "bef38d45163c2f1dde094a7dfd33ccf595c92905c8f8f4fdc18d06fb1037718a"
 
 [[package]]
 name = "bstr"
-version = "0.2.17"
+version = "1.0.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ba3569f383e8f1598449f1a423e72e99569137b47740b1da11ef19af3d5c3223"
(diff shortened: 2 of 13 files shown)


stmcginnis:
This would be helpful for the project I'm working on to get this. Anything we can do to help get this PR through?

pascalkuthe:
This is actually redundant now as @BurntSushi fixed this himself, glad to see this land

BurntSushi:
Aye yeah, thanks. Sorry I missed this PR. Generally speaking, I do all dependency upgrades myself so that I can scrutinize changes to the dependency tree.

pascalkuthe:
Fair enough, I assumed this was just an uncontroversial chore since you are also the author of `bstr` :D Thanks for all your work :)

PR #2315, opened 2022-09-29
Enable Dependabot for the Actions workflows

LingMan:
Dependabot automatically files PRs for updatable dependencies. As configured it watches all workflow files in `.github/workflows` for possible updates to any of the Actions depended upon.

---
It also supports updating rust dependencies, but I've kept this minimal to gauge interest. See https://github.com/eminence/terminal-size/pull/42 for an example of a PR you'll also automatically receive after merging this.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


BurntSushi (approved):
Hmmm, okay, I'll give this a try, but might end up removing it if it's too noisy. I didn't know Dependabot existed for GitHub Actions.

I'm pretty sure I do not want to enable this for Rust dependencies. I do update them occasionally, but I like to do it manually and deliberately. With Dependabot, I'll feel like I'm running in a hamster wheel. :-)

LingMan:
Yeah, if you have the lock file committed, Dependabot can be a bit overwhelming. Would be great if it could be configured to submit a single PR updating all dependencies in one go to cut down on the hamster wheel a bit.

For Github Actions it shouldn't be particularly noisy. You'll get a few PRs initially but actions update to a new major version relatively rarely.

PR #2316, opened 2022-09-29
Update CI

LingMan:
The major change here is to switch away from the `ubuntu-18.04` image. It is deprecated and will be removed by 2023-04-01 with scheduled brownouts starting on 2022-10-03.

While making changes in the area, I've also switched `dtolnay/rust-toolchain` to use the master branch and added two minor cleanups.

All of these have already been applied over in the bstr repository.


BurntSushi (approved):
Lovely, thank you!

PR #2330, opened 2022-10-10
Add install commands for ALT Linux

bircoph:


README.md
```diff
@@ -295,6 +295,13 @@ seem to work right and generate a number of very strange bug reports that I
 don't know how to fix and don't have the time to fix. Therefore, it is no
 longer a recommended installation option.)
 
+If you're an **ALT** user, you can install ripgrep from the
+[official repo](https://packages.altlinux.org/en/search?name=ripgrep):
+
+```
+$ apt-get install ripgrep
```
BurntSushi, line 302: Should this be `sudo apt-get` like the other `apt-get` instructions in this README?

bircoph: I think no, because there are different ways of obtaining root in ALT: su, ssh, sudo. On many setups sudo is absent at all out of the box. I may change prompt from "$ " to "# " so it will be highlighted that root access is required. But I think users will understand what is necessary even with $.

BurntSushi: All of that is true of other distros too. Please be consistent with other instructions and use sudo. Using `#` is waaaaay too subtle.

PR #2332, opened 2022-10-16
ci: Added uploading build artifacts and also build release

mitchcapper:
Upload build artifacts for CI to allow people to easily grab nightly builds

Ran into the long file path issue, found the fix from last year: #2049 but as there were no nightly artifacts would need to clone and build just to test.  
This also adds release building to the CI for tests (to get us both sets of artifacts).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,9 +19,14 @@ jobs:
       TARGET_DIR: ./target
       # Emit backtraces on panics.
       RUST_BACKTRACE: 1
+      # Additional build flags for cargo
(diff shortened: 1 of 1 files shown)


mitchcapper:
Here is a sample as well: https://github.com/mitchcapper/ripgrep/actions/runs/3258975771

BurntSushi:
I appreciate what you're trying to do here, but I do not want the headache of maintaining nightly release builds. IMO the build process for ripgrep is very easy and so I don't think it's a huge ask to build it yourself if you want to test/use whatever is on master.

PR #2335, opened 2022-10-20
ignore/types: add motoko and candid

smallstepman:
candid: https://internetcomputer.org/docs/current/developer-docs/build/candid/candid-concepts

motoko: https://internetcomputer.org/docs/current/developer-docs/build/cdks/motoko-dfinity/motoko/

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -32,6 +32,7 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("bzip2", &["*.bz2", "*.tbz2"]),
     ("c", &["*.[chH]", "*.[chH].in", "*.cats"]),
     ("cabal", &["*.cabal"]),
+    ("candid", &["*.did"]),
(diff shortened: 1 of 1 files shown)


BurntSushi:
It looks like that's a source-available and not a open-source project, right? I generally try to stick to FOSS stuff for the default types, and expect non-FOSS types to maintain their own file types (which can be added to ripgrep on every CLI invocation via `--type-add`, typically via an alias, wrapper script or config file).

smallstepman:
I believe both candid and motoko are FOSS but I'm no licensing expert (both are Apache 2.0):
- https://github.com/dfinity/candid
- https://github.com/dfinity/motoko

The blockchain platform which uses these languages (Internet Computer) is open-sourced, [but not FOSS](https://github.com/dfinity/ic/blob/master/LICENSE).

BurntSushi:
Ah okay, yeah, I was looking at https://github.com/dfinity/ic. Like I said, the dfinity web site makes it pretty hard to find the source.

Since candid and motoko are indeed open source, I'll bring this in. Thanks!

PR #2340, opened 2022-10-26
Add ReScript and ReasonML default types

glennsl:
[ReScript](https://rescript-lang.org/) is an OCaml-derived programming language that compiles to efficient and human-readable JavaaScript.

[ReasonML](https://reasonml.github.io/) is an alternative syntax for OCaml.

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -190,6 +190,7 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("rdoc", &["*.rdoc"]),
     ("readme", &["README*", "*README"]),
     ("red", &["*.r", "*.red", "*.reds"]),
+    ("rescript", &["*.res", "*.resi"]),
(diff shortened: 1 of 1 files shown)


crates/ignore/src/default_types.rs
```diff
@@ -190,6 +190,8 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("rdoc", &["*.rdoc"]),
     ("readme", &["README*", "*README"]),
     ("red", &["*.r", "*.red", "*.reds"]),
+    ("reasonml", &["*.re", "*.rei"]),
```
BurntSushi, line 193: This should come before `red` and after `readme`.

BurntSushi (changes requested):

crates/ignore/src/default_types.rs
glennsl: Indeed. Fixed!

BurntSushi (approved):

PR #2344, opened 2022-11-01
readme: add guix installation instructions

jgarte:
https://guix.gnu.org/

### DIFF
--- a/README.md
+++ b/README.md
@@ -271,6 +271,13 @@ $ nix-env --install ripgrep
 $ # (Or using the attribute name, which is also ripgrep.)
 ```
 
+If you're a **Guix** user, you can install ripgrep from the official
+package collection:
+
+```
+$ guix install ripgrep
+```
+
 If you're a **Debian** user (or a user of a Debian derivative like **Ubuntu**),
 then ripgrep can be installed using a binary `.deb` file provided in each
 [ripgrep release](https://github.com/BurntSushi/ripgrep/releases).


README.md
```diff
@@ -271,6 +271,12 @@ $ nix-env --install ripgrep
 $ # (Or using the attribute name, which is also ripgrep.)
 ```
 
+If you're a **Guix** user, you can install ripgrep from the official [package collection](https://guix.gnu.org/en/packages/ripgrep-13.0.0/):
```
BurntSushi, line 274: Please follow the style of the surrounding code and wrap to 79 columns (inclusive). If the link is too long, that's okay, but it should go on its own line. (Just follow the surrounding style.)

BurntSushi (changes requested):

README.md
jgarte: Hi, I pushed a update. Thanks for the review!

README.md
```diff
@@ -271,6 +271,13 @@ $ nix-env --install ripgrep
 $ # (Or using the attribute name, which is also ripgrep.)
 ```
 
+If you're a **Guix** user, you can install ripgrep from the official
+[package collection](https://guix.gnu.org/en/packages/ripgrep-13.0.0/):
```
BurntSushi, line 275: Also, sorry, I'm just noticing this now, but this URL has the version number in it. Could we replace it with a URL that doesn't have a version in it? Otherwise it's just going to bitrot. If there is no such URL, then I'd say to just remove it.

BurntSushi (changes requested):

README.md
jgarte: I updated and pushed with the url removed. Thanks for the review!

BurntSushi (approved):

PR #2349, opened 2022-11-11
ignore/types: add dts filetype

arbrauns:
See: https://www.devicetree.org/

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -56,6 +56,7 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("cython", &["*.pyx", "*.pxi", "*.pxd"]),
     ("d", &["*.d"]),
     ("dart", &["*.dart"]),
+    ("devicetree", &["*.dts", "*.dtsi"]),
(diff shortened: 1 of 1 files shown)


crates/ignore/src/default_types.rs
```diff
@@ -52,6 +52,7 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("cshtml", &["*.cshtml"]),
     ("css", &["*.css", "*.scss"]),
     ("csv", &["*.csv"]),
+    ("dts", &["*.dts", "*.dtsi"]),
```
BurntSushi, line 55: I don't think `dts` is widely recognized enough to be its own type name. Could you rename this to `devicetree` please? Also, please follow the conventions of the surrounding (and the comment above) and make sure the list is sorted lexicographically.

BurntSushi (changes requested):

crates/ignore/src/default_types.rs
arbrauns: Done. Not sure how that mis-sorting happened, but that's fixed too now.

BurntSushi (approved):

PR #2351, opened 2022-11-14
ignore/types: add automated test for sortedness

arbrauns:
After I accidentally mixed up the sorting in #2349 and didn't notice, I thought I'd add a test so hopefully the next time it happens to someone it can be caught before it's sent in as a PR.

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -291,3 +291,26 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ]),
     ("zstd", &["*.zst", "*.zstd"]),
 ];
+
+#[cfg(test)]
+mod tests {
+    use super::DEFAULT_TYPES;
+
+    #[test]
+    fn default_types_are_sorted() {
+        let mut names = DEFAULT_TYPES.iter().map(|(name, _exts)| name);
+
+        let Some(mut previous_name) = names.next() else { return; };
+
+        for name in names {
+            assert!(
+                name > previous_name,
+                r#""{}" should be sorted before "{}" in `DEFAULT_TYPES`"#,
+                name,
+                previous_name
+            );
+
+            previous_name = name;
+        }
+    }
+}


crates/ignore/src/default_types.rs
```diff
@@ -290,3 +290,22 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ]),
     ("zstd", &["*.zst", "*.zstd"]),
 ];
+
+#[test]
+fn test_default_types_are_sorted() {
+    let mut names = DEFAULT_TYPES.iter().map(|(name, _exts)| name);
+
+    if let Some(first) = names.next() {
```
BurntSushi, line 298: I would write this as:

```rust
let first = match names.next() {
    None => return,
    Some(first) => first,
};
```

Or perhaps even better, using the [newly stabilized `let else`](https://github.com/rust-lang/rust/pull/93628/):

```rust
let Some(first) = names.next() else { return };
```

Although in order to use `let else` we need to bump MSRV to Rust 1.65. Since ripgrep tracks the latest stable release of Rust, I have no problem with that. Indeed, [I just pushed a commit to do so](https://github.com/BurntSushi/ripgrep/commit/8905d54a9f25f4c1e4e3ca8331f517473e174d87), so all you need to do to use `let else` is to rebase on master.

BurntSushi: The higher level point here is to keep code linear via early returns. That way, you avoid nesting the main body here.

BurntSushi (changes requested):
Great idea, thank you!

crates/ignore/src/default_types.rs
arbrauns: That's much neater. Of course the best solution would be rust-lang/rust#53485, but that seems to still be quite a while away.

If ripgrep uses latest stable anyway, what's the opinion on `edition = "2021"`? It would allow for things like implicit arguments in panic/assert format strings: https://doc.rust-lang.org/edition-guide/rust-2021/panic-macro-consistency.html.

BurntSushi: I'll very likely do the edition migration myself before the next release, whenever that is. I'm in no rush to be honest. I'll likely only do things that `cargo fix` does automatically. Converting explicit to implicit arguments by hand is a fair bit of work.

With respect to `is_sorted`, I actually think your approach here is better. It gives a better failure mode by providing a witness. `is_sorted` wouldn't do that. :-)

BurntSushi (approved):

Issue #623, opened 2017-10-06
hidden files to be searched by default


akostadinov:
Am I the only one that thinks hidden files should be searched by default? Especially when trying to find something related to configuration, these things are often put into hidden files in user home.

And if I'm the only one, probably #196 is what can help me.

okdana:
Excluding hidden/dot files by default is common for this class of tool — notably, `ag` and `pt` both do it. `rg` was inspired directly by `ag`, so that's probably (?) the immediate reason for the current behaviour. But i think it makes sense even if you ignore the historical baggage, since these tools are mainly designed for searching through source code, and dot files in that context are usually just noise.

If you've read that other ticket you've probably already picked this up, but just in case, this is your best work-around for now:

```
alias rg='rg --hidden'
```

BurntSushi:
@okdana has got it. This isn't changing.

kriskhaira:
The following is @okdana's solution, but with the .git folder remaining ignored:
```
alias rg="rg --hidden --glob '!.git'"
```

Nakilon:
Personally I would never complain if the Github Actions wasn't invented as being put into `.github` and that now I need to grep those often to find how I solve common issues.

BurntSushi:
@Nakilon Sorry, what's the point of your comment? Just to complain? Why not put `!/.github/` in your `.ignore`? Boom. Problem solved.

Nakilon:
> put `!/.github/` in your `.ignore`?

Can't figure out how this autofiltering thing works. I have `mydir/infrared/.rgignore` with content:
```none
/*.cache.yaml
```
but
```
$ rg jobs: --sortr created mydir -g*.y*l
...mydir/infrared/token.cache.yaml
204267:Now people are getting food/gas priced out of their homes and jobs:
...
```
Why doesn't it ignore the file?

BurntSushi:
Please open a new Discussion question instead of posting in some old issue that may or may not be related.

Please also include a full reproduction that others can try.

Please also read the guide. For example, why haven't you tried using the `--debug` flag yet?

StackOverflowExcept1on:
I also ran into the same problem today when the `rg foo` command cannot find the code that is in `.github/workflows`.It makes sense to exclude the `.gitignore` files and the `.git` folder from hidden files. The hidden files should be searched by default.
[...]

PR #2356, opened 2022-11-20
add '--hidden' to example configuration

kevinushey:
Motivated by issues like:

https://github.com/BurntSushi/ripgrep/issues/623

I understand (and agree with) the decision to not change the defaults; this just presents the commonly-used:

```
--hidden
--glob=!.git/*
```

as a potential "default configuration" for users who want `rg` to scan dotfiles, and hopefully makes that more discoverable.

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -567,12 +567,15 @@ $ cat $HOME/.ripgreprc
 --type-add
 web:*.{html,css,js}*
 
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
Fair enough.

kevinushey:
Thank you!

PR #2358, opened 2022-11-23
build: Use GITHUB_REF_NAME instead of GITHUB_REF

jpmckinney:
GitHub added it more recently: https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables


BurntSushi (approved):
w00t! Thank you!

PR #2360, opened 2022-11-26
ci: Use GitHub CLI (set-output is deprecated)

jpmckinney:
I figured this out for my own project, so contributing here as I otherwise followed your `release.yml` example.

actions/create-release uses the deprecated `set-output` command, and it is archived, so it will never be updated.

actions/upload-release-asset is also archived.


BurntSushi (approved):
Hmmm, all righty, thanks!

I'll be (hopefully) putting a new release soon, so I'll look more closely at this and make sure it works.

Issue #2243, opened 2022-06-24
Strange behavior of `--sort modified`, giving incorrect order


qx220:
#### What version of ripgrep are you using?

```
ripgrep 11.0.2
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

#### How did you install ripgrep?

```
apt-get
```

#### What operating system are you using ripgrep on?

```
5.15.0-33-generic #34~20.04.1-Ubuntu SMP Thu May 19 15:51:16 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

#### Describe your bug.

When using the vim plugin LeaderF, which uses rg as its search engine, I realized that the sorting order seems to be "randomly" incorrect.

After a bit of tweaking, I produced a minimal script reproducing the issue.

#### What are the steps to reproduce the behavior?

Under a clean temporary folder:

```bash
echo string > modified_first; sleep 1; mkdir -p dir; echo string > dir/modified_second; find -type f | xargs stat
```

At this point, rg returns the correct order when executing:

```bash
rg --sort modified string
```

**However, if we modify the two files now, the order becomes incorrect**:

```bash
echo string > modified_first; sleep 1; mkdir -p dir; echo string > dir/modified_second; find -type f | xargs stat # as the output shows, "./modified_first" is modified before "./dir/modified_second"
```

Now if we run `rg --sort modified string`, we find the following output:

```
dir/modified_second
1:string

modified_first
1:string
```

when the expected behavior should be:

```
modified_first
1:string

dir/modified_second
1:string
```

I'm attaching the full transcript with output below:

```bash
l14:~/tmp/ripgrep$ echo string > modified_first; sleep 1; mkdir -p dir; echo string > dir/modified_second; find -type f | xargs stat
  File: ./modified_first
  Size: 7         	Blocks: 8          IO Block: 4096   regular file
Device: fd00h/64768d	Inode: 9699609     Links: 1
Access: (0664/-rw-rw-r--)  Uid: ( 1000/    user)   Gid: ( 1000/    user)
Access: 2022-06-23 22:29:05.404139561 -0400
Modify: 2022-06-23 22:29:05.404139561 -0400
Change: 2022-06-23 22:29:05.404139561 -0400
 Birth: -
  File: ./dir/modified_second
  Size: 7         	Blocks: 8          IO Block: 4096   regular file
Device: fd00h/64768d	Inode: 9699610     Links: 1
[...]

PR #2361, opened 2022-11-28
fix: sorting nested results by system time

nguyenvukhang:
Bug is aptly described in https://github.com/BurntSushi/ripgrep/issues/2243.

This PR aims to address this by delaying result sorting to after every match has been found.

### DIFF
--- a/crates/core/args.rs
+++ b/crates/core/args.rs
@@ -352,6 +352,9 @@ impl Args {
             Ok(v) => v,
             Err(_) => return subjects,
         };
+        if sorter.check().is_err() {
+            return subjects;
+        }
         #[inline]
         fn unzip<K, V>(v: Vec<(K, V)>) -> Vec<V> {
             v.into_iter().map(|v| v.1).collect()
@@ -444,6 +447,24 @@ impl SortBy {
 
     fn none() -> SortBy {
         SortBy::asc(SortByKind::None)
+    }
+
+    /// Try to check that the sorting criteria selected is actually supported.
+    /// If it isn't, then an error is returned.
+    fn check(&self) -> Result<()> {
+        match self.kind {
+            SortByKind::None | SortByKind::Path => {}
+            SortByKind::LastModified => {
+                env::current_exe()?.metadata()?.modified()?;
+            }
+            SortByKind::LastAccessed => {
+                env::current_exe()?.metadata()?.accessed()?;
+            }
+            SortByKind::Created => {
+                env::current_exe()?.metadata()?.created()?;
+            }
+        }
+        Ok(())
     }
 }
 
--- a/crates/core/args.rs
+++ b/crates/core/args.rs
@@ -346,48 +346,44 @@ impl Args {
         Ok(self.matches().walker_builder(self.paths())?.build())
     }
 
-    /// Sort subjects if a sorter is specified
-    pub fn sort(&self, subjects: Vec<Subject>) -> Vec<Subject> {
+    /// Returns true if and only if `stat`-related sorting is required
+    pub fn needs_stat_sort(&self) -> bool {
+        return self.matches().sort_by().map_or(
+            false,
+            |sort_by| match sort_by.kind {
+                SortByKind::LastModified
+                | SortByKind::Created
+                | SortByKind::LastAccessed => sort_by.check().is_ok(),
+                _ => false,
+            },
+        );
+    }
+
+    /// Sort subjects if a sorter is specified, but only if the sort requires
+    /// stat calls. Non-stat related sorts are handled during file traversal
+    ///
+    /// This function assumes that it is known that a stat-related sort is
+    /// required, and does not check for it again.
+    ///
+    /// It is important that that precondition is fulfilled, since this function
+    /// consumes the subjects iterator, and is therefore a blocking function.
+    pub fn sort_by_stat<I>(&self, subjects: I) -> Vec<Subject>
+    where
+        I: Iterator<Item = Subject>,
+    {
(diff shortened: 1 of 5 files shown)


crates/core/args.rs
```diff
@@ -346,6 +346,50 @@ impl Args {
...
+            Ok(v) => v,
+            Err(_) => return subjects,
+        };
+        if sorter.check().is_err() {
+            return subjects;
+        }
+        #[inline]
+        fn unzip<K, V>(v: Vec<(K, V)>) -> Vec<V> {
+            v.into_iter().map(|v| v.1).collect()
+        }
```
BurntSushi, line 361: So this isn't really `unzip`. `unzip` is `Vec<(K, V)> -> (Vec<K>, Vec<V>)`. And functions like this shouldn't be marked as `#[inline]`: they are polymorphic so it will get inlined anyway. And even if it weren't, the perf overhead of a function call in this context is negligible.

In any case, I would get rid of this function entirely. Instead, do:

```rust
let items = match sorted.kind { ... };
items.into_iter().map(|v| v.1).collect()
```

crates/core/main.rs
```diff
@@ -79,17 +79,12 @@ fn try_main(args: Args) -> Result<()> {
...
     let mut matched = false;
     let mut searched = false;
-
-    for result in args.walker()? {
-        let subject = match subject_builder.build_from_result(result) {
-            Some(subject) => subject,
-            None => continue,
-        };
+    let subjects = sorted_results(args);
+    for subject in subjects {
```
BurntSushi, line 87: This is a big no-no from what I can tell. It looks like you're _unconditionally_ collecting all paths before iterating over them even if no `--sort` flag is passed. That means you're incurring the perf/memory/latency---at least in part---of sorting even when you don't need to.

crates/core/main.rs
```diff
@@ -215,14 +210,10 @@ fn eprint_nothing_searched() {
...
-    let subject_builder = args.subject_builder();
     let mut matched = false;
     let mut path_printer = args.path_printer(args.stdout())?;
-    for result in args.walker()? {
-        let subject = match subject_builder.build_from_result(result) {
-            Some(subject) => subject,
-            None => continue,
-        };
+    let subjects = sorted_results(args);
+    for subject in subjects {
```
BurntSushi, line 216: Same deal here as for `search` above.

BurntSushi (changes requested):
Nice work but yeah, there is a fair bit of work left here. The two big items:

1. It is a hard requirement that if `--sort` is not given, then there should be no additional overhead here. But in this PR, it looks like you always collect all of the paths before starting a search. Today, ripgrep will start searching well before it finishes directory traversal.
2. There are a number of places where errors are being swallowed. We can't do that.

I'm also not sure about moving sorting to the `main` functions. I think it's probably fine, although I was trying to keep the `main` functions as simple as I could. But if we are going to deal with sorting in `main.rs`, then I think it would be nice to add a short comment in the parallel searcher stating why it doesn't handle sorting. (Because if sorting is enabled, then parallel searching is disabled in `args.rs`.)

nguyenvukhang:
> 1. It is a hard requirement that if `--sort` is not given, then there should be no additional overhead here. But in this PR, it looks like you always collect all of the paths before starting a search. Today, ripgrep will start searching well before it finishes directory traversal.

Yep. Fixed it. Top-level `search()` and `files()` functions will now specifically check if a sort that requires calls to `stat` is required.

The latest implementation using new functions `continue_search()` and `continue_files()` is a result of trying to coerce the types of the two possible states of preliminary search results:
1. uncollected and unconsumed iterator. (no sort)
2. collected, sorted, and re-made into an iterator.

I'd prefer to stick to having just one function for `search` and `files` each, though I'm not sure if that's possible.

> 2. There are a number of places where errors are being swallowed. We can't do that.

Agreed. I noticed that the previously reviewed code swallowed the `Result<Walk>` that `args.walker()?` emitted, and that has been fixed.

https://github.com/BurntSushi/ripgrep/pull/2361#discussion_r1035183685 has also been addressed.

crates/core/args.rs
nguyenvukhang: Yep, changed it.

crates/core/main.rs
nguyenvukhang: resolved by https://github.com/BurntSushi/ripgrep/pull/2361/commits/d4706913adf88cb3c6b395d0afcf493f39208169

crates/core/main.rs
nguyenvukhang: also resolved by https://github.com/BurntSushi/ripgrep/pull/2361/commits/d4706913adf88cb3c6b395d0afcf493f39208169

crates/ignore/src/walk.rs
```diff
@@ -826,12 +826,11 @@ impl WalkBuilder {
     /// by `sort_by_file_name`.
     ///
     /// Note that this is not used in the parallel iterator.
-    pub fn sort_by_file_path<F>(&mut self, cmp: F) -> &mut WalkBuilder
```
BurntSushi, line 829: Why did you do this? It's a breaking API change for the `ignore` crate. Same for `sort_by_file_name` below.

nguyenvukhang: Sorry about that, this change was not necessary. It's been reverted by a8c3e89.

BurntSushi:
You're all set. I already did the fixups needed to get this PR into shape. You can see it in #2555.

nguyenvukhang:
I see. Thanks for patching up my tests and adding delays so the `stat` calls are correct.

PR #2373, opened 2022-12-20
Set rust-version in Cargo.toml

atouchet:
This MSRV was bumped higher than 1.56 in 8905d54a9f25f4c1e4e3ca8331f517473e174d87. Should this be added to any other crates as well?


BurntSushi (approved):
> Should this be added to any other crates as well?

Probably, for crates with an MSRV at or greater than 1.56. I'm not sure if many of mine have reached that point.

PR #2378, opened 2022-12-28
Try harder to fix #916

### LINKED ISSUE #916: Zombie processes cause rg --files to hang in /proc

tavianator:
The underlying Rust issue[1] was recently re-introduced and re-fixed.
Rather than wait for a new Rust release, apply effectively the same fix
as a workaround in the ignore crate itself.

[1]: https://github.com/rust-lang/rust/issues/50619

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -1495,6 +1495,7 @@ impl<'s> Worker<'s> {
             return WalkState::Skip;
         }
         for result in readdir {
+            let is_err = result.is_err();
             let state = self.generate_work(
                 &work.ignore,
                 depth + 1,
@@ -1503,6 +1504,12 @@ impl<'s> Worker<'s> {
             );
             if state.is_quit() {
                 return state;
+            }
+            // Break after we see an error.
+            // See https://github.com/rust-lang/rust/issues/50619
+            // and https://github.com/BurntSushi/ripgrep/issues/916
+            if is_err {
+                break;
             }
         }
         WalkState::Continue


tmccombs (approved):

BurntSushi:
My PR review latency is very high, so this sort of strategy tends not to work so well unless we happen to get lucky and I see and respond to it quickly. But in this case, it looks like a few Rust releases have come out since the issue was re-fixed, so I think we can avoid bringing this in.

PR #2388, opened 2023-01-01
fix(globset/serde1): allow deserializing owned strings into `Glob`

### LINKED ISSUE #2386: globset: `serde` feature - cannot deserialize `String` into `Glob`
Hi, I hope that posting issues about subcrates is ok.

#### What version of globset are you using?

`globset = { version = "0.4.9", features = ["serde1"] }`.

#### Describe your bug.

Because the implementation of trait `Deserialize` uses `&str`, it does not accept deserializing owned `String`.

https://github.com/BurntSushi/ripgrep/blob/61101289fabc032fd8e90009c41d0b78e6dfc9a2/crates/globset/src/serde_impl.rs#L15-L22

This problem occurred to me when I tried to parse `Glob`s from a `toml::Value` object (already obtained from a file). As `toml::Value` owns strings, it caused a problem.

If needed, I can include a MWE (but quite large). 

When parsing some globs, here is the error obtained:

```
Error: TomlDecode(Error { inner: ErrorInner { kind: Custom, line: None, col: 0, at: None, message: "invalid type: string \"src/**.rs\", expected a borrowed string", key: [] } })
```

Not sure if it is interesting performance-wise, but using a custom `deserialize_with` function, with `String` instead of `&str`, fixes the problem (`glob.as_str()`) has to be used afterward).

I can provide tests and make a PR if you are ok with this solution. Otherwise, I am also open to discussions :-)

jeertmans:
This PR modifies how `Glob` is deserialized such that it now allows for owned strings (previously, only borrowed) to be deserialized into `Glob`. This takes inspiration from what is done in the [`regex_serde` crate](https://github.com/tailhook/serde-regex/blob/336bb456ecd146ba9e3fcc2fef71870f603d72c5/src/lib.rs#L179-L191).

For more details, see the discussion in #2386.

The first commit includes tests, which fail without the solution.
The second commit include the solution.

This fixes #2386.

I'd be happy to hear from your thoughts and I'm open to any review :-)

### DIFF
--- a/crates/globset/src/serde_impl.rs
+++ b/crates/globset/src/serde_impl.rs
@@ -1,6 +1,7 @@
+use std::fmt;
+
 use serde::de::{Error, Visitor};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
-use std::fmt;
 
 use crate::Glob;
 
@@ -40,8 +41,9 @@ impl<'de> Deserialize<'de> for Glob {
 
 #[cfg(test)]
 mod tests {
-    use crate::glob::Glob;
     use std::collections::HashMap;
+
+    use crate::glob::Glob;
 
     #[test]
     fn glob_deserialize_borrowed() {
--- a/crates/globset/src/serde_impl.rs
+++ b/crates/globset/src/serde_impl.rs
@@ -1,6 +1,5 @@
-use serde::de::{Error, Unexpected, Visitor};
+use serde::de::{Error, Visitor};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
-use std::borrow::Cow;
 use std::fmt;
 
 use crate::Glob;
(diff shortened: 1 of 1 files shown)


jeertmans:
Like so @LPGhatguy? I have updated the code according to your comment.

I have merged `StrVisitor` and `StringVisitor` implementation into `CowStrVisitor` (see [here](https://rust-random.github.io/rand/src/serde/de/impls.rs.html#599)).

After checking, the first test uses a `Cow::Owned`, while the two others use a `Cow::Borrowed`.

LPGhatguy:
`Cow` doesn't need to be involved here. You should make a `GlobVisitor` that constructs the glob directly from the value instead of pulling it into an intermediate `Cow`.

I don't think that `visit_borrowed_bytes` or `visit_byte_buf` will be reached. Because we're calling `deserialize_str`, we're telling Sede that we're going to deserialize a string, not an arbitrary sequence of bytes.

jeertmans:
Thanks, @LPGhatguy! Indeed, after my commit, I realized that using `Cow` was no longer needed. I implemented a new version where I directly parse a string into a `Glob`.

crates/globset/src/serde_impl.rs
```diff
@@ -12,18 +13,62 @@ impl Serialize for Glob {
     }
 }
 
+struct GlobVisitor;
+
+impl<'a> Visitor<'a> for GlobVisitor {
+    type Value = Glob;
+
+    fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+        formatter.write_str("a string or a borrowed string")
```
LPGhatguy, line 22: This should probably be "a glob pattern" or something more descriptive here.

LPGhatguy (commented):
Last comment, then LGTM.

crates/globset/src/serde_impl.rs
jeertmans: Thanks for noticing! There you go :)

jonasbb (approved):

LPGhatguy (approved):
lgtm

jeertmans:
Hello @BurntSushi, any plan on merging this? Or is it waiting for something else?

crates/globset/src/serde_impl.rs
```diff
@@ -1,5 +1,6 @@
-use serde::de::Error;
+use serde::de::{Error, Visitor};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt;
```
BurntSushi, line 3: Please group imports. std imports should come first, then an empty line, then third party imports (like `serde`) and then finally crate local imports.

BurntSushi: Or just write `std::fmt::Formatter` and `std::fmt::Result` below.

crates/globset/src/serde_impl.rs
```diff
@@ -12,18 +13,62 @@ impl Serialize for Glob {
...
-        Glob::new(glob).map_err(D::Error::custom)
+        deserializer.deserialize_str(GlobVisitor)
     }
 }
 
 #[cfg(test)]
 mod tests {
-    use Glob;
+    use crate::glob::Glob;
+    use std::collections::HashMap;
```
BurntSushi, line 44: Same nit here about import grouping.

BurntSushi (changes requested):
I think this LGTM and looks much better than what it was before. There are some style nits though.

BurntSushi:
> Or is it waiting for something else?

It's waiting on me to review it. In general, my review latency is pretty high.

crates/globset/src/serde_impl.rs
jeertmans: Like so (see recent commit)?

jeertmans:
> It's waiting on me to review it. In general, my review latency is pretty high.

No problem, take your time :-)

I think I have updated the code accordingly.

BurntSushi (approved):
LGTM

BurntSushi:
In the future, for changes this small, they should be one commit. If you need to make changes after opening the PR, then modify the commit via `git commit --amend` and force push.

Basically, treat commits like code. Optimize for the reader.

PR #2391, opened 2023-01-05
fix(ignore): change msrv to 1.63

nshaaban-cPacket:
Recent changes introduced by version 0.4.19 require rust 1.63. This formalizes that requirement.

See https://github.com/BurntSushi/ripgrep/commit/e95254a86f484eec663be4714924d91d3cf39cb8#commitcomment-95170910 for more context

See https://github.com/rust-lang/api-guidelines/discussions/231 for community discussion for how changes to MSRV impact versioning. Would be nice to bump MINOR to avoid compatibility issuers, however ignore is pre-v1, so you can do what you want.

### DIFF
--- a/crates/ignore/Cargo.toml
+++ b/crates/ignore/Cargo.toml
@@ -13,6 +13,7 @@ readme = "README.md"
 keywords = ["glob", "ignore", "gitignore", "pattern", "file"]
 license = "Unlicense OR MIT"
 edition = "2018"
+rust-version = "1.63"
 
 [lib]
 name = "ignore"


atouchet:
If the ignore crate has the same MSRV as ripgrep I believe this should be 1.65 rather than 1.63: https://github.com/BurntSushi/ripgrep/blob/master/Cargo.toml#L20

nshaaban-cPacket:
1.63 works with the current version of ignore.

iitalics:
encountered this when trying to build `cargo-make`, so FYI this affects more than just ripgrep

BurntSushi:
> however ignore is pre-v1, so you can do what you want

This is not accurate in the Cargo ecosystem. Bumping the leftmost non-zero number in a version results in a semver incompatible change. `ignore`'s leftmost non-zero number is the minor version, so bumping that would result in putting out a semver incompatible release. I have no desire to do such things and will never do it just because of an MSRV bump. So even though `ignore` is pre-v1, it is not "you can do what you want." Bumping the patch version still needs to produce a semver compatible release.

> If the ignore crate has the same MSRV as ripgrep I believe this should be 1.65 rather than 1.63

I'm not quite ready to make a decision about this. And in particular, whatever decision I do make, I'd like to apply it to all the crates in this repo. Maintaining different MSRVs for each sounds like a maintenance burden to me.

nshaaban-cPacket:
Thanks for your care and maintenance, even if it is just a hobby project!

>  I have no desire to do such things and will never do it just because of an MSRV bump. 

Don't blame ya, MSRV compatibility is an unsolved issue. https://blog.illicitonion.com/rust-minimum-versions-semver-is-a-lie/

> I'm not quite ready to make a decision about this. And in particular, whatever decision I do make, I'd like to apply it to all the crates in this repo. Maintaining different MSRVs for each sounds like a maintenance burden to me.

One MSRV for all crates sounds sensible, that way the CI can catch it and us developers don't need to think about it.

FYI tools like https://crates.io/crates/cargo-msrv exists to help detect which rust version is required for a given codebase.

Have a good one!

BurntSushi:
Thank you. :-) I'm on libs-api and have been part of the MSRV discussions for many years.

nshaaban-cPacket:
> I'm on libs-api and have been part of the MSRV discussions for many years.

Ouch. Thanks for that too!

PR #2396, opened 2023-01-12
ignore: respect local .git/config excludes

### LINKED ISSUE #2392: ripgrep not respecting global .gitignore
```
ripgrep 13.0.0
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

`brew`

macOS Ventura 13.1

#### Describe your bug.

I have some local backup files in my dev repo and I'd like to ignore them from all `rg` functionalities. Since they exist on my local machine I don't add them to the repo tracked .gitignore, so I configured a global gitignore but it seems to not work. https://github.com/BurntSushi/ripgrep/issues/9 seems to create support for it, but it's quite old already so perhaps it's now broken.

```sh
$ git config core.excludesFile
~/.gitignore

$ cat ~/.gitignore
*.sql

$ rg --files | rg backup
backup
backup_21_07_2022.sql
backup_11-12-22.sql
db_backup.sql

$ git add db_backup.sql
The following paths are ignored by one of your .gitignore files:
db_backup.sql
hint: Use -f if you really want to add them.
hint: Turn this message off by running
hint: "git config advice.addIgnoredFile false"

$ echo "*.sql" >> .gitignore

$ rg --files | rg backup
backup
```

You can see that git refuses to add the file since it is ignored by the global .gitignore and then that ripgrep would only respect it if is present in the working directory.

Patterns from global .gitignore are ignored by ripgrep

Patterns from global .gitignore are respected by ripgrep

#### What do you think ripgrep should have done?
Ignore the patterns present in the global .gitignore

zaidhaan:
Currently, ripgrep respects `$HOME/.gitconfig` and `$XDG_CONFIG_HOME/git/config`, but not `.git/config` (the "local" or "project" config). Git makes it such that `.git/config` overrides the other two options, while still respecting the `./.gitignore`. This commit brings this functionality to ripgrep (while also cleaning a few repetitive segments of code).

To demonstrate with a reproducible example (of course, be careful with overriding any files):
```sh
$ mkdir test && cd test
$ touch file file.customexclude file.globalignore file.pwdignore
$ echo '*.pwdignore' > .gitignore
$ mkdir -p /tmp/tempdir && echo '*.customexclude' > /tmp/tempdir/.gitignore
$ git config core.excludesfile /tmp/tempdir/.gitignore
$ echo '*.globalignore' > ~/.gitignore
$ git config --global core.excludesfile ~/.gitignore
$ git status
...
	.gitignore
	file
	file.globalignore
...
$ rg --files # without this patch
file.customexclude
file
$ rg --files # with the patch
file
file.globalignore
```

Note that the global ignore file is [intended to be ignored](https://stackoverflow.com/questions/8801729/is-it-possible-to-have-different-git-configuration-for-different-projects) when a `.git/config` exists (as demonstrated by the output of git).

Fixes #2392

### DIFF
--- a/crates/ignore/src/dir.rs
+++ b/crates/ignore/src/dir.rs
@@ -20,7 +20,9 @@ use std::io::{self, BufRead};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, RwLock};
 
-use crate::gitignore::{self, Gitignore, GitignoreBuilder, parse_excludes_file, read_file_contents};
+use crate::gitignore::{
+    self, parse_excludes_file, read_file_contents, Gitignore, GitignoreBuilder,
+};
 use crate::overrides::{self, Override};
 use crate::pathutil::{is_hidden, strip_prefix};
 use crate::types::{self, Types};
@@ -250,10 +252,11 @@ impl Ignore {
         } else {
             has_git_config = true;
             match resolve_git_commondir(dir, git_type) {
-                Ok(git_dir) => {
-                    read_file_contents(&git_dir.join("config")).and_then(|config_contents| {
+                Ok(git_dir) => read_file_contents(&git_dir.join("config"))
+                    .and_then(|config_contents| {
                         parse_excludes_file(&config_contents)
-                    }).map(|excludes_file| {
+                    })
+                    .map(|excludes_file| {
                         let (m, err) = create_gitignore(
                             &dir,
                             &dir,
@@ -262,11 +265,11 @@ impl Ignore {
                         );
                         errs.maybe_push(err);
                         m
-                    }).unwrap_or_else(|| {
+                    })
+                    .unwrap_or_else(|| {
                         has_git_config = false;
                         Gitignore::empty()
-                    })
-                }
+                    }),
                 Err(err) => {
                     errs.maybe_push(err);
                     has_git_config = false;
@@ -450,7 +453,14 @@ impl Ignore {
             mut m_gc,
             mut m_gi_exclude,
             mut m_explicit,
-        ) = (Match::None, Match::None, Match::None, Match::None, Match::None, Match::None);
+        ) = (
+            Match::None,
+            Match::None,
+            Match::None,
+            Match::None,
+            Match::None,
+            Match::None,
+        );
         let any_git =
             !self.0.opts.require_git || self.parents().any(|ig| ig.0.has_git);
         let mut saw_git = false;
--- a/crates/ignore/src/dir.rs
+++ b/crates/ignore/src/dir.rs
@@ -20,7 +20,7 @@ use std::io::{self, BufRead};
 use std::path::{Path, PathBuf};
 use std::sync::{Arc, RwLock};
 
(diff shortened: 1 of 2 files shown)


BurntSushi:
After a quick look I don't think this is right? It appears to assume that `.git` must be in the CWD. But whether `.git/config` is used or not shouldn't depend on whether ripgrep is run at the root of the repo. It might be run above the root or even below, but in both cases, `.git/config` should be respected. This will unfortunately probably require quite a gnarly refactor that I don't think I have the bandwidth for right now.

zaidhaan:
> It appears to assume that .git must be in the CWD

Whoops, you're right.
```sh
$ mkdir subdir && cd subdir
$ touch subfile subfile.customexclude subfile.globalignore subfile.pwdignore
$ # only subfile and subfile.globalignore should be seen
$ rg --files
subfile
subfile.customexclude
$ rg-patched --files
subfile
subfile.customexclude
```

I'm not too familiar with the code but it would appear that in dirs.rs there's going to have to be a matcher specifically made for `.git/config`'s `excludesfile` then? Which I suppose would bring in a bit of logic (`parse_excludes_file`) into dirs.rs. After such a matcher is made that would produce a `Gitignore`, then some logic could be used to decide whether to use the `global_matcher` if the config (which would override it) already exists.

Is this somewhere in the general right direction (at least with the way the code is now)? I've been working a bit on an approach like what I just described, might be able to make another commit in a bit.

(This admittedly isn't that elegant since anything to do with the git dir, git config and excludesfile is done in gitignore.rs, with the apparent exception of .git/info/exclude logic which is handled in dir.rs)

There's still a lot of cases I haven't explored (would global ignore be used if ~/.gitconfig exists with an excludes file, etc.), so I'll mark this PR as a draft. I'll try to see how git behaves in all cases this weekend and document it here, that way if this PR doesn't land at the very least it might help whoever ends up implementing this.

zaidhaan:
The above commit removes the incorrect CWD parsing and moves the `.git/config` excludesfile ignoring logic to `add_child_path` in dirs.rs (which looks at all the changes in `IgnoreOptions`), I didn't create another `IgnoreOptions` option, but instead made it such that it respects the `git_exclude` option (which prior to this only looked at `.git/info/exclude`, but not would also look at `./git/config`s excludesfile).

Haven't tested this that thoroughly, still a WIP, but so far it does work in subdirectories since it correctly identifies the git path.

```sh
$ # all the setup stuff
$ mkdir subdir && cd subdir
$ touch subfile subfile.customexclude subfile.globalignore subfile.pwdignore
$ git add -A

$ git ls-files
subfile
subfile.globalignore

$ rg --files
subfile
subfile.customexclude

$ rg-patched --files
subfile
subfile.globalignore
```

BurntSushi:
Thanks for the work on this, but I'm going to close it. The `ignore` crate badly needs an overhaul, and hopefully I can address this use case when I do that. But I'm not sure when it's going to happen. As of now, I can't really stomach adding more stuff like this because I can't be confident that it won't break other stuff.

I've opened #2553 to track this particular use case so that it's factored into the overhaul.

PR #2397, opened 2023-01-15
Added Clone and Debug for OverrideBuilder

vallentin:
I couldn't find a reason for this being intentional. `Clone` and `Debug` is also already implemented for `Override`, `Gitignore`, and `GitignoreBuilder`. But not for `OverrideBuilder`.

In my own project, I have a custom builder containing an `OverrideBuilder`, and would like to be able to add `Clone` to my builder.


BurntSushi:
This is on crates.io in `ignore 0.4.20`.

PR #2403, opened 2023-01-27
docs: add examples for ignore::WalkBuilder::sort_by_*

tshepang:


BurntSushi:
These examples aren't really formatted properly and there's no prose around them describing it.

Normally I'd just take this PR and fix them up, but I'm not keen on investing polishing effort into `ignore` at the moment, because I want to give it an overhaul.

PR #2405, opened 2023-01-27
deps: require secure versions of regex

robjtede:
addresses warnings seen here: https://deps.rs/repo/github/BurntSushi/ripgrep#vulnerabilities


BurntSushi:
The lock file already has a more up to date `regex` crate used.

Both of those "vulnerabilities" are false positives in the context of ripgrep. The `thread_local` vulnerability doesn't impact ripgrep because it doesn't use the iteration APIs. The `regex` vulnerability is ReDoS which doesn't make sense in the context of ripgrep since it's a command line tool.

I'm sure these versions will get updated at some point, but I don't see any compelling reason to do that now.

PR #2408, opened 2023-02-01
crates/core/app.rs: Make "-D" an alias for "--max-depth"

jurajlutter:
For better usability (at least for me), add an alias "-D" for the "--max-depth" option.


jurajlutter:
A very friendly "ping"?

BurntSushi:
My initial instinct here is to say no, because the number of available short flags remaining is very small. I'm not convinced this flag is common enough to warrant one of those remaining short flags.

PR #2415, opened 2023-02-10
ignore/types: add *.sln for msbuild

dmringo:
.sln is the extension for Visual Studio Project Soltion files, one of the file types accepted as inputs by MSBuild.

I'm not sure if/how this kind of thing is tested, but I'd be happy to add a test if required.


BurntSushi (approved):
LGTM. The specific default rules aren't tested explicitly. The underlying glob engine is tested, but otherwise, the actual default type rules are mostly just "correct by visual inspection."

PR #2426, opened 2023-02-22
Add How is ripgrep faster than GNU grep? to FAQ

four43:


four43:
Totally up to you! Reddit can just be a poor archive and having info like
this out there is great. As we saw from that mailing list.

BurntSushi:
See message above. In theory I'm not opposed to an FAQ item about this, but the answer is hard to write succinctly.

PR #2431, opened 2023-02-26
"ignore" crate: Parallel walker should ignore broken symlinks

ruffe972:
This fixes https://github.com/sharkdp/fd/issues/746. "fd": uses "ignore" crate from this repo. Steps to reproduce:

- Create parallel walker with "follow links" option enabled.
- Create broken symlink with the name broken_symlink, for example.
- Add "broken_symlink" to ignored files.
- Get walker dir entry results.

Expected: walker doesn't return dir entry result for the link.
Actual: walker returns error result for the link.

Note that I didn't fix similar bug in single-threaded walker. I'm new to Rust so it's a bit difficult for me, but if it's required for this PR to be merged, I'll fix single-threaded walker too.

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -1543,6 +1543,9 @@ impl<'s> Worker<'s> {
         };
         let is_symlink = dent.file_type().map_or(false, |ft| ft.is_symlink());
         if self.follow_links && is_symlink {
+            if should_skip_entry(ig, &dent) {
+                return WalkState::Continue;
+            }
             let path = dent.path().to_path_buf();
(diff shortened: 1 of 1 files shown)


BurntSushi:
I've opened #2552 for this. I'm not sure this is the right way to fix this issue, and the fact that the single threaded case isn't fixed means I don't want to bring this in as-is. I looked and fixing the single threaded case appears tricky.

I'm going to close this for now in favor of tracking #2552 as a bug. Since I don't consider this to be a high priority bug, I'm going to defer fixing this until I've had a chance to overhaul `ignore`'s internals.

PR #2432, opened 2023-02-27
ignore/types: add USD default types

marksisson:
[USD](https://www.pixar.com/usd) is open-source software that can robustly and scalably interchange 3D scenes that may be composed of many different assets, sources, and animations, while fostering highly collaborative workflows.


BurntSushi:
This is not particularly clear to me.

Firstly, the better link seems to be this: https://openusd.org/release/index.html

Second, ripgrep generally only works with plain text formats, and it's not at all clear that USD is a plain text format. I had to hunt around to find an actual example: https://github.com/PixarAnimationStudios/OpenUSD/blob/release/pxr/usd/usd/testenv/testUsdBug141491.testenv/root.usda

So I guess this LGTM.

PR #2437, opened 2023-03-01
ignore/types: add Gentoo eclass type

thesamesam:
Eclasses are "ebuild libraries" and generally if you're filtering for/filtering out an ebuild/eclass, you don't want the other either.

Followup to 4dfea016b915bb1e88679361de83a91e60447835.


thesamesam:
Thanks!

PR #2450, opened 2023-03-13
Update default_types.rs for elixir

angrycandy:
Elixir https://hexdocs.pm/phoenix_live_view/assigns-eex.html:

Phoenix template language is called HEEx (HTML+EEx). Those templates are either files with the .heex extension...


angrycandy:
> There is also `leex`. but both of these extensions are for Phoenix, not core Elixir.

@Stratus3D do you think supporting `heex` & `leex` degrades the non-Phoenix Elixir programming experience?

I don't think so, and I think it enhances the Phoenix experience.

BTW, [Upgrading and deprecations]:(https://github.com/phoenixframework/phoenix_live_view/blob/main/CHANGELOG.md#upgrading-and-deprecations)
`
The main deprecation in this release is that the ~L sigil and the .leex extension are now soft-deprecated. The docs have been updated to discourage them and using them will emit warnings in future releases. We recommend using the ~H sigil and the .heex extension for all future templates in your application. You should also plan to migrate the old templates accordingly using the recommendations below.
`
I think I should update the PR to include `leex`.

Stratus3D:
I don't think it's a negative to treat `heex` as Elixir no, but I figured I'd point out the distinction in case it mattered.

rodrigues:
It would be nice also to support [Livebook](https://livebook.dev) (`*.livemd` - Elixir code notebooks) 👌

Glad the CLI allows type customisation in the meantime:

```fish
function au
  rg --type-add "elixir:*.heex" --type-add "elixir:*.livemd" -telixir $argv
end

```

PR #2453, opened 2023-03-14
Update rg.1.txt.tpl

ruppde:
Clarify arguments in config files

### DIFF
--- a/doc/rg.1.txt.tpl
+++ b/doc/rg.1.txt.tpl
@@ -186,7 +186,7 @@ ripgrep supports reading configuration files that change ripgrep's default
 behavior. The format of the configuration file is an "rc" style and is very
 simple. It is defined by two rules:
 
-    1. Every line is a shell argument, after trimming whitespace.
+    1. Every shell argument needs to go in one line, with e.g. '-j 4' counting as two arguments.
     2. Lines starting with *#* (optionally preceded by any amount of
        whitespace) are ignored.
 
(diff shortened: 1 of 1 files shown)


doc/rg.1.txt.tpl
```diff
@@ -186,7 +186,7 @@ ripgrep supports reading configuration files that change ripgrep's default
 behavior. The format of the configuration file is an "rc" style and is very
 simple. It is defined by two rules:
 
-    1. Every line is a shell argument, after trimming whitespace.
+    1. Every shell argument needs to go in one line, with e.g. '-j 4' counting as two arguments.
```
BurntSushi, line 189: This should follow the style consistent with the surrounding content and be wrapped to 79 columns (inclusive).

Also, this phrasing kind of just seems to confuse things even more from my perspective. The point is that something like `-j 4` would be _one_ argument when put on the same line. So it needs to be:

```
-j
4
```

or

```
-j4
```

So I'm not sure if this is the right phrasing here...

BurntSushi (changes requested):
Whatever changes we make here should also go in the GUIDE too.

doc/rg.1.txt.tpl
ruppde: Fixed it

The usual perspective is "I want -j 4 to go into the config file" but the text was written the other way around: How the config file is parsed.

BurntSushi:
I ended up going with different wording. The fact that the docs describe how the config file is parsed is intentional and I don't want to change that. I also feel like the examples in the man page already clarified your use case pretty well. But I added one more example.

PR #2469, opened 2023-03-21
Add support for searching in Docker compose files

manu-paylead:
Default file is docker-compose.yml and the documentation mentions overrides in the form of docker-compose.*.yml


BurntSushi (approved):
Thanks!

PR #2481, opened 2023-03-30
Replace atty with is-terminal

tottoto:
Replaces atty with is-terminal as atty is unmaintained.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,17 +12,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "atty"
-version = "0.2.14"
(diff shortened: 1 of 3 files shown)


BurntSushi:
Thanks for the PR! I'm going to stick with `atty` for now. The `is-terminal` crate brings in too many dependencies IMO.

It does look like the `IsTerminal` API in std will be landing soon. Once that's stable, we can switch to that and drop `atty` completely.

tottoto:
I got it. Thanks for your review.

tottoto:
I understand. It makes sense. Thanks for your review.

PR #2482, opened 2023-04-01
Make gitignore.rs's `gitconfig_excludes_path` public

Eric-Arellano:
In Pantsbuild, we use `GitignoreBuilder` to determine which files our file watcher should look at.

We want to improve the file watcher to consider the global gitignore file. It would be exceptionally useful if we could reuse your logic: https://github.com/pantsbuild/pants/pull/18649

P.S. Thanks for ripgrep! You've saved me so much time 🙌


Eric-Arellano:
Thank you!

PR #2490, opened 2023-04-13
Fix code block label in docs

BD103:
I changed the attribute (info string) `ignore` to `text`. `ignore` is used for Rust code that should not be run and also receives a label in the generated documentation, which is incorrect for the modified block. Changing it to `text` removes syntax highlighting (though the previous highlighting was incorrect).

Additional links:
- [Rustdoc code attributes](https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html#attributes)
- [Commonmark spec on info string](https://spec.commonmark.org/0.30/#info-string)


BD103:
Closing due to inactivity. Should this ever gain attention, I can create another PR.

PR #2503, opened 2023-05-02
deps: update to aho-corasick 1.0.1

poliorcetics:
I think nothing changed in public APIs, but maybe MSRV has been bumped by this


BurntSushi:
Thank you, but I prefer to do all dependency changes/upgrades myself. See https://github.com/BurntSushi/ripgrep/pull/2390 for example.

PR #2508, opened 2023-05-06
ignore/types: add 'graphql'

jparise:
GraphQL file extensions: .graphql and .graphqls (schema)

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -84,6 +84,7 @@ pub const DEFAULT_TYPES: &[(&str, &[&str])] = &[
     ("gn", &["*.gn", "*.gni"]),
     ("go", &["*.go"]),
     ("gradle", &["*.gradle"]),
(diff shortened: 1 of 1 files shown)


jparise:
Whoops, I hadn't noticed #2439 was already proposing a `graphql` type. I think the file extensions I chose (`.graphql`, `.graphqls`) are slightly more correct, however:

- `.graphql` is the canonical GraphQL file extension (graphql/graphql-spec#203)
- `.graphqls` is the recommended GraphQL SDL file extension (https://www.apollographql.com/docs/kotlin/essentials/file-types/)

I excluded `.gql`, a commonly used abbreviated extension. It could conflict with other existing types (e.g. https://filext.com/file-extension/GQL), although I doubt many people are going to use `-tgraphql` in contexts where non-GraphQL `.gql` files appear.

BurntSushi:
I'll go with this PR for now, since it's more conservative.

We can add `.gql` later if there's demand. It's inevitable that some extensions will conflict. That's OK.

PR #2515, opened 2023-05-13
grep-cli: Replace unmaintained `atty` dep with `is-terminal`

Enselic:
Hi! The `atty` crate is [unmaintained](https://rustsec.org/advisories/RUSTSEC-2021-0145.html) and the de-facto replacement is `is-terminal`. This PR replaces `atty` with `is-terminal`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -12,17 +12,6 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


BurntSushi:
Duplicate of #2481.

PR #2525, opened 2023-06-03
Upgrade `aho-corasick` to `1.0.1` in `globset`

madkinsz:
Hi!

I noticed that `aho-corasick` was still on version `0.7.20` so it's getting built twice in downstream projects that use the latest version alongside `globset`.

Updating required:

- Change from `AhoCorasick::new_auto_configured` to `AhoCorasick::new`
- A new `GlobSet::ErrorKind` for handling of `AhoCorasick::BuildError`

I named the error `MatchBuild` following the pattern used for `Regex` compilation errors.

I'm new to Rust so any feedback is welcome!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -142,7 +142,7 @@ checksum = "d2fabcfbdc87f4758337ca535fb41a6d701b65693ce38287d856d1674551ec9b"
 name = "globset"
 version = "0.4.10"
 dependencies = [
- "aho-corasick 0.7.20",
+ "aho-corasick 1.0.1",
  "bstr",
  "fnv",
  "glob",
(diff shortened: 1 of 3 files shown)


BurntSushi:
Hmmm so this isn't quite right. I don't think we need a new error type. The previous code would panic if `AhoCorasick` failed to build, so it seems fine to keep those semantics.

Otherwise, there are other concerns here:

1. I generally prefer to do all dependency upgrades myself so that I can audit any and all changes if necessary. In this case, I'm also the author of `aho-corasick`, so it's fine.
2. I'm planning to transition ripgrep to `regex 1.9` once https://github.com/rust-lang/regex/pull/978 lands. In the mean time, there may indeed be some duplicate dependencies (notably `regex-syntax` and `aho-corasick`) getting compiled. Hopefully it gets resolved soon.

madkinsz:
👍 thanks for the reply!

PR #2529, opened 2023-06-07
Add `.cml` to default_types.rs

tengyifei:
Use on Fuchsia to mean "component manifest language": [1].

It's also the "Chemical Markup Language" [2] and shares the same extension.

[1]: https://fuchsia.dev/reference/cml?hl=en
[2]: https://www.reviversoft.com/en/file-extensions/cml


BurntSushi:
This probably isn't right. We can't have the same name referring to two semantically different file types. The problem arises when there are diverging file extensions.

Probably we should make `cml` refer to Fuchsia's "component manifest language" and then have a distinct `chemical` abbreviation for "chemical markup language."

PR #2539, opened 2023-06-19
Fix rust-version in Cargo.toml

jschwe:
ripgrep requires Rust 1.70, which is already documented in the Readme. This change adjusts the Cargo.toml to reflect this.


ltrzesniewski:
There's still a mention of Rust 1.65 in the readme BTW (in the *Building* section), you could fix that in this PR as well. 🙂

memark (approved):
LGTM

PR #2556, opened 2023-07-09
ignore/types: add Windows Command Prompt files

mataha:
This PR adds `*.bat` and `*.cmd` file types.

In doing so, it makes a distinction between batch files (old standard from the MS-DOS era) and command scripts (new flavor - can operate on batch files, although `*.cmd` is preferred for various reasons, the main one being batch files will set `ERRORLEVEL` following inconsistent MS-DOS style rules).

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -24,6 +24,7 @@ pub const DEFAULT_TYPES: &[(&[&str], &[&str])] = &[
     (&["ats"], &["*.ats", "*.dats", "*.sats", "*.hats"]),
     (&["avro"], &["*.avdl", "*.avpr", "*.avsc"]),
     (&["awk"], &["*.awk"]),
+    (&["bat", "batch"], &["*.bat"]),
     (&["bazel"], &[
         "*.bazel", "*.bzl", "*.BUILD", "*.bazelrc", "BUILD", "MODULE.bazel",
         "WORKSPACE", "WORKSPACE.bazel",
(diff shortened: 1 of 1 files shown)


crates/ignore/src/default_types.rs
```diff
@@ -24,6 +24,7 @@ pub const DEFAULT_TYPES: &[(&[&str], &[&str])] = &[
     (&["ats"], &["*.ats", "*.dats", "*.sats", "*.hats"]),
     (&["avro"], &["*.avdl", "*.avpr", "*.avsc"]),
     (&["awk"], &["*.awk"]),
+    (&["batchfile"], &["*.bat"]),
```
BurntSushi, line 27: Would `bat` be better here?

mataha:
> Is it possible to provide a source for this change? Ideally something from Microsoft's docs.

[This][1] is the only credible source though.

[1]: https://groups.google.com/g/microsoft.public.win2000.cmdprompt.admin/c/XHeUq8oe2wk/m/LIEViGNmkK0J#i106

crates/ignore/src/default_types.rs
mataha: May I suggest `bat` in addition to `batchfile`? As these are commonly called "batch files", even by Windows.

BurntSushi: I suppose that's fine.

BurntSushi: Well, I'd prefer `batch` and `bat`, not `batchfile`. `file` seems redundant in this case and no other file type uses `file` in the name.

mataha: `bat` and `batch` it is then.

BurntSushi (approved):
Thanks!

PR #2560, opened 2023-07-17
use issue forms

okdana:
Would you want something like this?

[Creating issue forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms)

I thought it might help reduce the number of invalid/incomplete reports

Here's one i put together, probably needs edited for formatting/wordiness. I don't think you can see how it renders from this PR, so i've attached a screen-shot below

Obv the feature-request template could be converted as well if you like it

Possible concerns:

* This feature is considered beta (though it's been available for 2 years now)
* The formatting of the file is a little weird due to interactions between YAML and GFM
* It could be annoying for more advanced users
* I'm not sure how scalable it is to list out common issues on the report page

form


BurntSushi (approved):
Nice, thank you for doing this. I didn't even know about this feature of GitHub.

I will admit that I am a little unsure of doing this. I always hate filling out forms like this myself. But it's worth a try because I spend an amazing amount of time just trying to get people to include relevant information. (Although that's not specific to ripgrep.)

PR #2561, opened 2023-07-18
ignore/types: add TCC files

mataha:
This PR adds file types for TCC, a `cmd.exe` extension.

Extensions are the same as for `cmd` itself + `*.btm`[1].

[1]: https://jpsoft.com/help/batchtype.htm


mataha:
That's fair (though I believe the runtime is free for unlimited use).

Issue #275, opened 2016-12-08
Windows: slash vs backslash in matched filenames 


ljantzen:
I'm on 0.2.3. 

I am using rg in git-bash shell on a win10 box.  It works, but the filenames /directories it outputs follows the windows/dos standard, using the backslash character as directory separator.  This makes it difficult to copy/paste the filename into other commands in the bash shell.  

I could do something like 

     rg -l SearchPattern | sed -e 's/\\/\//g'

but that loses the coloured output and context.

Is there a way to tell rg that I want slash as directory separator character?

BurntSushi:
I doubt it. It shows paths as they are supposed to be shown for your particular platform. Why doesn't your Bash shell handle Windows paths?

BurntSushi:
It's not clear to me what specific problem the OP is facing here and whether ripgrep is the one to blame. If there's more clarity that can be provided, please feel free to re-open!

ljantzen:
I agree that running bash on windows perhaps isn't the most common use-case, but it is pretty common at least around where I am ( .net appl dev using git ).   I have installed msysgit, which is the most common git command line client on windows.  It provides a pretty standard bash shell, where I can access the usual bash/unixy commands and run shell scripts.  As a former unix/linux based developer converting to .net I find that is a tremendous productivity benefit. I live in bash, even when I'm on windows. 

I was really excited when I found ripgrep.  As a unix junkie, I find it really hits the sweet spot.  It even works on windows on the command line. Yeah!  Bash? No problem, it rips through the source code in no time. When ripgrep finds something interesting, I usually want to open the file to edit or take a closer look.  But, bash uses forward slashes, whereas dos/windows uses backslash as the directory separator character.  In bash the backslash is the escape character, and requires special handling.  '\t' is the tab character.  So now I need to copy and paste the direcory/filename and manually replace the backslashes with slashes before I can open the file in vim or whatever. 

Cutting and pasting a standard windows directory name into a bash shell simply does not work as you would hope.  Is that a missing feature of bash? I could argue that with the bash/unix maintainers, but I doubt I would make much headway in that discussion. 

I would really appreciate a '-XUseUnixDirsep' option or something similar to force ripgrep to ouput src/test/apptest.cs instead of src\test\apptest.cs.   Would you accept a pull request to that effect?

BurntSushi:
Is ripgrep really the only tool with this problem? How do you deal with other tools that output paths using the separator defined by your platform?

[...]

Issue #1667, opened 2020-08-27
document Windows locale and path separator settings


SandraEickel:
#### What version of ripgrep are you using?

ripgrep 12.1.1 (rev 7cb211378a)
-SIMD -AVX (compiled)
+SIMD -AVX (runtime)

#### How did you install ripgrep?

Unzipped Windows 64 bit versions (both MSVC and GNU behave identical, using MSVC).

#### What operating system are you using ripgrep on?

Windows 10 Pro
Build 18362.19h1_release.190318-1202

#### Describe your bug.

Inconsistent output of umlauts on Windows depending on running ripgrep from CMD or Git Bash.

#### What are the steps to reproduce the behavior?

Both shells are configured to use Lucida Console font.

**Using standard CMD (PowerShell not tested):**
cd ripgrep-test

type subdir\ExampleWithUmlauts.cs
- Displays garbled umlauts

rg SubjectCodes
- Displays file name with Windows-like backslash
- Displays correct umlauts

**Using Git Bash (MINGW64):**
cd ripgrep-test

cat subdir/ExampleWithUmlauts.cs
- Displays correct umlauts

rg SubjectCodes
- Displays file name with unwanted backslash instead for UNIX-like (forward) slash
- Displays garbled umlauts

#### What is the actual behavior?

CMD:
```
C:\Users\Sandra.Eickel\Documents\ripgrep-test>type subdir\ExampleWithUmlauts.cs
namespace ZUGFeRD_Test
{
    class ZugFerd1ExtendedWarenrechnungGenerator
    {
        private InvoiceDescriptor _generateDescriptor()
        {
            desc.AddNote("Es bestehen Rabatt- oder Bonusvereinbarungen.", SubjectCodes.AAK, ContentCodes.ST3);
            desc.AddNote("Der Verk├ñufer bleibt Eigent├╝mer der Waren bis zu vollst├ñndigen Erf├╝llung der Kaufpreisforderung.", SubjectCodes.AAJ, ContentCodes.EEV);
            desc.AddNote("MUSTERLIEFERANT GMBH BAHNHOFSTRASSE 99 99199 MUSTERHAUSEN Gesch├ñftsf├╝hrung: Max Mustermann USt-IdNr: DE123456789 Telefon: +49 932 431 0 www.musterlieferant.de HRB Nr. 372876 Amtsgericht Musterstadt GLN 4304171000002 WEEE-Reg-Nr.: DE87654321",
                         SubjectCodes.REG);
[...]

PR #2562, opened 2023-07-19
windows: set default path separator to '/' in MSYS

EJammy:
Adapted from https://github.com/sharkdp/fd/pull/730
Also see https://github.com/sharkdp/fd/issues/537#issuecomment-757519135

This removes the need to use `--path-separator` for MSYS and MSYS2 environments (such as Git Bash), as per #275, but not for Cygwin (see linked pr and comment).

Not sure if this is welcomed, since it is handling a special case for specific environment. But I think this is a good improvement as it aligns with the behavior of `fd` and removes error when running commands such as `rg foobar --files-with-matches | xargs ` without `--path-separator`.

### DIFF
--- a/crates/printer/src/util.rs
+++ b/crates/printer/src/util.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::env;
 use std::fmt;
 use std::io;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


BurntSushi:
In terms of implementation, I think this logic should probably live in ripgrep proper and not inside `grep-printer`.

In terms of idea, I'm less sure about this. It seems like it's easy enough to work-around by defining an alias or a config file to set the default path separator yourself. It doesn't feel right to be doing this automatically behind your back. And this is a very subtle (IMO) change in behavior and it's not clear to me what the impact of it would be.

EJammy:
I'm aware of the workaround. I proposed this change because it took me a while to figure out the error after piping the results of `rg --files-with-matches`, while most tools under MSYS, like fd and git, "just works". I installed ripgrep through MSYS2 with `pacman -S mingw-w64-x86_64-ripgrep`, so I expected it to work well under that environment.

However, this method of installation is not documented, so I figured it might not have the best support. And I have to agree when I discover this solution it also felt a bit wrong to me too. Might be best to just leave it for the user to configure and improve documentation (#1667).

BurntSushi:
Apologies for my potential ignorance here, but could you say more about the relationship between cygwin and MSYS? In my brain, I seem to remember them being connected or the same in some way, but I find myself unable to articulate precisely their relationship.

EJammy:
I'm not an expert either. I've only recently started using MSYS2 to work in windows and have not touched cygwin before.

But from my understanding, MSYS is built on top of cygwin and provides tools to build and run native windows software. It also has a package manager (`pacman`). It's also worth noting that git bash (from git for windows) is based on MSYS2, so when people mention git bash it's just repackaged MSYS2.

There's more info on their home page: https://www.msys2.org/
Some additional differences: https://www.msys2.org/wiki/How-does-MSYS2-differ-from-Cygwin/

BurntSushi:
All righty, thanks. I think I'm going to close this then.

I'm not necessarily 100% opposed to something like this, but I just don't have a ton of visibility into its impact. And in general, I've found the whole cygwin/MSYS2 environment to be pretty difficult to fit into. The fundamental problem is that there is no cygwin Rust target, so Rust programs in cygwin don't get linked with the cygwin POSIX emulation layer (or whatever it's called). They stay native Windows programs. So it's best to think about ripgrep as _always_ a native Windows program and not something that is built for cygwin.

PR #2575, opened 2023-07-31
Adds .csproj to default_types.rs

vidarasberg:
Supports the .NET C# Project file extension


BurntSushi (approved):
LGTM. How many more C# types are we going to have? This makes 4 by my count.

vidarasberg:
> LGTM. How many more C# types are we going to have? This makes 4 by my count.

Thank you!

Haha yes but csharp and cs are both *.cs.

I now see that I could have used msbuild as *.csproj was included there. I hope it helps someone that now both options exist.

ltrzesniewski:
> I hope it helps someone that now both options exist.

I'll certainly use `-tcsproj` as it feels better than `-tmsbuild`, thanks! 🙂

`-tcsharp` is a bit redundant IMO as it's longer than `-tcs`.

PR #2598, opened 2023-08-25
implemented filelist to get list of paths

eyalk11:
Added option to get list of paths. I got into trouble trying to work with list of files in neovim (leaderf or other), when the list of files I wanted to search over was very long.  I believe it solves https://github.com/BurntSushi/ripgrep/issues/273 . 

See also https://vi.stackexchange.com/questions/42785/how-to-do-fuzzy-live-grep-on-git-files/42915#42915

### DIFF
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -570,6 +570,7 @@ pub fn all_args_and_flags() -> Vec<RGArg> {
     flag_engine(&mut args);
     flag_field_context_separator(&mut args);
     flag_field_match_separator(&mut args);
+    flag_file_list(&mut args);
     flag_file(&mut args);
     flag_files(&mut args);
     flag_files_with_matches(&mut args);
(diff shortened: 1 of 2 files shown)


BurntSushi:
Thank you, but there are numerous issues with this PR as-is. The flag doesn't follow ripgrep's naming convention. Errors are squashed. The documentation for the flag is very insufficient. There are no tests. This also won't handle the case of file paths that contain invalid UTF-8. And perhaps more that I haven't noticed yet.

If you want to try to fix these things that would be great, but this is likely something I'll get to eventually myself if not.

BurntSushi (changes requested):
Thank you, but there are numerous issues with this PR as-is. The flag doesn't follow ripgrep's naming convention. Errors are squashed. The documentation for the flag is very insufficient. There are no tests. This also won't handle the case of file paths that contain invalid UTF-8. And perhaps more that I haven't noticed yet.

If you want to try to fix these things that would be great, but this is likely something I'll get to eventually myself if not.

eyalk11:
Thanks for the comment. I appreciate it. 

It was quick-and-a-bit-dirty patch . I don't really know rust (chatgpt wrote half of it) . 
I see what I can do. I can improve documentation and naming, but won't add tests. 
Also I already implemented it in my code with the naming as it is. 

I feel like in any case it would be better to leave it up to you.

BurntSushi:
OK, I'm going to close this out. I think this PR is pretty far from what this should ideally look like. I will very likely tackle this myself at some point.

PR #2599, opened 2023-08-26
removed dollar symbols from installation commands in order to make copying them useful

bwv847:


BurntSushi:
I'm personally not a huge fan of how this looks personally. It feels like it all just runs together. Although I do admit that this makes copying commands much more useful because of GitHub's "copy" button. On the other hand, I'm not necessarily a huge fan of encouraging folks to blindly run commands from a GitHub README, especially when they contain `sudo` (as many of these do). So... the extra speed bump might actually be desirable.

hey-leon:
@BurntSushi I understand you already have a stance on this but if I could play devils advocate here. I think there are two camps of developer on your argument:

__Camp A__: Thoroughly reads the command before even copying the command (who would benefit from it working without modification).

__Camp B__: Will blindly copy and paste see the error because of the `$` and will proceed to remove the dollar sign, still without fully considering the command.

While I appreciate there is probably also developers that don't fit these two camps. I think in practice the speedbump probably would not save many developers 🤷‍♂️.

Either way thanks for taking the time to review these PRs!

PR #2601, opened 2023-08-29
ci: replace mips with powerpc64, aarch64 and s390x

BurntSushi:
We drop our MIPS target because it no longer works.[1] We were previously using it as a means of testing ripgrep in a big endian environment. So to achieve that without MIPS, we test on powerpc64 and s390x. (No particular reason to do both, but why not.)

We also add aarch64 as a proxy for at least ensuring everything works for the same architecture as Apple silicon. It's not a guarantee that everything works, but it seems better than nothing until we can actually test Apple silicon in CI.

[1]: https://github.com/rust-lang/regex/commit/c788378d6fe407f4774df98a78436cea5d98525b


BurntSushi:
This PR is a good example of why I generally dislike using tooling in CI that I don't control. Something about how [Cross](https://github.com/cross-rs/cross) works changed and has cost me several hours of debugging this evening. I hit at least a few problems that are difficult for me to describe because I don't really understand them (and not all of which are necessarily Cross's fault), but the big one is that it looks like the `CROSS_RUNNER` environment variable is now useless. It, I believe, used to point to the `qemu` executable that one can run the cross compiled `rg` executable with inside of the Docker container. But now it just points to `qemu-user` in all of the images (except for native `x86_64`) and there is no `qemu-user` binary in the Docker image. So I had to do some detective work to figure out how the entire enterprise worked.

If Cross didn't do so damn much and save so much work when it came to cross-compilation I'd drop it in a heart beat.

PR #2612, opened 2023-09-25
Fix a few issues in the hyperlink docs

ltrzesniewski:
Many thanks for merging the hyperlinks feature and improving the implementation!

I read your changes and found a few minor issues in the docs, which I've fixed here.

FYI I tested your version on Windows and WSL, and everything works fine there. 👍

### DIFF
--- a/crates/core/app.rs
+++ b/crates/core/app.rs
@@ -1526,7 +1526,7 @@ fn flag_hyperlink_format(args: &mut Vec<RGArg>) {
 Set the format of hyperlinks to match results. Hyperlinks make certain elements
 of ripgrep's output, such as file paths, clickable. This generally only works
 in terminal emulators that support OSC-8 hyperlinks. For example, the format
-*file://{host}{file}* will emit an RFC 8089 hyperlink.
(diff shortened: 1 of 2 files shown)


crates/core/app.rs
```diff
@@ -1536,7 +1536,7 @@ put into a URI. Note that a path is guaranteed to start with a */*.
 
 *{host}*: Optional. This is replaced with your system's hostname. On Unix,
 this corresponds to calling *gethostname*. On Windows, this corresponds to
-calling *GetComputerNameExW* to fetch the system's \"physical DNS hostname.\"
+calling *GetComputerNameExW* to fetch the system's \"physical DNS hostname\".
```
BurntSushi, line 1539: I had it this was when I first wrote it, but it looked wrong to me (I'm in the US). Can you change it back please? And in the place above too.

ltrzesniewski: Oh, I didn't know this is the way it's written in the US. I reverted it.

BurntSushi: Yeah it's annoying. I sometimes put the period outside the quotes anyway.

BurntSushi (approved):

PR #2614, opened 2023-09-26
lots of little polishing

BurntSushi:
This PR brings a lot of the code into line with my current style, which has slightly shifted over the years. I've also dropped dependencies where possible. For example, `globset` no longer depends on the `fnv` crate, and instead just rolls the FNV hashing itself. Similarly, the `regex` crate is dropped as a dependency entirely, and we instead depend on `regex-automata` directly. This leads to the interesting result that ripgrep no longer depends on `regex` _anywhere_ in its tree(!), but it is still most convenient to say that it uses the regex crate when talking about the kind of regex flavor it supports.


crates/core/app.rs
```diff
@@ -1526,60 +1526,71 @@ fn flag_hyperlink_format(args: &mut Vec<RGArg>) {
...
 when ripgrep is not writing to a tty. (Unless one forces the issue by setting
-*--color=always*.)
+--color=always.)
+
+If you're searching a file directly, for example:
+
+    rg foo path/to/file
+
+then hyperlinks will not be emitted since the path given does not appear
+in the output. TO make the path appear, and thus also a hyperlink, use the
```
ltrzesniewski, line 1592:
```suggestion
in the output. To make the path appear, and thus also a hyperlink, use the
```

😅

PR #2615, opened 2023-09-29
`ignore`: Add `overrides::Glob::inner`

9999years:
This returns the inner glob, if any, and can be used to differentiate between matching against an explicit ignore glob and not matching against any globs.

### DIFF
--- a/crates/ignore/src/overrides.rs
+++ b/crates/ignore/src/overrides.rs
@@ -35,6 +35,17 @@ impl<'a> Glob<'a> {
     fn unmatched() -> Glob<'a> {
         Glob(GlobInner::UnmatchedIgnore)
     }
+
+    /// Return the matched glob if it exists.
+    ///
+    /// This can be used to differentiate between matching against an explicit ignore glob and not
+    /// matching against any globs.
+    pub fn inner(&self) -> Option<&gitignore::Glob> {
+        match self.0 {
+            GlobInner::UnmatchedIgnore => None,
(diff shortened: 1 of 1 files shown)


BurntSushi:
I'm going to pass on this. The override glob is purposely an opaque type and this exposes its implementation to a point that I'm not quite comfortable with. Also, the fact that it returns an `Option` makes this API a bit confusing and also somewhat dependent on implementation details.

9999years:
Why bother exposing it at all, then? The docs say "This is used to report information about the highest precedent glob that matched", but it doesn't report any information outside of the `Debug` implementation.

BurntSushi:
> Why bother exposing it at all, then?

I don't know. I designed the API many years ago and I've been unhappy with it ever since. It wouldn't surprise me if I exposed it for an unprincipled reason or if I just copied the design of the `gitignore` sub-module.

I would be open to exposing routines on an override glob that let you query information about it.

9999years:
Gotcha. Maybe it's worth reworking the API and pushing a 1.0 release? Let me know if you want to collaborate on it.

BurntSushi:
Oh it's absolutely worth doing. I plan on it. Problem is, I've been wanting to do it for years. The entire crate needs to be rewritten. It will take me some time to get to it. I can ping you in issues if that's okay when I start thinking about it and write something down. My first step will probably be to collect all of the issues on this tracker about the crate, including open bugs that have their issues in how the crate works.

9999years:
Sure, sounds good.

PR #2622, opened 2023-10-06
add rust dev container

craiglpeters:
This configuration will enable the use of dev containers to work on an environment that has Rust tools preconfigured from the [Rust base image](https://github.com/devcontainers/images/tree/main/src/rust). This configuration is based on the spec at https://containers.dev/ and works with the reference cli, the VS Code Dev Container extension (to run locally on a machine with Docker installed), or in various cloud based development environments including GitHub Codespaces.


BurntSushi:
Thanks! I don't anticipate using this at present, so I'm going to close this out. But thank you for filing it. :-)

craiglpeters:
No problem! Was just for illustration :)

PR #2624, opened 2023-10-11
ignore: Simplify the work-stealing strategy

tavianator:


BurntSushi:
Thanks! Out of curiosity, what prompted you to make this change?

tavianator:
No particular reason, I happened to be looking at the code again and realized that stealing from your left neighbour or your right neighbour shouldn't make a difference (and indeed perf is the same in my benchmarks)

BurntSushi (approved):

PR #2635, opened 2023-10-16
globset: Add GlobSet::builder

jplatte:
It's a somewhat common pattern to have `Type::builder` be an alias for `TypeBuilder::new` (or even the only way of constructing `TypeBuilder`). It means you don't have to import the builder type separately for common usage.

Thanks for this very handy small library 💚

### DIFF
--- a/crates/globset/src/lib.rs
+++ b/crates/globset/src/lib.rs
@@ -304,6 +304,14 @@ pub struct GlobSet {
 }
 
 impl GlobSet {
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
Thanks! I've taken to doing this in other crates as well. I like it.

jplatte:
I'm not in a hurry to have this merged, but is it normal that it takes many weeks from creation of rollup to merge?

PR #2642, opened 2023-10-30
ignore: Avoid contention on num_pending

tavianator:
Previously, every worker would increment the shared num_pending count on
every new work item, and decrement it after finishing them, leading to
lots of contention.  Now, we only track the number of workers actively
running, so there is no contention except when workers go to sleep or
wake up.

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -1681,20 +1681,20 @@ impl<'s> Worker<'s> {
                     return None;
                 }
                 None => {
-                    let active = self
-                        .active_workers
-                        .fetch_sub(1, AtomicOrdering::Relaxed);
-                    if active == 1 {
-                        // Every other thread is blocked at the next recv().
-                        // Send the initial quit message and quit.
+                    if self.deactivate_worker() == 0 {
+                        // If deactivate_worker() returns 0, every worker thread
+                        // is currently within the critical section between the
+                        // acquire in deactivate_worker() and the release in
+                        // activate_worker() below.  For this to happen, every
+                        // worker's local deque must be simultaneously empty,
+                        // meaning there is no more work left at all.
                         self.send_quit();
                         return None;
                     }
                     // Wait for next `Work` or `Quit` message.
                     loop {
                         if let Some(v) = self.recv() {
-                            self.active_workers
-                                .fetch_add(1, AtomicOrdering::Relaxed);
+                            self.activate_worker();
                             value = Some(v);
                             break;
                         }
@@ -1733,6 +1733,16 @@ impl<'s> Worker<'s> {
     /// Receive work.
     fn recv(&self) -> Option<Message> {
         self.stack.pop()
+    }
+
+    /// Deactivates a worker and returns the number of currently active workers.
+    fn deactivate_worker(&self) -> usize {
+        self.active_workers.fetch_sub(1, AtomicOrdering::Acquire) - 1
+    }
+
+    /// Reactivates a worker.
+    fn activate_worker(&self) {
+        self.active_workers.fetch_add(1, AtomicOrdering::Release);
     }
 }
 
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -1279,7 +1279,7 @@ impl WalkParallel {
         }
         // Create the workers and then wait for them to finish.
         let quit_now = Arc::new(AtomicBool::new(false));
-        let num_pending = Arc::new(AtomicUsize::new(stack.len()));
+        let active_workers = Arc::new(AtomicUsize::new(threads));
         let stacks = Stack::new_for_each_thread(threads, stack);
         std::thread::scope(|s| {
             let handles: Vec<_> = stacks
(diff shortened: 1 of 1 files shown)


tavianator:
Quick benchmark results:

```console
tavianator@tachyon $ hyperfine -w2 fd-{before,after}" -j24 -u '^$' ~/code/bfs/bench/corpus/chromium"
Benchmark 1: fd-before -j24 -u '^$' ~/code/bfs/bench/corpus/chromium
  Time (mean ± σ):     293.9 ms ±   4.1 ms    [User: 3624.1 ms, System: 3138.2 ms]
  Range (min … max):   288.2 ms … 298.8 ms    10 runs
 
Benchmark 2: fd-after -j24 -u '^$' ~/code/bfs/bench/corpus/chromium
  Time (mean ± σ):     227.3 ms ±   2.0 ms    [User: 1576.7 ms, System: 3562.5 ms]
  Range (min … max):   224.2 ms … 230.6 ms    13 runs
 
Summary
  fd-after -j24 -u '^$' ~/code/bfs/bench/corpus/chromium ran
    1.29 ± 0.02 times faster than fd-before -j24 -u '^$' ~/code/bfs/bench/corpus/chromium
tavianator@tachyon $ hyperfine -w2 fd-{before,after}" -j48 -u '^$' ~/code/bfs/bench/corpus/chromium"
Benchmark 1: fd-before -j48 -u '^$' ~/code/bfs/bench/corpus/chromium
  Time (mean ± σ):     210.6 ms ±   2.5 ms    [User: 4673.9 ms, System: 4393.0 ms]
  Range (min … max):   205.6 ms … 216.5 ms    13 runs
 
Benchmark 2: fd-after -j48 -u '^$' ~/code/bfs/bench/corpus/chromium
  Time (mean ± σ):     190.7 ms ±   5.1 ms    [User: 2173.8 ms, System: 5836.8 ms]
  Range (min … max):   185.1 ms … 201.6 ms    15 runs
 
Summary
  fd-after -j48 -u '^$' ~/code/bfs/bench/corpus/chromium ran
    1.10 ± 0.03 times faster than fd-before -j48 -u '^$' ~/code/bfs/bench/corpus/chromium
```

crates/ignore/src/walk.rs
```diff
@@ -1682,15 +1681,10 @@ impl<'s> Worker<'s> {
...
-                    // work. We have this guarantee because num_pending is
-                    // always incremented before each job is submitted and only
-                    // decremented once each job is completely finished.
-                    // Therefore, if this reaches zero, then there can be no
-                    // other job running.
-                    if self.num_pending() == 0 {
+                    let active = self
+                        .active_workers
+                        .fetch_sub(1, AtomicOrdering::Relaxed);
+                    if active == 1 {
```
BurntSushi, line 1687: Can you add a comment explaining the termination logic here?

crates/ignore/src/walk.rs
```diff
@@ -1699,6 +1693,8 @@ impl<'s> Worker<'s> {
                     // Wait for next `Work` or `Quit` message.
                     loop {
                         if let Some(v) = self.recv() {
+                            self.active_workers
+                                .fetch_add(1, AtomicOrdering::Relaxed);
```
BurntSushi, line 1697: So I did like how previously there were methods with decentish names that made call sites a bit more descriptive. Maybe something like this:

```
// Inactivates a single worker and
// returns the total number of remaining
// active workers.
fn worker_inactivate(&mut self) -> bool { ... }

// Activates a single inactive worker.
fn worker_activate(&mut self) { ... }
```

BurntSushi (changes requested):
Thanks! I'll need to think on this one. I've had problems with termination in the past and this changes the logic around that.

crates/ignore/src/walk.rs
tavianator: I updated the comment with a correctness argument in d6f850f57436301fea602a5b4419ffacb464bcc9

crates/ignore/src/walk.rs
tavianator: Good idea, done in d6f850f57436301fea602a5b4419ffacb464bcc9

tavianator:
> Thanks! I'll need to think on this one. I've had problems with termination in the past and this changes the logic around that.

Understandable!  I am reasonably sure of the correctness now that I've strengthened the memory orderings to acquire/release.  But it's not a formal proof, and I haven't tried too hard to break it either.

BurntSushi:
Oh, also, what about the relaxed loads? It looks like they should be fine to me here, but I wonder if Acquire/Release are necessary. I think that's probably why I used `SeqCst` originally (because I didn't think through what was actually required).

tavianator:
> Oh, also, what about the relaxed loads? It looks like they should be fine to me here, but I wonder if Acquire/Release are necessary. I think that's probably why I used `SeqCst` originally (because I didn't think through what was actually required).

I *think* it was still correct with relaxed accesses due to other orderings imposed by the deque implementation, but it's easier to argue correctness with acquire/release and they're not performance-critical any more so I switched them.

tmccombs (approved):

BurntSushi (approved):
OK, I've tried this out:

1. I think I buy that this is a correct change.
2. I have actually tried it on a number of corpora and everything looks OK.
3. I haven't been able to observe any reliable performance difference. Nevertheless, I buy the argument here.

Thank you!!!

PR #2657, opened 2023-11-27
doc: fix typo, Update GUIDE.md

sorrycc:


BurntSushi:
The way it's written is correct. Both `-trust` and` -t rust` are accepted and are equivalent.

PR #2669, opened 2023-11-30
Fix Guix install instructions

TimoWilken:
`guix install` should not be run using `sudo`, as per <https://packages.guix.gnu.org/packages/ripgrep/>.

Guix is more like Nix in that users install packages without root privileges.


BurntSushi (approved):
Thanks!

TimoWilken:
Wow that was quick, thank you @BurntSushi!

Issue #1352, opened 2019-08-17
How to Show Excerpt Near the Match?


johnyradio:
# version
% rg --version
ripgrep 11.0.2
-SIMD -AVX (compiled)
+SIMD -AVX (runtime)

# install ripgrep?
pacman -S ripgrep

# operating system
% hostnamectl 
  Operating System: Arch Linux
            Kernel: Linux 5.2.8-arch1-1-ARCH
      Architecture: x86-64

# question

On very long lines, how can i return a limited number of characters before and after the match?
--max-columns-preview seems to preview the _beginning_ of the line, which may not contain the match, so that's less useful. 

thx!

BurntSushi:
> On very long lines, how can i return a limited number of characters before and after the match?

You can't.

> Maybe that's a feature request?

Could you please provide a more complex specification for how this feature is supposed to work? If it helps, write it as if you were adding documentation for the new feature to the man page.

johnyradio:
Use -o with --max-columns-preview to display the matching text in context. 

If the original text is:

Every year since 1938, the organization presents Stern Grove Festival, a free concert series, in Sigmund Stern Grove, a beautiful outdoor amphitheater located at 19th Avenue and Sloat Boulevard in San Francisco.


Then this
rg  -o -M 10 --max-columns-preview Grove

will return
und Stern Grove, a beauti

There might be more correct options than how i've done , but this shows the basic idea.

BurntSushi:
Thanks, but that's not good enough. `-o -M10 --max-columns-preview` already has a defined behavior.

I don't see a way to do this without an additional flag.

borekb:
I'd also love this feature (minified JS in my case...), it's probably true that it cannot be done without an additional flag. I imagine having something like this in my `.ripgreprc`:

```
--max-columns=150
--max-columns-preview
--max-columns-preview-type=contextual
```

This hypothetical `--max-columns-preview-type` option would support two values:

- `beginning-of-line` (current behavior)
- `contextual`

rod-stuchi:
This would be a really wanted feature, like borekb said to look in minified JS files.

I think this is like Context, flags `-A, -B and -C` but instead of lines, will be applied in columns.

New flags (like context in lines) should be the way to go, imho.

BurntSushi:
Plus-1 comments aren't going to move this forward. A specification, as requested above, would be more helpful.

rod-stuchi:
[...]

PR #2683, opened 2023-12-11
printer: implement --max-columns-preview-before

kkocdko:
See #1352 .


kkocdko:
Currently:

```sh
rg --max-columns 30 --max-columns-preview --max-columns-preview-before 10 --per-match '=require\('
```

screenshot_0

crates/core/flags/defs.rs
```diff
@@ -3768,6 +3770,53 @@ fn test_max_columns_preview() {
     assert_eq!(false, args.max_columns_preview);
 }
 
+/// --max-columns-preview-before
```
kkocdko, line 3773: Maybe a better name?

crates/core/flags/hiargs.rs
```diff
@@ -65,6 +65,7 @@ pub(crate) struct HiArgs {
     line_number: bool,
     max_columns: Option<u64>,
     max_columns_preview: bool,
+    max_columns_preview_before: Option<u64>,
```
kkocdko, line 68: We can combine this to `max_columns_preview: Option<u64>`, but will cause breaking changes. Is this acceptable?

crates/core/flags/defs.rs
```diff
@@ -5335,6 +5384,42 @@ fn test_pcre2_version() {
     assert_eq!(Some(SpecialMode::VersionPCRE2), args.special);
 }
 
+/// --per-match
```
kkocdko, line 5387: Exporse this to cli.

crates/printer/src/standard.rs
```diff
@@ -1360,6 +1375,13 @@ impl<'a, M: Matcher, W: WriteColor> StandardImpl<'a, M, W> {
     ) -> io::Result<()> {
         if self.config().max_columns_preview {
             let original = line;
+            if let Some(p) = self.config().max_columns_preview_before {
+                let start = matches[0].start().saturating_sub(p as usize);
+                if start != 0 {
+                    self.write(b"[... omitted start of long line] ")?;
+                    line = line.with_start(start);
+                }
+            }
```
kkocdko, line 1384: Compare to the full `--column-context-before` + `--column-context-after` support that needs to modify everywhere, the implementation code here is quiet simple.

kkocdko:
Hi, @BurntSushi , what is your opinion about this design? Many people work on JavaScript bundler and web reverse engineering rely on this feature because we needs to deal with very long line and many huge files.

It's fine if you don't like this. I will still maintenance this patch for my friends.

BurntSushi:
Dunno. I haven't reviewed this yet. I find the flag name long and overall unwieldy to use IMO. And I'm immediately skeptical of exposing `per-match` due to the problems with `--vimgrep`. (The size of the output is worst case quadratic in the size of the input.)

I don't have any timeline on when I'll review this, sorry.

To be clear, I'm open to adding some kind of feature for this functionality.

BurntSushi:
I think I'm going to pass on this patch. I don't think this is the right way to approach this problem. I think that if we're going to have this functionality, we should do it better than forcing each match to print on its own line. It's a clever hack to keep the implementation simple, I'll give you that, but this isn't something I want to commit to long term.

PR #2684, opened 2023-12-11
complete/fish: Make fish completion more intelligent

blyxxyz:
- Stop using `-n __fish_use_subcommand`. This had the effect of ignoring options if a positional argument has already been given, but that's not how ripgrep works.

- Only suggest negation options if the option they're negating is passed (e.g., only complete `--no-pcre2` if `--pcre2` is present). The zsh completions already do this.

- Take into account whether an option takes an argument. If an option is not a switch then it won't suggest further options until the argument is given, e.g. `-C<tab>` won't suggest options but `-i<tab>` will.

- Suggest correct arguments for options. We already completed a fixed set of choices where available, but now we go further:

  - Filenames are only suggested for options that take filenames.

  - `--pre` and `--hostname-bin` suggest binaries from `$PATH`.

  - `-t`/`--type`/&c use `--type-list` for suggestions, like in zsh, with a preview of the glob patterns:
  image

  - `--encoding` uses a hardcoded list extracted from the zsh completions. This has been refactored into a separate file, and the range globs (`{1..5}`) replaced by comma globs (`{1,2,3,4,5}`) since those work in both shells. I verified that this produces the same list as before in zsh, and the same list in fish (albeit in a different order).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+14.1.0 (TBD)
+============
+This is a minor release with a few small new features and bug fixes.
+
+Feature enhancements:
+
+* [FEATURE #2684](https://github.com/BurntSushi/ripgrep/issues/2684):
+  Improve completions for the `fish` shell.
+
+
 14.0.3 (2023-11-28)
 ===================
 This is a patch release with a bug fix for the `--sortr` flag.
--- a/crates/core/flags/complete/encodings.sh
+++ b/crates/core/flags/complete/encodings.sh
@@ -1,0 +1,29 @@
+# This is impossible to read, but these encodings rarely if ever change, so
+# it probably does not matter. They are derived from the list given here:
+# https://encoding.spec.whatwg.org/#concept-encoding-get
+#
+# The globbing here works in both fish and zsh (though they expand it in
+# different orders). It may work in other shells too.
+
+{{,us-}ascii,arabic,chinese,cyrillic,greek{,8},hebrew,korean}
+logical visual mac {,cs}macintosh x-mac-{cyrillic,roman,ukrainian}
+866 ibm{819,866} csibm866
+big5{,-hkscs} {cn-,cs}big5 x-x-big5
+cp{819,866,125{0,1,2,3,4,5,6,7,8}} x-cp125{0,1,2,3,4,5,6,7,8}
+csiso2022{jp,kr} csiso8859{6,8}{e,i}
+csisolatin{1,2,3,4,5,6,9} csisolatin{arabic,cyrillic,greek,hebrew}
+ecma-{114,118} asmo-708 elot_928 sun_eu_greek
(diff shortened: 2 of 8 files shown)


BurntSushi:
Nice thank you! I made one small change to ensure that `include_str!("encodings.sh")` is only included once. I'm not sure if the compiler would automatically de-duplicate them, so I'd just rather do it myself. I also wrapped stuff to 80 columns.

I don't personally test these sorts of things (well, I do test zsh because I use it), so I'm just going to merge it as-is and we can always improve it as feedback comes in.

cstyles:
> Only suggest negation options if the option they're negating is passed (e.g., only complete `--no-pcre2` if `--pcre2` is present). The zsh completions already do this.

I think the intent here is good but unfortunately completions aren't "config-file-aware" which is kinda annoying. For example, I have `--max-columns-preview` in my `.ripgreprc` file and if I want to disable it for a particular invocation of ripgrep, now I have to type `--no-max-columns-preview` out each time.

blyxxyz:
The zsh completions have the same behavior, so I figured it'd be OK, but maybe I should have given it more thought.

I think it's fixable by reimplementing `__fish_contains_opt`. I'll look into that.

BurntSushi:
@okdana Do you have thoughts? Ideally we would keep the behavior roughly the same between zsh and fish if possible. (You might have five the zsh completions before config file support landed?)

BurntSushi:
@cstyles In the future, concerns like this should be in new issues. As it is, this concern will be very difficult to track.

blyxxyz:
Thanks, I see now. I conflated that conditional handling with this feature:
```zsh
  # ripgrep has many options which negate the effect of a more common one — for
  # example, `--no-column` to negate `--column`, and `--messages` to negate
  # `--no-messages`. There are so many of these, and they're so infrequently
  # used, that some users will probably find it irritating if they're completed
  # indiscriminately, so let's not do that unless either the current prefix
  # matches one of those negation options or the user has the `complete-all`
  # style set. Note that this prefix check has to be updated manually to account
  # for all of the potential negation options listed below!
  if
    # We also want to list all of these options during testing
    [[ $_RG_COMPLETE_LIST_ARGS == (1|t*|y*) ]] ||
    # (--[imnp]* => --ignore*, --messages, --no-*, --pcre2-unicode)
    [[ $PREFIX$SUFFIX == --[imnp]* ]] ||
    zstyle -t ":completion:${curcontext}:" complete-all
  then
    no=
  fi
```
zsh's behavior is indeed very different from what I implemented.

Fish completions commonly omit negation flags entirely. At least that's what the completions for curl and HTTPie do. Making them conditional seems better than that, and once it's config-aware I don't see significant downsides.

Hiding them unless the user has entered a prefix (like the above) is possible but definitely unconventional for fish completions.

BurntSushi:
Aye okay, I agree that it makes sense to do the conventional thing for each shell. That makes the most sense to me in terms of what users are going to expect.

I defer to Fish users to implement what is conventional though.

PR #2708, opened 2024-01-10
complete/fish: Take RIPGREP_CONFIG_PATH into account

blyxxyz:
The fish completions now also pay attention to the configuration file to determine whether to suggest negation options and not just to the current command line.

This doesn't cover all edge cases. For example the config file is cached, and so changes may not take effect until the next shell session. But the cases it doesn't cover are hopefully very rare.

See the post hoc discussion on https://github.com/BurntSushi/ripgrep/pull/2684.

cc @cstyles


BurntSushi (approved):
Thank you!

PR #2720, opened 2024-01-24
Add `Arbitrary` to `Glob`

johnsonw:
Derive `Arbitrary` for `Glob` struct. This feature is optional so the derive will only take place when the feature is enabled. This feature is mandatory when using Glob in fuzz testing.


crates/globset/Cargo.toml
```diff
@@ -21,6 +21,7 @@ bench = false
 
 [dependencies]
 aho-corasick = "1.1.1"
+arbitrary = { version = "1.3.2", optional = true, features = ["derive"] }
```
BurntSushi, line 24: I don't like implicitly depending on crate names for features. Could you explicitly add an `arbitrary` feature? Example: https://github.com/rust-lang/regex/blob/0c0990399270277832fbb5b91a1fa118e6f63dba/regex-syntax/Cargo.toml#L18

BurntSushi: Please also add docs for this feature in the crate docs, e.g., https://docs.rs/regex-syntax/latest/regex_syntax/#crate-features

BurntSushi (changes requested):
Overall LGTM, with a couple changes.

Note that I am working on a rewrite of `globset` and it's not obvious to me that maintaining this `Arbitrary` impl will be straight-forward. So this might get lost in a future upgrade that will probably happen this year.

(I also resisted adding this to `regex-syntax` pretty hard, but it turned out that this was the least bad option after a lot of discussion.)

crates/globset/Cargo.toml
johnsonw: Would you be ok with me adding a `fuzz` folder with a simple fuzz test?

BurntSushi: Probably fine. But it needs to be tested in CI.

johnsonw: Thanks @BurntSushi. I've setup a `fuzz` directory and added a `fuzz_testing` action to ci.yml. Let me know if you want to adjust the duration for the fuzz testing as this can be modified. I've also addressed the other comments above.

crates/globset/src/lib.rs
```diff
@@ -16,6 +16,19 @@ globs as matching.
 
 This example shows how to match a single glob against a single file path.
 
+
+# Crate features
+
+This crate includes optional features that can be enabled if neceesary. These features are
```
jgrund, line 22:
```suggestion
This crate includes optional features that can be enabled if necessary. These features are
```

johnsonw:
Happy Friday @BurntSushi, I just wanted to check and see if there is anything else you would like me to address?

.github/workflows/ci.yml
```diff
@@ -215,3 +215,22 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install packages (Ubuntu)
+        run: |
+          sudo apt-get update
+          sudo apt-get install g++ --yes
+
+      - uses: dtolnay/rust-toolchain@nightly
```
BurntSushi, line 230: Is nightly needed here? If not, please use `@master` like the rest of the jobs.

.github/workflows/ci.yml
```diff
@@ -215,3 +215,22 @@ jobs:
...
+
+  fuzz_testing:
+    name: Run Fuzz Tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install packages (Ubuntu)
+        run: |
+          sudo apt-get update
```
BurntSushi, line 227: Is this necessary?

crates/globset/src/lib.rs
```diff
@@ -16,6 +16,19 @@ globs as matching.
 
 This example shows how to match a single glob against a single file path.
 
+
+# Crate features
+
+This crate includes optional features that can be enabled if necessary. These features are
```
BurntSushi, line 22: Lines should be wrapped to 79 columns (inclusive).

crates/globset/src/lib.rs
```diff
@@ -16,6 +16,19 @@ globs as matching.
 
 This example shows how to match a single glob against a single file path.
 
+
+# Crate features
```
BurntSushi, line 20: Why is this added in the middle of an example? It should go at the bottom of the crate docs.

fuzz/.cargo/config.toml
```diff
@@ -0,0 +1,4 @@
+[alias]
+fuzz_install = "install cargo-fuzz"
+fuzz_list = "fuzz list"
+fuzz_run = "fuzz run"
```
BurntSushi, line 4: Please include newlines at the end of files.

fuzz/.cargo/config.toml
```diff
@@ -0,0 +1,4 @@
+[alias]
```
BurntSushi, line 1: Why does this file exist?

fuzz/README.md
```diff
@@ -0,0 +1,39 @@
+# Fuzz Testing
+
+## Introduction
+
+Fuzz testing produces pseudo-random / arbitrary data that is used to find stability issues within a code base. While Rust provides a strong type system, this does not guarantee that an object will convert properly from one struct to another. It is the responsibility of the developer to ensure that a struct is converted properly. Fuzz testing will generate input within the domain of each property. This arbitrary data can then be used to convert from ObjectA to ObjectB and then back. This type of testing will help catch bugs that the type system is not able to see.
```
BurntSushi, line 5: This is one long line. It should be wrapped to 79 columns (inclusive).

BurntSushi: And the same for lines below.

fuzz/.cargo/config.toml
BurntSushi: I don't think this file buys enough to warrant its existence. Having aliases for `cargo install cargo-fuzz` doesn't seem useful to me. And aliasing `cargo fuzz_list` to `cargo fuzz list` seems a little weird? Like, why?

fuzz/fuzz_targets/fuzz_glob.rs
```diff
@@ -0,0 +1,25 @@
+#![no_main]
+
+extern crate globset;
+extern crate libfuzzer_sys;
```
BurntSushi, line 4: I don't believe you need these two lines.

fuzz/fuzz_targets/fuzz_glob.rs
```diff
@@ -0,0 +1,25 @@
...
+
+use globset::Glob;
+use libfuzzer_sys::fuzz_target;
+use std::str::FromStr;
```
BurntSushi, line 8: This should be written like this:

```rust
use std::str::FromStr;

use {globset::Glob, libfuzzer_sys::fuzz_target};
```

That is, imports should be grouped by: 1) things from std, 2) things from third party dependencies and 3) (none in this case) crate-internal imports.

fuzz/fuzz_targets/fuzz_glob.rs
```diff
@@ -0,0 +1,25 @@
...
+
+fuzz_target!(|glob_str: &str| {
```
BurntSushi, line 10: I'd just write this as `libfuzzer_sys::fuzz_target!` and dispense with the import above.

fuzz/rust-toolchain
```diff
@@ -0,0 +1,3 @@
+[toolchain]
+channel = "nightly"
+profile = "default"
```
BurntSushi, line 3: Why is this using nightly?

BurntSushi: Ideally this file wouldn't exist at all.

fuzz/scripts/run-all.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -ex
+
+# Number of seconds to run the script. Default to 5 minutes.
+duration="${1:-300}"
+start_time=$SECONDS
+elapsed_time=0
+
+while (( $elapsed_time < $duration )); do
+    for target in $(cargo fuzz_list); do cargo fuzz_run $target --release -- -max_total_time=60 -verbosity=0; done
```
BurntSushi, line 9: Why is this smushed on to one line? Vertical space is cheap. Use it. :)

fuzz/scripts/run-all.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -ex
```
BurntSushi, line 1: The name of this script should have `fuzz` in it somewhere.

fuzz/scripts/run-all.sh
```diff
@@ -0,0 +1,12 @@
...
+
+# Number of seconds to run the script. Default to 5 minutes.
+duration="${1:-300}"
+start_time=$SECONDS
+elapsed_time=0
```
BurntSushi, line 6: This should output some kind of message to stderr saying that it's going to run for `n` seconds.

fuzz/scripts/run-all.sh
```diff
@@ -0,0 +1,12 @@
...
+
+while (( $elapsed_time < $duration )); do
+    for target in $(cargo fuzz_list); do cargo fuzz_run $target --release -- -max_total_time=60 -verbosity=0; done
+    
+    elapsed_time=$(($SECONDS - $start_time))
+done
```
BurntSushi, line 12: I think overall, I would get rid of this script. It seems a little weird to be honest. I can't imagine it being used for anything really.

.github/workflows/ci.yml
```diff
@@ -215,3 +215,22 @@ jobs:
...
+          sudo apt-get install g++ --yes
+
+      - uses: dtolnay/rust-toolchain@nightly
+
+      - run: cargo fuzz_install
+        working-directory: fuzz
+
+      - run: ./scripts/run-all.sh
+        working-directory: fuzz
```
BurntSushi, line 236: For CI, I don't think we need to actually run fuzzing. Let's just make sure all of the fuzz targets build.

BurntSushi (changes requested):
Thanks. I normally take a long time to review things. So I wouldn't necessarily expect a quick turn-around here. With that said, I left a number of comments. Thank you for working on this.

johnsonw:
Good afternoon, I just wanted to follow up on this. Let me know if there is anything else I need to address.

PR #2725, opened 2024-01-31
feat: add vuejs as a default type

DMunkei:
Just adding `vue` as a valid flag to pass to `-t`.

Using `rg -tvue <pattern>` now works.


BurntSushi:
Closed by #2772

PR #2726, opened 2024-01-31
Detect WORKSPACE.bzlmod for bazel file type

keith:
This file came alongside MODULE.bazel and I should have added it here previously


keith:
thanks!

PR #2727, opened 2024-02-01
Stop all threads and write ANSI-Reset on Ctrl-C

th1000s:
On Ctrt-C stop all threads currently writing to the terminal, ignoring whatever they have locked, and write the ANSI-Reset code directly to the terminal, then exit.

This uses `pthread_kill()` or `SuspendThread()`, then `libc::write()` or `WriteConsoleA()` on Unix or Windows, respectively.

-------

So far only tested and enabled on Linux, macOS and Windows (cmd.exe, Powershell, Mingw64), but I am sure this will
work on the BSDs and several others which have `pthread_kill()` - but I'd rather have someone confirm this first.

And to compare the before/after, running `rg` with a `-j`  of 7, 11, 22, 33.. etc. will disable this handler.


th1000s:
When suggesting to someone to use `rg` instead of "legacy" `grep -R .. | grep -v -e .git -e build` or `git grep ..`
on the _very first try_ (maybe spooked by the speed of the output) after ^C the terminal remained blood red and I had a slightly embarrassed look on my face.

A lot of users have shells or prompts which handle this automatically. But now that `rg` is packaged by most Linux distros, more users which don't customize their shells use it, and first impressions matter as to not end up as "newfangled nonsense" :)

BurntSushi:
ripgrep has been in Debian for years.

I'm not sure if I'm going to take this. It is so much code to solve an issue that I don't think is a big deal.

BurntSushi:
I'm going to close this. I don't think my opinion has changed since the above. I've tried this in the past, and it just lead to worse problems. 500 lines of code to deal with this just isn't worth it IMO.

PR #2760, opened 2024-03-19
doc: playground (#2758)

nalgeon:
Playground mentioned in the readme as discussed in #2758.

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,9 +222,12 @@ multiline search and opt-in fancy regex support via PCRE2.
 
 ### Playground
 
-If you'd like to try ripgrep before installing, there's a [playground](https://codapi.org/ripgrep/) and an [interactive tutorial](https://codapi.org/try/ripgrep/).
+If you'd like to try ripgrep before installing, there's an unofficial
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -220,6 +220,13 @@ configuration files, passthru, support for searching compressed files,
 multiline search and opt-in fancy regex support via PCRE2.
 
 
+### Playground
+
+If you'd like to try ripgrep before installing, there's a [playground](https://codapi.org/ripgrep/) and an [interactive tutorial](https://codapi.org/try/ripgrep/).
+
+If you have any questions about these, please open an issue in the [tutorial repo](https://github.com/nalgeon/tryxinyminutes).
```
BurntSushi, line 227: Could you add the word "unofficial" somewhere in this text. Maybe, "there's an unofficial playground." Just so folks are clear that it isn't maintained as part of the ripgrep project.

Also, could you follow the stylistic convention of the surrounding text and wrap lines to 79 columns (inclusive)?

BurntSushi (changes requested):

README.md
nalgeon: Sure!

Issue #2767, opened 2024-03-27
ignore: WalkBuilder::build_parallel() example does not compile


cgzones:
The example given in the documentation for `ignore::WalkBuildeer::build_parallel()`
https://github.com/BurntSushi/ripgrep/blob/eca13f08a2a0ece89a423565891e48a90a8aedef/crates/ignore/src/walk.rs#L594
does not compile, since the returned closure is expected to return a `WalkState`, not `()`.

BurntSushi:
PRs are welcome to fix this sort of thing.

PR #2768, opened 2024-03-27
ignore/walk: correct build_parallel() documentation

cgzones:
The returned closure should return `WalkState`, not `()`.

Closes: #2767

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -591,7 +591,7 @@ impl WalkBuilder {
     ///
     /// Note that this *doesn't* return something that implements `Iterator`.
     /// Instead, the returned value must be run with a closure. e.g.,
-    /// `builder.build_parallel().run(|| |path| println!("{:?}", path))`.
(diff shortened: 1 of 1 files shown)


crates/ignore/src/walk.rs
```diff
@@ -591,7 +591,8 @@ impl WalkBuilder {
     ///
     /// Note that this *doesn't* return something that implements `Iterator`.
     /// Instead, the returned value must be run with a closure. e.g.,
-    /// `builder.build_parallel().run(|| |path| println!("{:?}", path))`.
+    /// `builder.build_parallel().run(||
+    ///     |path| { println!("{:?}", path); WalkState::Continue })`.
```
BurntSushi, line 595: Can you use `println!("{path:?}")` to shorten things? And put it on one line? Even if it exceeds my typical line length limit, I'd rather that because I've found newlines inside of inline code fences to sometimes produce weird rendering behavior.

BurntSushi (changes requested):

crates/ignore/src/walk.rs
cgzones: Applied

BurntSushi (approved):

PR #2774, opened 2024-04-02
ignore/types: add WGSL to the default file types

rhysd:
[WGSL][1] is a shading language for WebGPU. As defined in [Appendix A][2], the file extension is `.wgsl`.

This PR adds a file type support for WGSL to ripgrep.

[1]: https://www.w3.org/TR/WGSL/
[2]: https://www.w3.org/TR/WGSL/#text-wgsl-media-type


BurntSushi (approved):
Thanks!

PR #2784, opened 2024-04-18
strip dollar signs from the commands readme

brayo-pip:
People tend to click on the copy to clipboard button then paste it. But it has a leading dollar sign. The command fails. It takes almost no time to remove the dollar sign but it's a bit annoying. Consider merging


BurntSushi:
Duplicate of https://github.com/BurntSushi/ripgrep/pull/2599

PR #2787, opened 2024-04-23
Set up ripgrep for compilation on non-unix, non-windows platforms

holzschu:
The code of ripgrep compiles on almost any kind of architecture, including WebAssembly, except in two tiny places related to hyperlinks: 
- when it searches for the local hostname in `hostname.rs`, 
- when it does path canonalization in `hyperlink.rs`.

In both cases, there is a path with `#[cfg(unix)]` and a path with `#[cfg(windows)]`. WebAssembly being neither unix nor windows, compilation fails. This PR adds a third path:
- for localhost, it returns "localhost" instead of calling `gethostname()`, for all environments that are neither unix nor windows.
- for path canonalization, it uses `std::os::wasi::ffi::OsStrExt` instead of `std::os::unix::ffi::OsStrExt`, which is not available; this part is specific for wasm32-wasi. It could be possible to deactivate path canonalization entirely for architectures that are neither unix nor windows, but I aimed for the smallest working change.

With these two changes, it's possible to compile ripgrep with `cargo build --target wasm32-wasi` (and, more importantly, it works in wasm32-wasi environments).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,6 +189,21 @@ jobs:
       shell: bash
       run: ${{ env.CARGO }} test --bin rg ${{ env.TARGET_FLAGS }} flags::defs::tests::available_shorts -- --nocapture
 
+     # Setup and compile on the wasm32-wasi target
+  wasm:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Install Rust
+      uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: stable
+    - name: Add wasm32-wasi target
+      run: rustup target add wasm32-wasi
+    - name: Basic build
+      run: cargo build --verbose
+
   rustfmt:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,9 +189,12 @@ jobs:
       shell: bash
       run: ${{ env.CARGO }} test --bin rg ${{ env.TARGET_FLAGS }} flags::defs::tests::available_shorts -- --nocapture
 
-     # Setup and compile on the wasm32-wasi target
+  # Setup and run tests on the wasm32-wasi target via wasmtime.
   wasm:
     runs-on: ubuntu-latest
+    env:
+      # The version of wasmtime to download and install.
+      WASMTIME_VERSION: 15.0.1
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
@@ -201,9 +204,19 @@ jobs:
         toolchain: stable
     - name: Add wasm32-wasi target
       run: rustup target add wasm32-wasi
+    - name: Download and install Wasmtime
+      run: |
+        echo "CARGO_BUILD_TARGET=wasm32-wasi" >> $GITHUB_ENV
+        echo "RUSTFLAGS=-Ctarget-feature=+simd128" >> $GITHUB_ENV
+        curl -LO https://github.com/bytecodealliance/wasmtime/releases/download/v$WASMTIME_VERSION/wasmtime-v$WASMTIME_VERSION-x86_64-linux.tar.xz
+        tar xvf wasmtime-v$WASMTIME_VERSION-x86_64-linux.tar.xz
+        echo `pwd`/wasmtime-v$WASMTIME_VERSION-x86_64-linux >> $GITHUB_PATH
+        echo "CARGO_TARGET_WASM32_WASI_RUNNER=wasmtime run --wasm simd --" >> $GITHUB_ENV
     - name: Basic build
       run: cargo build --verbose
-
+    - name: Run tests
+      run: cargo test --verbose
+    
   rustfmt:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -189,12 +189,9 @@ jobs:
       shell: bash
       run: ${{ env.CARGO }} test --bin rg ${{ env.TARGET_FLAGS }} flags::defs::tests::available_shorts -- --nocapture
 
-  # Setup and run tests on the wasm32-wasi target via wasmtime.
+     # Setup and compile on the wasm32-wasi target
   wasm:
     runs-on: ubuntu-latest
-    env:
-      # The version of wasmtime to download and install.
-      WASMTIME_VERSION: 15.0.1
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


ltrzesniewski:
> It could be possible to deactivate path canonalization entirely for architectures that are neither unix nor windows

`std::path::absolute` [should be stabilized soon](https://github.com/rust-lang/rust/issues/92750), and I suppose it will replace canonicalization in ripgrep, which will both simplify and optimize this part of the code.

crates/printer/src/hyperlink.rs
```diff
@@ -853,6 +853,57 @@ impl HyperlinkPath {
         }
         HyperlinkPath(result)
     }
+    #[cfg(target_os = "wasi")]
+    pub(crate) fn from_path(original_path: &Path) -> Option<HyperlinkPath> {
```
blyxxyz, line 857: The [WASI spec](https://github.com/WebAssembly/wasi-filesystem/blob/174dbd7fa134672e594fd38dba9acfb53afaf312/imports.md#import-interface-wasifilesystemtypes020) seems to ban absolute paths and the [stdlib](https://github.com/rust-lang/rust/blob/c67277301c896857d0534f2bb7431680796833fb/library/std/src/sys/pal/wasi/fs.rs#L655) doesn't implement `canonicalize`. I don't know how black-and-white this is (especially in the long term) but right now this might as well be stubbed out with `None`.

When running on top of Windows I imagine you'd want to return a Windows path here, even though WASI is more Unix-like internally.

crates/cli/src/hostname.rs
```diff
@@ -25,10 +25,8 @@ pub fn hostname() -> io::Result<OsString> {
     }
     #[cfg(not(any(windows, unix)))]
     {
-        io::Error::new(
-            io::ErrorKind::Other,
-            "hostname could not be found on unsupported platform",
-        )
+        // backup solution for systems that do not have gethostname():
+        Ok(OsString::from("localhost"))
```
BurntSushi, line 29: I'm not a huge fan of giving wrong answers like this. And it seems like WASI can't really support hyperlinks anyway, due to the ban on absolute paths. So I think this change should probably be reverted.

crates/printer/src/hyperlink.rs
blyxxyz: AFAICT it works like this:
- WASI syscalls take a directory descriptor and a path relative to that directory descriptor.
- The WASI runtime gives the WASM binary a list of directory descriptors plus their actual original paths ([preopens](https://github.com/WebAssembly/wasi-filesystem/blob/e79b05803e9ffd3b0cfdc0a8af20ac743abbe36a/imports.md#import-interface-wasifilesystempreopens020)). So the runtime grants selective access. It could easily pretend the filesystem is smaller than it really is, chroot-style, but it doesn't have to: it can grant a `/home/user` descriptor with the `/home/user` path.
- wasi-libc takes absolute paths and [resolves](https://github.com/WebAssembly/wasi-libc/blob/9e8c542319242a5e536e14e6046de5968d298038/libc-bottom-half/sources/preopens.c#L182) them to a directory descriptor + relative path based on the longest matching prefix from the preopens, if any. (This is inspired by [libpreopen](https://github.com/musec/libpreopen).)
- wasi-libc also [emulates](https://github.com/WebAssembly/wasi-libc/blob/9e8c542319242a5e536e14e6046de5968d298038/libc-bottom-half/sources/getcwd.c) a working directory in WASM-space.

So it's not as bad as I thought. Thanks to the emulation absolute paths should for the most part just work as long as access to that part of the filesystem is granted, particularly on Unix. (Except for missing syscalls I guess.)

BurntSushi: Interesting. So `std::env::current_dir()` works, and thus `std::path::absolute` might in turn work here as well.

crates/cli/src/hostname.rs
holzschu: The problem I was trying to fix here, and that's also a question I have about setting up CI, is that the current version of the code causes a compilation error:    
```
Compiling grep-cli v0.1.10 (/Users/holzschu/src/Xcode_iPad/ripgrep_clone/crates/cli)
error[E0308]: mismatched types
  --> crates/cli/src/hostname.rs:28:9
   |
16 |   pub fn hostname() -> io::Result<OsString> {
   |                        -------------------- expected `Result<OsString, std::io::Error>` because of return type
...
28 | /         io::Error::new(
29 | |             io::ErrorKind::Other,
30 | |             "hostname could not be found on unsupported platform",
31 | |         )
   | |_________^ expected `Result<OsString, Error>`, found `Error`
   |
   = note: expected enum `Result<OsString, std::io::Error>`
            found struct `std::io::Error`
help: try wrapping the expression in `Err`
   |
28 ~         Err(io::Error::new(
29 |             io::ErrorKind::Other,
30 |             "hostname could not be found on unsupported platform",
31 ~         ))
   |

For more information about this error, try `rustc --explain E0308`.
```
So if the goal of setting up CI is to get the same result after the PR as before, that's not going to happen.

BurntSushi: Oh I see. I think you just want `Err(io::Error::new(...))`, as suggested in the error message you're showing. This particular code path is probably entirely untested in CI at present, and thus the fact that it is wrong now and has always been wrong was never noticed.

My main objection was to silently using a potentially incorrect value. I'm open to doing that in the future to get hyperlinks working in WASI after we've considered the possible alternatives. But since `std::fs::canonicalize` isn't going to work on WASI right now anyway, we should just avoid changing the semantic intent of this code.

holzschu: I've done the changes for `hostname.rs`. But then, for symmetry, I think I should have a branch  `#[cfg(not(any(windows, unix)))]`  instead of `#[cfg(target_os = "wasi")]` in `hyperlink.rs`. But what should `from_path()` return in that case?

BurntSushi: `None`

BurntSushi: With a DEBUG log message. Mentioned here: https://github.com/BurntSushi/ripgrep/pull/2787#discussion_r1576137687

holzschu:
I've changed the files and the title of the PR.
Now, the goal is for ripgrep to compile without error on platforms that are neither unix nor windows (that includes WebAssembly, but there might be other platforms as well.

It will emit an error in `hostname.rs` and a debug message in `hyperlinks.rs`.

.github/workflows/ci.yml
```diff
@@ -189,6 +189,21 @@ jobs:
...
+    - name: Checkout repository
+      uses: actions/checkout@v4
+    - name: Install Rust
+      uses: dtolnay/rust-toolchain@master
+      with:
+        toolchain: stable
+    - name: Add wasm32-wasi target
+      run: rustup target add wasm32-wasi
+    - name: Basic build
+      run: cargo build --verbose
```
BurntSushi, line 205: Did running tests via `wasmtime` fail? If so this is fine, but if they worked, then we should just include them here.

BurntSushi (changes requested):

.github/workflows/ci.yml
holzschu: Indeed, running tests via `wasmtime` does fail (but compilation works):
```
error[E0433]: failed to resolve: could not find `unix` in `os`
   --> tests/util.rs:197:22
    |
197 |         use std::os::unix::fs::symlink;
    |                      ^^^^ could not find `unix` in `os`
    |
```

crates/printer/src/hyperlink.rs
```diff
@@ -853,6 +853,13 @@ impl HyperlinkPath {
         }
         HyperlinkPath(result)
     }
+
+    // For other platforms (not windows, not unix), return None and log a debug message.
+    #[cfg(not(any(windows, unix)))]
+    pub(crate) fn from_path(original_path: &Path) -> Option<HyperlinkPath> {
+        log::debug!("Hyperlinks are not supported on this platform");
```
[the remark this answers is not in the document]
holzschu: This is done.

crates/printer/src/hyperlink.rs
```diff
@@ -853,6 +853,13 @@ impl HyperlinkPath {
         }
         HyperlinkPath(result)
     }
+
+    // For other platforms (not windows, not unix), return None and log a debug message.
```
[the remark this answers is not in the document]
holzschu: These changes have been incorporated.

.github/workflows/ci.yml
BurntSushi: Ah fair enough. We can save that for another day.

holzschu:
Thank *you* for your patience and pedagogy!

PR #2797, opened 2024-05-02
stats fixup

robert-bryson:
Tiny change, looks like the text for seconds spent processing for `--stats` was missing.


BurntSushi:
It's meant to be the total time. So I don't think I was missing anything here.

robert-bryson:
@BurntSushi , that makes more sense. Changed to total.

Issue #2795, opened 2024-05-01
The same `LineTerminator` is being used for searching and printing


ltrzesniewski:
### Please tick this box to confirm you have reviewed the above.

- [X] I have a different issue.

### What version of ripgrep are you using?

ripgrep 14.1.0

features:-simd-accel,+pcre2
simd(compile):+SSE2,-SSSE3,-AVX2
simd(runtime):+SSE2,+SSSE3,+AVX2

PCRE2 10.43 is available (JIT is available)

### How did you install ripgrep?

`cargo install ripgrep --features pcre2`

### What operating system are you using ripgrep on?

Ubuntu 22.04.4 LTS on WSL

### Describe your bug.

I noticed that ripgrep could insert CR characters into its output on Linux, which I didn't expect.

I took a quick look out of curiosity, and saw that the same `LineTerminator` is being used for searching *and* for printing, which I thought could be an oversight, so I figured I'd report it just in case. I don't think this has a real impact on anything though, so feel free to simply close this issue if that's ok for you.

I tracked down the culprit to this function: https://github.com/BurntSushi/ripgrep/blob/bb8601b2bafb5e68181cbbb84e6ffa4f7a72bf16/crates/printer/src/standard.rs#L1493-L1495

### What are the steps to reproduce the behavior?

```bash
$ echo hello | rg hello --no-config --count --no-crlf | xxd
00000000: 310a                                     1.

$ echo hello | rg hello --no-config --count --crlf | xxd
00000000: 310d 0a                                  1..
```

### What is the actual behavior?

The output changes between `--crlf` and `--no-crlf`.

Even on Linux, a CRLF end-of-line (`0d 0a`) is emitted when `--crlf` is used. I used `--count` to force `write_line_term()` to be called, but other flags such as `--pretty` work just as well.

### What is the expected behavior?

I expected the outputs of the two commands to be the same.

BurntSushi:
This was definitely intentional as written. The issue of line terminators is quite a tricky one. But there are cases where you have to _choose_ the line terminator. The summary printer is one such example. Another example is adding a line terminator at the end of output that doesn't otherwise contain a line terminator. Or when `-o` is used.

So how do you make the choice? There's no guarantee that the input uses line terminators consistently. So ripgrep has an internal notion of what the line terminator "ought" to be. When you use `--null-data`, that's `\x00`. When you use `--crlf`, that's `\r\n`.

[...]

PR #2805, opened 2024-05-12
Always use `\n` as line terminator in the summary printer

ltrzesniewski:
This contributes to #2795.

I took a closer look at the issue, and I noticed the following:

```bash
$ echo hello | rg hello --no-config --count --null-data | xxd
00000000: 3100                                     1.
```

i.e. the summary printer outputs null bytes instead of newlines when `--null-data` is used. I don't think that's intentional, and it's a noticeable behavior, so this PR changes this to always output `\n`.

I wasn't sure if I should use CRLF or LF on Windows, but since LF is used by default unless `--crlf` is enabled I used LF.

---

I also made changes to the standard printer (**not** included here), which I can submit in another PR if you think it's worth it. I just need to know if you'd rather output LF or CRLF on Windows in the standard printer. Using LF is a [simple change](https://github.com/ltrzesniewski/ripgrep/commit/cdfe0854ac21c22f8d86a383744bfec2a5aa4195#diff-9edc643fe15c729d9ad9528b23556005eb27db181e25f5f260962e5ed468b9dd), while using CRLF requires changing lots of unit tests so they pass on Windows.

### DIFF
--- a/crates/printer/src/summary.rs
+++ b/crates/printer/src/summary.rs
@@ -561,13 +561,13 @@ impl<'p, 's, M: Matcher, W: WriteColor> SummarySink<'p, 's, M, W> {
     /// write that path to the underlying writer followed by a line terminator.
     /// (If a path terminator is set, then that is used instead of the line
     /// terminator.)
-    fn write_path_line(&mut self, searcher: &Searcher) -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


BurntSushi:
See https://github.com/BurntSushi/ripgrep/issues/2795#issuecomment-3065702275

PR #2827, opened 2024-05-28
Add `*.tfvars` to list of Terraform filetypes

sorliem:
Hello,
I'd like to propose adding `*.tfvars` to the list of terraform filetypes. In setups with multiple workspaces, we have different configuration for say `staging.tfvars` and `production.tfvars`. This might override `terraform.tfvars`, so not sure if that should be removed.

Thank you 🙂


sorliem:
Actually, nevermind 😄 I discovered the arg: `--type-add='tf:*.tfvars'` which fixes my issue.

PR #2833, opened 2024-06-05
Optimize character escaping.

TDecking:
Rewrites the `char_to_escaped_literal` and `bytes_to_escaped_literal` functions in a way that minimizes heap allocations.
After this, the resulting string is the only allocation remaining.


BurntSushi (approved):
Thanks!

PR #2846, opened 2024-06-28
add Makefile.inc to "make" type

bad:
The *BSD build systems make use of "Makefile.inc" a lot.  Make the "make" type recognize this file by default.


bad:
Is there a reason this PR hasn't been merged yet?

BurntSushi:
Sometimes it just takes me a while.

Otherwise I think I agree with @okdana's feedback here.

bad:
Thanks for the quick response.

I'm fine with @okdana's suggestion and changed the code.

bad:
Is there anything I can do to get this merged?

PR #2875, opened 2024-08-21
chore: remove $ in code fenced in README.md

pwn-0x309:
We just copy and run


BurntSushi:
Duplicate of #2599 and #2784.

PR #2876, opened 2024-08-21
Fix README typo in ugrep command for searches without literal optimization opportunities

hjmeijer:
It seems the entry for ugrep in this table was copied-and-not-updated from the previous table. The timing measurements indicate that the correct command was measured, so I considered this the most likely command that was used.

I'm unable to be 100% sure obviously.


BurntSushi (approved):
Yeah, looks like a copy & paste error. I double checked the timings again and they are still correct for this particular benchmark.

PR #2880, opened 2024-09-03
fix the mistake of overriding mode

QWZeng:
a non-searching mode can override non-searching mode.
    It’s likely just a mistake or typo :
    if !matches!(*self, Mode::Search(_));
    it should be：
    if !matches!(new, Mode::Search(_))

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,15 +13,15 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.82"
+version = "1.0.86"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f538837af36e6f6a9be0faa67f9a314f8119e4e4b5867c6ab40ed60360142519"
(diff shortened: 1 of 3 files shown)


BurntSushi:
I think the docs are wrong here.

Also, why are you opening multiple PRs? I see this one and #2879.

QWZeng:
> I think the docs are wrong here.
> 
> Also, why are you opening multiple PRs? I see this one and #2879.



> I think the docs are wrong here.
> 
> Also, why are you opening multiple PRs? I see this one and #2879.

Excuse me, I'm not sure what did you mean "the docs wrong here",  the comments are wrong in code?
I've closed the #2879 PR, sorry for that.

BurntSushi:
From what I can tell, the code on master is correct, so I'm going to close this PR.

I may still be wrong. But you'll need to demonstrate that through actual examples. Also, this PR changes the results of existing tests, including things that were specifically commented on as intended behavior. So that would need to be justified as well.

Finally, please don't make spurious updates to dependencies. PRs should be as small as possible. There's no reason to include a `cargo update` in the patch.

Issue #2869, opened 2024-08-12
feat(globset): better way to see if ALL globs match


tmccombs:
A have a use case where I have a set of 1 or more globs, and want to know if all of them match, rather than if any of them match.

I can of course use a `Vec<Glob>` and do something like `globs.iter().all(|g| g.is_match(path))`, but it sounds like using a `GlobSet` may be more efficient, since it can use a single pass. 

I think I can accomplish this with `glob_set.matches(path).len() == glob_set.len()`, but then it has to allocate a `Vec`, and then I just throw it away, because I only care about the length. 

Possible ways this could be improved are:

1. Add a `matches_all` method that returns true if all globs match (fwiw, I'd also like that on `RegexSet`)
2. Return an iterator similar to how `RegexSet.matches` returns an iterator, so it can be counted instead of storing in a Vec.

PR #2900, opened 2024-09-19
globset: add matches_all method

tmccombs:
This returns true if all globs in the set match the supplied file.

Fixes: #2869

### DIFF
--- a/crates/globset/src/lib.rs
+++ b/crates/globset/src/lib.rs
@@ -351,6 +351,27 @@ impl GlobSet {
         false
     }
 
+    /// Returns true if ALL globs in this set match the path given.
+    pub fn matches_all<P: AsRef<Path>>(&self, path: P) -> bool {
+        self.matches_all_candidate(&Candidate::new(path.as_ref()))
+    }
+
+    /// Returns ture if all globs in this set match the path given.
+    ///
+    /// This takes a Candidate as input, which can be used to amortize the
+    /// cost of peparing a path for matching.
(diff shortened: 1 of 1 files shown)


tmccombs:
See also https://github.com/rust-lang/regex/pull/1228

crates/globset/src/lib.rs
```diff
@@ -351,6 +351,27 @@ impl GlobSet {
         false
     }
 
+    /// Returns true if ALL globs in this set match the path given.
```
BurntSushi, line 354: I don't think we need to use caps for `ALL` here. Just "all" is fine.

crates/globset/src/lib.rs
```diff
@@ -351,6 +351,27 @@ impl GlobSet {
...
+        self.matches_all_candidate(&Candidate::new(path.as_ref()))
+    }
+
+    /// Returns ture if all globs in this set match the path given.
+    ///
+    /// This takes a Candidate as input, which can be used to amortize the
+    /// cost of peparing a path for matching.
+    ///
+    /// This will return true if the set of globs is empty, as in that case all `0` of
+    /// the globs will match.
```
BurntSushi, line 365: Please wrap lines to 79 columns (inclusive).

BurntSushi: Also, this note should be added to `matches_all` too.

BurntSushi (changes requested):

PR #2909, opened 2024-10-03
Add `.svelte.ts` file extension to Svelte glob

### LINKED ISSUE #2874: Add .svelte file type to default type list

Svelte is popular enough with it's own file types .svelte and .svelte.ts (the latter would still be picked up by searching `-tts` but would probably make sense to add to the `.svelte` glob.

I would like to be able to simply run `rg "my query" -tsvelte` and not add the `--type-add` flag each time. I think this makes sense because it is now an established file type, and not my own made up filetype.

Maybe add to the existing list:

```
svelte: *.svelte, *.svelte.ts
```
or similar? :)

colin-heffernan:
Closes #2874.

While the `.ts` extension is for TypeScript, it is useful to see `.svelte.ts` files as part of the Svelte glob.


BurntSushi:
I am unclear on whether this is appropriate or not: https://github.com/BurntSushi/ripgrep/issues/2874#issuecomment-3122085648

PR #2918, opened 2024-10-23
Add --histogram=NUM argument

Cabbache:
With this output mode, `rg` will print a histogram of the number of matches within certain ranges of byte offsets. The width of the ranges is the bin size which is specified in the argument.


BurntSushi:
Thanks for working on this. It looks neat. But I think this is too niche of a use case to have in ripgrep proper. And it seems likely to be a feature that begets more requests in terms of customizing how it works. It seems like you should be able to calculate this in an external tool by reading ripgrep's JSON output.

Issue #2778, opened 2024-04-10
.gitignore rule is matched incorrectly while in a subdir


woess:
### Please tick this box to confirm you have reviewed the above.

- [X] I have a different issue.

### What version of ripgrep are you using?

ripgrep 14.1.0

### How did you install ripgrep?

cargo install

### What operating system are you using ripgrep on?

Linux

### Describe your bug.

.gitignore rule to ignore `/dir/*.ext` works correctly when running rg from the repo root, but incorrectly ignores `*.ext` in all subdirs when running from `dir`.

### What are the steps to reproduce the behavior? / What is the actual behavior?

```
mkdir /tmp/repro
cd /tmp/repro
git init
mkdir parent
mkdir parent/subdir
echo "/parent/*.txt" > .gitignore
echo "please ignore me" > parent/ignore-me.txt
echo "please don't ignore me" > parent/subdir/dont-ignore-me.txt
```

while in git repo root dir (/tmp/repro), everything is working as expected:
```
$ rg ignore
parent/subdir/dont-ignore-me.txt
1:please don't ignore me
```
but once you `cd` into `./parent`, suddenly the rule is unexpectedly applied to the file in `subdir`, too, and nothing is found:
```
$ cd parent
$ rg ignore
rg: No files were searched, which means ripgrep probably applied a filter you didn't expect.
Running with --debug will show why files are being skipped.
```

Debug output (argument parsing omitted: "no extra arguments found from configuration file", "heuristic chose to search ./"):
```
/tmp/repro $ rg ignore
rg: DEBUG|grep_regex::config|/…/grep-regex-0.1.12/src/config.rs:175: assembling HIR from 1 fixed string literals
rg: DEBUG|globset|/…/globset-0.4.14/src/lib.rs:453: built glob set; 0 literals, 0 basenames, 12 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 0 regexes
rg: DEBUG|globset|/…/globset-0.4.14/src/lib.rs:453: built glob set; 0 literals, 0 basenames, 0 extensions, 0 prefixes, 0 suffixes, 1 required extensions, 0 regexes
rg: DEBUG|ignore::walk|/…/ignore-0.4.22/src/walk.rs:1799: ignoring ./.git: Ignore(IgnoreMatch(Hidden))
[...]

PR #2933, opened 2024-11-15
ignore: fix filtering searching subdir or .ignore in parent dir

### LINKED ISSUE #2836: Some directories in .ignore aren't ignored

- [X] I have a different issue.

ripgrep 14.1.0

pacman

Arch Linux

### Describe your bug.

Some directories in ~/.ignore aren't ignored correctly.

Run this script `~/test` under `~/testdir`:

```
#!/bin/bash

mkdir -p ~/testdir/sub/sub2
cd ~/testdir
chmod -r sub/sub2
echo 'sub/sub2' >> ~/.ignore
echo 'sub/sub2/' >> ~/.ignore
echo '/testdir/sub/sub2' >> ~/.ignore
echo '/testdir/sub/sub2/' >> ~/.ignore
rg --debug test
```

```
[~/testdir]$ ~/test
rg: DEBUG|rg::flags::parse|crates/core/flags/parse.rs:97: no extra arguments found from configuration file
rg: DEBUG|rg::flags::hiargs|crates/core/flags/hiargs.rs:1099: using heuristics to determine whether to read from stdin or search ./ (is_readable_stdin=false, stdin_consumed=false, mode=Search(Standard))
rg: DEBUG|rg::flags::hiargs|crates/core/flags/hiargs.rs:1109: heuristic chose to search ./
rg: DEBUG|rg::flags::hiargs|crates/core/flags/hiargs.rs:1260: found hostname for hyperlink configuration: t480s
rg: DEBUG|rg::flags::hiargs|crates/core/flags/hiargs.rs:1270: hyperlink format: ""
rg: DEBUG|rg::flags::hiargs|crates/core/flags/hiargs.rs:174: using 8 thread(s)
rg: DEBUG|grep_regex::config|crates/regex/src/config.rs:175: assembling HIR from 1 fixed string literals
rg: DEBUG|globset|crates/globset/src/lib.rs:453: built glob set; 0 literals, 0 basenames, 12 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 0 regexes
rg: DEBUG|globset|crates/globset/src/lib.rs:453: built glob set; 8 literals, 0 basenames, 0 extensions, 0 prefixes, 0 suffixes, 0 required extensions, 0 regexes
rg: ./sub/sub2: Permission denied (os error 13)
rg: No files were searched, which means ripgrep probably applied a filter you didn't expect.
Running with --debug will show why files are being skipped.
[~/testdir]$ 
```

`~/testdir/sub/sub2` should be ignored, i.e. the `rg: ./sub/sub2: Permission denied (os error 13)` error is unexpected.

WalterScottYoung:
The previous code deleted too many parts of the path when constructing the absolute path, resulting in a shortened final path. This patch creates the correct absolute path by only removing the necessary parts.

Fixes #2836


WalterScottYoung:
One thing to notice is that commit cad1f5f didn't seems to fully fix https://github.com/BurntSushi/ripgrep/issues/1757, the test in issue 1757 will fail if we add one more directory in the path. I don't know if we should reopen 1757 or something like that to note others this?

The following test from 1757 with "onemore" directory will fail before this fix.

```bash
mkdir tmp
cd tmp
mkdir rust
mkdir -p rust/target/onemore
echo needle > rust/source.rs
echo needle > rust/target/onemore/rustdoc-output.html
echo rust/target/onemore > .ignore
rg --files-with-matches needle
rg --files-with-matches needle rust
echo target >> .ignore
rg --files-with-matches needle rust
```

gstokkink-bex:
@BurntSushi do you have time to review/merge this?

Lucus16:
I can confirm this fixes some bugs I was about to report where some files listed in gitignore glob rules were not excluded depending on the directory that `rg --files` was run from. Thanks!

woess:
This PR also fixes https://github.com/BurntSushi/ripgrep/issues/2778.

beeb:
@BurntSushi sorry for the mention, my project is affected by the underlying bugs fixed in this PR. Would it be possible to review it?

fenuks:
I'm also affected by this bug in ripgrep, but also in https://github.com/sharkdp/fd/issues/1506 which also depends on ignore crate.

reneleonhardt:
Great work, thank you! 👍
Just wondering, one test was enough for negation and nesting?

gstokkink-bex:
@BurntSushi any chance of a merge & release?

WalterScottYoung:
> Great work, thank you! 👍
> Just wondering, one test was enough for negation and nesting?

It's been some time and I don't really remember the detail of this, but I'm pretty sure that the original problem was cause by the path concatenation and the target file's relative depth to where ripgrep was called be positive or 0 or negative was the boardline that this  bug would be triggered, so only tests for those boardline scenarios was added, since more layer of dir won't matter as the bug would be triggered in the first nested layer of dir. Thank you for your question and you can add some more tests that you think is valuable  :-)

PR #2944, opened 2024-12-03
fix missing stats when search interrupted

yorkart:
When `-m` is specified, the `bytes searched` part of  stats is missing or small.

eg:
command： 
```bash
rg --stats -m 10 -e 'keyword' /path/to/file
```

output
```bash
...
(10 lines)
...

10 matches
10 matched lines
1 files contained matches
1 files searched
4535 bytes printed
0 bytes searched   ------------------------>  (the field is missing)
0.001751 seconds spent searching
0.003299 seconds
```


BurntSushi (approved):
Thank you! Nice find!

PR #2945, opened 2024-12-09
ignore: add `inverted_matching` option for `OverrideBuilder`

yuuhikaze:
The `overrides` module description for the `ignore` crate is as follows:
```markdown
The overrides module provides a way to specify a set of override globs.
This provides functionality similar to `--include` or `--exclude` in command
line tools.
```
However there is no way for the user to configure the `OverrideBuilder` to act as an `--exclude`, precisely this PR implements that.
> Usage of the option can be found [here](https://github.com/yuuhikaze/tern/blob/9932b43905ebf07c499904d0234f6b5c5ec173aa/src/converter.rs#L41).

### DIFF
--- a/crates/ignore/src/overrides.rs
+++ b/crates/ignore/src/overrides.rs
@@ -4,7 +4,10 @@ This provides functionality similar to `--include` or `--exclude` in command
 line tools.
 */
 
-use std::path::Path;
+use std::{
+    path::Path,
+    sync::atomic::{AtomicBool, Ordering},
+};
 
 use crate::{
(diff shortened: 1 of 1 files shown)


BurntSushi:
Using shared mutable state like this in a library is totally nuts.

And the presumption here isn't correct anyway. The `--exclude` is done by adding a `!` to beginning of the glob. This is even in the public API documentation.

yuuhikaze:
Good afternoon,
Can you explain me why "using shared mutable state like this in a library is totally nuts"?
That's all
Greetings!

PR #2957, opened 2024-12-30
feat(completion): support sourcing zsh completion dynamically

### LINKED ISSUE #2956: Can't source zsh completions directly

The way I load many completions is by sourcing them in my `.zshrc`, for example

```zsh
  if type fzf > /dev/null; then
	  source <(fzf --zsh)
  fi

  if type gh > /dev/null; then
	source <(TCELL_MINIMIZE=1 gh completion -s zsh)
  fi

  if type fd > /dev/null; then
	  source <(fd --gen-completions)
  fi
```

However, this doesn't work for ripgrep.  If I run `rg --generate=complete-zsh`, I get

```zsh
$ source <(rg --generate=complete-zsh)

_arguments:comparguments:327: can only be called from completion function
```

As a workaround, I can put `rg --generate=complete-zsh > /somewhere/in/fpath/_rg` and it works, but I'm hoping to avoid that

vegerot:
Summary:
Previously, you needed to save the completion script to a file and then
source it.  Now, you can dynamically source completions in zsh by
running

```zsh
$ source <(rg --generate complete-zsh)
```

Test plan:
1. Run `source <(rg --generate complete-zsh)`
2. Run `rg --generate=complete-zs<TAB>`

Before this commit, you would get an error after step 1.
After this commit, it should work as expected.

Closes #2956

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -115,10 +115,16 @@ $ rg --generate complete-fish > "$dir/rg.fish"
 
 For **zsh**:
 
-```
+```zsh
 $ dir="$HOME/.zsh-complete"
 $ mkdir -p "$dir"
 $ rg --generate complete-zsh > "$dir/_rg"
+```
+
+or add the following to your `.zshrc`:
+
+```zsh
+$ source <(rg --generate complete-zsh) # note: this is about 4ms slower than the previous method
 ```
 
 For **PowerShell**, create the completions:
--- a/FAQ.md
+++ b/FAQ.md
@@ -94,7 +94,7 @@ Does ripgrep have support for shell auto-completion?
 
 Yes! If you installed ripgrep through a package manager on a Unix system, then
 the shell completion files included in the release archive should have been
-installed for you automatically. If not, you can generate completes using
+installed for you automatically. If not, you can generate completions using
 ripgrep's command line interface.
 
 For **bash**:
@@ -113,7 +113,7 @@ $ mkdir -p "$dir"
 $ rg --generate complete-fish > "$dir/rg.fish"
 ```
 
-For **zsh**:
+For **zsh**, the recommended approach is:
 
 ```zsh
 $ dir="$HOME/.zsh-complete"
@@ -121,11 +121,22 @@ $ mkdir -p "$dir"
 $ rg --generate complete-zsh > "$dir/_rg"
 ```
 
-or add the following to your `.zshrc`:
+And then add `$HOME/.zsh-complete` to your `fpath` in, e.g., your
+`$HOME/.zshrc` file:
 
 ```zsh
-$ source <(rg --generate complete-zsh) # note: this is about 4ms slower than the previous method
+fpath=($HOME/.zsh-complete $fpath)
 ```
+
+Or if you'd prefer to load and generate completions at the same time, you can
+add the following to your `$HOME/.zshrc` file:
+
+```zsh
+$ source <(rg --generate complete-zsh)
+```
+
+Note though that while this approach is easier to setup, is generally slower
+than the previous method, and will add more time to loading your shell prompt.
 
 For **PowerShell**, create the completions:
 
--- a/crates/core/flags/complete/rg.zsh
+++ b/crates/core/flags/complete/rg.zsh
@@ -434,7 +434,12 @@ _rg_types() {
   fi
 }
 
-_rg "$@"
+# don't run the completion function when being sourced by itself. See #2956
+if [[ $funcstack[1] == _rg ]] || (( ! $+functions[compdef] )); then
+  _rg "$@"
+else
+    compdef _rg rg
+fi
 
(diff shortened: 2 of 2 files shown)


FAQ.md
```diff
@@ -115,12 +115,18 @@ $ rg --generate complete-fish > "$dir/rg.fish"
...
 $ dir="$HOME/.zsh-complete"
 $ mkdir -p "$dir"
 $ rg --generate complete-zsh > "$dir/_rg"
 ```
 
+or add the following to your `.zshrc`:
+
+```zsh
+$ source <(rg --generate complete-zsh) # note: this is about 4ms slower than the previous method
+```
```
BurntSushi, line 128: Why are we suggesting this method? I'm okay with supporting it as long as it doesn't have any costs to doing so, but I don't see a good reason to specifically highlight it in the docs.

BurntSushi (commented):
r? @okdana

FAQ.md
vegerot: Personal preference.  I personally like doing it this way because it makes installation easier.  Since the generation is fast, I don't need to think about installing `_rg` and can just put in my zshrc

```zsh
  if type rg > /dev/null; then
	  eval "$(rg --generate=complete-zsh)"
  fi
```

I assume other people prefer it this way too, which is why projects like [fzf](https://github.com/junegunn/fzf?tab=readme-ov-file#setting-up-shell-integration) suggest it by default.  That is why I suggested it in the doc.

If you disagree, then I can remove this bit 🙂

crates/core/flags/complete/rg.zsh
```diff
@@ -434,7 +434,12 @@ _rg_types() {
   fi
 }
 
-_rg "$@"
+# don't run the completion function when being sourced by itself. See #2956
+if [[ "$funcstack[1]" = "_rg" ]] || ! type compdef >/dev/null 2>&1; then
```
vegerot, line 438: I originally wrote this as 

```zsh
if [[ "$funcstack[1]" = "_rg" ]]; then
  _rg "$@"
elif type compdef >/dev/null;
    compdef _rg rg
fi
```

But `ci/test-complete` didn't like that.  I feel like the test is wrong, because realistically you will either have `compdef` or will run the completion as `_rg`.  I stared at the test for a few seconds and didn't want to spend more time digging into it.

FAQ.md
```diff
@@ -115,12 +115,18 @@ $ rg --generate complete-fish > "$dir/rg.fish"
 
 For **zsh**:
 
-```
+```zsh
 $ dir="$HOME/.zsh-complete"
 $ mkdir -p "$dir"
 $ rg --generate complete-zsh > "$dir/_rg"
 ```
```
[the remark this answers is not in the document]
vegerot: I can do that.

...orr, we could suggest the easier way I'm adding in this commit.

> If you installed ripgrep with a package manager, you should have shell completions automatically (Brew users see [here](https://docs.brew.sh/Shell-Completion))
> However, if completions aren't working, then add to your `zshrc`
> ```zsh
> source <(rg --generate=complete-zsh)
> ```

I like the idea of hoping their package manager does the right thing, but if doesn't then suggesting a simple one-liner instead of a multi-step process of creating directories, files, and adding things to your `fpath`.  On my machine (granted, it's over $1k) it only adds 4ms of overhead which is acceptable imo.

crates/core/flags/complete/rg.zsh
vegerot: Thanks! Done

FAQ.md
vegerot: Based on how difficult it is to explain (see other thread), I think it makes more sense to show new users the easiest method by default, and if they want the more "idiomatic" / performant (slightly) option we can show it as an alternative.  The FAQ is going to be read by people who may not want to create directories, new files, and update their fpath.

vegerot: @okdana @BurntSushi how about I remove the FAQ patch, and we can discuss how to improve the FAQ in a new issue? (continued from https://github.com/BurntSushi/ripgrep/pull/2957#discussion_r1899864739 )

BurntSushi:
I've left it in the FAQ and fixed up the wording by adding an appropriate caveat emptor. I guess since other projects have found it useful to suggest this method it's probably not too big of a deal for ripgrep to do it too. But the text now makes it clear that the "generate and source" approach is slower. 4ms might not seem like much, and indeed, I cannot perceive a 4ms lag, but if you accrue 10 of those kinds of things, it starts to pile up into something that is noticeable.

vegerot:
@BurntSushi I'm using a self-built version now, but jooc when will the next release be so others can use it?

BurntSushi:
I don't know, sorry. See: https://github.com/BurntSushi/ripgrep/blob/master/FAQ.md#when-is-the-next-release

PR #2969, opened 2025-01-10
Clarify the default number of threads

HenrikBengtsson:


BurntSushi:
I'm going to pass on this one. I'd prefer that the language be vague. If I state the heuristic in the docs, then it becomes easier for other folks to rely upon it.

HenrikBengtsson:
That's quite unfortunate, but thanks for thinking about it

PR #2970, opened 2025-01-13
Fix the CI workflow broken by WASI compilation target rename

Kizhyk:
For more info visit the official rust [blog](https://blog.rust-lang.org/2024/04/09/updates-to-rusts-wasi-targets.html).


BurntSushi (approved):
Thanks!

PR #2982, opened 2025-01-30
Adding the .rake extension to default Ruby file types

melopilosyan:
Rake can read from [multiple `.rake` files](https://github.com/ruby/rake/blob/master/doc/rakefile.rdoc#label-Multiple+Rake+Files) other than the `Rakefile` (already listed). And they are widely used in Ruby projects.

Let's also include the `.rake` extension for completeness.


BurntSushi (approved):
Thanks!

BurntSushi:
Oops, duplicate of #2921

PR #3002, opened 2025-02-27
Add tests for binary file detection when using memory maps

Erio-Harrison:
The original code contained a TODO comment to add tests for binary file detection
when using memory maps. This PR implements those tests, noting the differences
between mmap and no-mmap behavior regarding binary detection.

The tests show that with memory maps:
- NUL bytes are only searched for in the first few KB and in matches
- Explicit file paths behave differently than glob patterns
- Matches can appear both before and after NUL bytes
- Binary detection is more permissive with memory maps

### DIFF
--- a/tests/binary.rs
+++ b/tests/binary.rs
@@ -22,8 +22,119 @@ use crate::util::{Dir, TestCommand};
 // searched for in the first few KB of the file and in a match. Normally, NUL
 // bytes are searched for everywhere.
 //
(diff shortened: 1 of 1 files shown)


Erio-Harrison:
> Thanks! I've left feedback, but please don't work on it. I've already resolved the feedback in my rollup branch.

That's cool! Thank you for your work.

PR #3005, opened 2025-03-01
Support search Podfile and .podspec when set type option as ruby

Pancf:


BurntSushi:
I can find no reference that these files are actually evaluated as Ruby code. Examples I see certainly _look_ like Ruby, but authoritative sources seem amazingly vague.

If you can find a better source that these are definitively Ruby, then I'd be happy to re-open this PR.

It's also possible this might be something that warrants a different file type.

PR #3008, opened 2025-03-05
doc(FAQ): add option -d for xargs in example

sharpchen:


BurntSushi:
This doesn't really make sense in the broader context of the prose. If you continue reading, it suggests an even more robust solution than `-d '\n'` by using the NUL terminator. That works in the presence of any kind of whitespace.

PR #3010, opened 2025-03-06
ignore: gracefully quit worker threads upon panic in ParallelVisitor

### LINKED ISSUE #3009: ignore: parallel walker occasionally hangs if visitor panics

master (de4baa10024f2cb62d438596274b9b710e01c59b)

I am able to reproduce this on `master`.

Arch Linux (64-bit, x86):

```shell
> uname -a
Linux terrestrial-gamma-ray-flash 6.13.5-zen1-1-zen #1 ZEN SMP PREEMPT_DYNAMIC Thu, 27 Feb 2025 18:09:27 +0000 x86_64 GNU/Linux
```

### Describe your bug.

`WalkParallel::visit()` will nondeterministically hang if the `ParallelVisitor::visit()` implementation panics. This also occurs when providing a closure to `WalkParallel::run()`.

I am working on a minimal fix to this right now (see https://github.com/BurntSushi/ripgrep/compare/master...cosmicexplorer:ripgrep:parallel-panic-quit?expand=1), but I think it will require a change to the parallelization methods we use from crossbeam, so I wanted to create this issue beforehand. I am currently able to reproduce this with the following:

```diff
diff --git a/crates/ignore/src/walk.rs b/crates/ignore/src/walk.rs
index d6ea9c2..4387fca 100644
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -2221,6 +2221,18 @@ mod tests {
         assert!(!dents[0].path_is_symlink());
     }
 
+    #[test]
+    #[should_panic(expected = "oops!")]
+    fn panic_in_parallel() {
+        let td = tmpdir();
+        wfile(td.path().join("foo.txt"), "");
+
+        WalkBuilder::new(td.path())
+            .threads(40)
+            .build_parallel()
+            .run(|| Box::new(|_| panic!("oops!")));
+    }
+
     #[cfg(unix)] // because symlinks on windows are weird
     #[test]
     fn symlink_loop() {

```

Apply the above diff, then run the following command:

```shell
> cd crates/ignore
> timeout -k 2 1 cargo test --lib -- parallel
```

I typically have to run this many many times on my very fast laptop to get it to trigger. More reliably, you can run:

```shell
> cd crates/ignore
> cargo test
```

And running all the tests at once makes this trigger much more reliably.

The process hangs, typically somewhere in the `Worker::get_work()` method. If I add a print statement within this sub-`loop`: https://github.com/BurntSushi/ripgrep/blob/de4baa10024f2cb62d438596274b9b710e01c59b/crates/ignore/src/walk.rs#L1695-L1706 it will print this out endlessly.

This especially seems to occur more often if the panic handler is non-trivial: I first noticed this when using `color-eyre`, which prints the panic to the console after composing a backtrace.

The `WalkParallel::visit()` method should not hang, and should instead propagate the panic.

My first attempt to solve this (see https://github.com/cosmicexplorer/ripgrep/commit/dd60fa6f3b2de148ca858a0aaf454435342c6629) set the `quit_now` flag if any worker panics, and checked `self.is_quit_now()` within the sub-`loop`. This seems to have made it repro less often, but I am still able to get it to repro regardless, even without modifying the panic handler.

There are some other changes I want to propose to the `ParallelVisitor{,Builder}` traits as well which may make this easier (such as propagating unwind safety), but I'm currently hoping to solve this without making any changes to the external interface. This may require changing the parallelism primitives we use, but maybe not.

cosmicexplorer:
*Fixes #3009.*

# Problem
`WalkParallel::visit()` will nondeterministically hang if the `ParallelVisitor::visit()` implementation panics. This also occurs when providing a closure to `WalkParallel::run()`. Minimal repro is provided in #3009:

```rust
        WalkBuilder::new(path)
            .build_parallel()
            .run(|| Box::new(|_| panic!("oops!")));
```

The above code will nondeterministically hang, because of an infinite loop when no new work is available: https://github.com/BurntSushi/ripgrep/blob/de4baa10024f2cb62d438596274b9b710e01c59b/crates/ignore/src/walk.rs#L1695-L1707

# Solution
1. Check the `quit_now` flag in our wait loop.
2. Catch any panic in the `run()` method and set `quit_now` before propagating the panic.

**Breaking change:** In order to ensure soundness, we also enforce that the filter method provided to `WalkBuilder#filter_entry()` is `UnwindSafe`. Users can circumvent this by wrapping in `AssertUnwindSafe` as needed.

# Result
The added test `panic_in_parallel()` always succeeds instead of hanging.

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -2221,6 +2221,18 @@ mod tests {
         assert!(!dents[0].path_is_symlink());
     }
 
+    #[test]
+    #[should_panic(expected = "oops!")]
+    fn panic_in_parallel() {
+        let td = tmpdir();
+        wfile(td.path().join("foo.txt"), "");
+
+        WalkBuilder::new(td.path())
+            .threads(40)
+            .build_parallel()
+            .run(|| Box::new(|_| panic!("oops!")));
+    }
+
     #[cfg(unix)] // because symlinks on windows are weird
     #[test]
     fn symlink_loop() {
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -1316,6 +1316,7 @@ impl WalkParallel {
         // Retain panic objects and re-throw them outside the thread scope.
         let mut err: Option<Box<dyn std::any::Any + Send>> = None;
         std::thread::scope(|s| {
+            // Spawn a thread per worker, and catch any inner panics.
             let handles: Vec<_> = workers
                 .into_iter()
                 .map(|worker| {
@@ -1348,9 +1349,13 @@ impl WalkParallel {
                     })
                 })
                 .collect();
+            // Iterate through the threads in the order they were spawned.
             for handle in handles {
                 if let Err(e) = handle.join() {
-                    // If any panic occurs, only retain the first.
+                    // If any panic occurs, pick whichever one we get first and discard the rest.
+                    // NB: this *doesn't* correspond to the first thread to actually panic, but just
+                    //     the first thread in the spawn order. If one thread panics as the *result*
+                    //     of another thread panic, we won't have provided complete information.
                     let _ = err.get_or_insert(e);
                 }
             }
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -500,7 +500,15 @@ enum Sorter {
 }
 
 #[derive(Clone)]
-struct Filter(Arc<dyn Fn(&DirEntry) -> bool + Send + Sync + 'static>);
+struct Filter(
+    Arc<
+        dyn Fn(&DirEntry) -> bool
+            + std::panic::UnwindSafe
+            + Send
+            + Sync
+            + 'static,
+    >,
+);
 
 impl std::fmt::Debug for WalkBuilder {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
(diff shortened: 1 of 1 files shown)


BurntSushi:
Closing, but keeping #3010 open.

BurntSushi:
Also, thank you for your work on this! Even though I ended up not taking this PR, your work will be something worth building on for `ignore 0.5`.

cosmicexplorer:
Wanted to confirm here that:
- I always love working with you and your projects,
- this is exactly the choice I would make as a maintainer,
- and thanks so much for your careful and thorough reply!

And I am super glad to hear `ignore 0.5` is on the horizon! 

I would like you to know that as usual, it is extremely difficult to improve upon the kind of work you do. I have tried to argue with your comments in my head and largely failed. I have spent months investigating this, and I have had to introduce some immense complexity in order to make a pretty minor use case more robust. Great work.

I'm also looking at the API, and while I'm not done with mine yet, I would absolutely recommend considering the use of [`ops::ControlFlow`](https://doc.rust-lang.org/std/ops/enum.ControlFlow.html), either internally or in the API. I think using `ControlFlow` may contain strictly more information than the untagged stop token. Personally, I have also separated the worker threads by category, and given each category individually configurable thread counts. I don't know if that will help performance, but I *do* think that the definition of "done" differs based upon whether the current node is a directory or not, which I think may justify slightly different looping techniques.

I also think the deadlock that occurred here is a symptom of a more general misalignment of the threading model to the input distribution. I think work stealing + *recursive* data generation (so each work item can always produce more) is necessarily prone to this deadlock. I do not have a proof of this, and I also do not have a more appropriate answer myself yet. I *do* understand that work stealing is a good approach to avoid having a single very large directory fill up its own queue. I *also* wonder whether there is any benefit to maximizing thread locality.

Finally, I was informed about a very widely available syscall `getdents()`, which is provided in Linux and BSD stdlibs, but not macOS, which writes multiple directory entries into a provided buffer. There is even a `posix_getdents()` from POSIX 2024, and musl supports it, but I have been having a very difficult time getting it accepted to the rust libc crate: https://github.com/rust-lang/libc/pull/4522.

I am actually very confident that supporting `getdents()` will produce a drastic performance improvement for `fs::read_dir()`, so I'm going to try demonstrating that today. If I can demonstrate its utility with benchmarks, I will probably ping you again, and ask you to help support my quest to get this syscall into the stdlib.

`ignore 0.5` may be able to just call upon `fs::read_dir()`, but for pipelining it can be nice to control the buffer size. Note that the alignment of the output is (1) very stable (2) ridiculous.
https://codeberg.org/cosmicexplorer/deep-link/src/commit/1ea3eba5d599d8c48ea56816b7103b12ee49d505/d-major/readdir-sys/src/getdents.rs#L101-L188

Again, very impressed by your engineering judgement, and I appreciate when you write a comment about decisions you *didn't* take too. Keep it up!!

Issue #2990, opened 2025-02-14
Globs that end with a dot don't work properly


tavianator:
### Please tick this box to confirm you have reviewed the above.

- [x] I have a different issue.

### What version of ripgrep are you using?

ripgrep 14.1.1

features:+pcre2
simd(compile):+SSE2,-SSSE3,-AVX2
simd(runtime):+SSE2,+SSSE3,+AVX2

PCRE2 10.43 is available (JIT is available)

### How did you install ripgrep?

pacman

### What operating system are you using ripgrep on?

Arch Linux

### Describe your bug.

Ripgrep seems to ignore glob patterns that end with a dot (`.`)

### What are the steps to reproduce the behavior?

```console
tavianator@graphene $ mkdir asdf asdf.
tavianator@graphene $ touch asdf/foo asdf./foo
```

### What is the actual behavior?

```console
tavianator@graphene $ rg --files -g '!asdf/'
asdf./foo
tavianator@graphene $ rg --files -g '!asdf./'
asdf./foo
asdf/foo
```

### What is the expected behavior?

```console
tavianator@graphene $ rg --files -g '!asdf./'
asdf/foo
```

AnghelescuAlbert:
I'll start working on this now. Let me know if there are any specific guidelines to follow.

PR #3019, opened 2025-03-29
Fix glob pattern to correctly ignore directories with only dots

pi55man:
This PR addresses [issue #2990](https://github.com/BurntSushi/ripgrep/issues/2990), where `ripgrep` did not properly ignore directories named only with dots (e.g., `.`, `..`, `...`) when using negated glob patterns (`-g '!dir'`).  

Previously, the `file_name` function in `globset` only checked if the last byte was a dot (`.`), which led to incorrect behavior. This caused:  
- `rg --files -g '!asdf.'` still listing `asdf./foo`  
- But `rg --files -g '!asdf'` correctly ignoring `asdf/foo`  

### Fix  
The function now ensures that entire filenames made only of dots are ignored by using:  
```rust
if path.iter().all(|&b| b == b'.') {
    return None;
}

### DIFF
--- a/crates/globset/src/pathutil.rs
+++ b/crates/globset/src/pathutil.rs
@@ -7,7 +7,7 @@ use bstr::{ByteSlice, ByteVec};
 /// If the path terminates in `.`, `..`, or consists solely of a root of
 /// prefix, file_name will return None.
 pub(crate) fn file_name<'a>(path: &Cow<'a, [u8]>) -> Option<Cow<'a, [u8]>> {
-    if path.last_byte().map_or(true, |b| b == b'.') {
+    if path.iter().all(|&b| b == b'.') {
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thank you for looking into this!

The status quo was certainly wrong, but this isn't quite right either. This is checking the entire path, which does more work than is necessary (although it is likely to short circuit very early on most paths). It's also not quite the semantics I was going for.

I wanted to [imitate the corresponding routine on `std::path::Path`](https://doc.rust-lang.org/std/path/struct.Path.html#method.file_name), but it looks like I flubbed it.

What this should be doing is checking if the entire file name is `..`, and if it is, returning `None`.

Also, for changes like this, it is customary to include a regression test.

Issue #3020, opened 2025-04-01
gzip suffixed files overlooked by rg


bcling:
### Please tick this box to confirm you have reviewed the above.

- [x] I have a different issue.

### What version of ripgrep are you using?

RHEL 8.10
ripgrep 13.0.0
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)

---

Ubuntu 24.04.2 LTS
ripgrep 14.1.0

features:-simd-accel,+pcre2
simd(compile):+SSE2,-SSSE3,-AVX2
simd(runtime):+SSE2,+SSSE3,+AVX2

PCRE2 10.42 is available (JIT is available)

### How did you install ripgrep?

dnf 
apt-get

### What operating system are you using ripgrep on?

RHEL 8.10
Ubuntu 24.04.2 LTS

### Describe your bug.

ripgrep supports gz extention but fails to detect gz files with gzip extension.

### What are the steps to reproduce the behavior?

1. Create gzip file with gz suffix.
2. rg will search file properly.
3. Rename gz file with gzip suffix.
4. rg will ignore the renamed file.

### What is the actual behavior?

```
$ echo "hello world" > hello.txt
$ gzip hello.txt
$ rg -z hello hello.txt.gz
1:hello world
$ mv hello.txt.gz hello.txt.gzip
$ rg -z hello.txt.gzip
$
```

### What is the expected behavior?

rg should search gzip-suffixed files created by some log appliances.

BurntSushi:
PRs are welcome. It should be pretty easy to add `gzip` here: https://github.com/BurntSushi/ripgrep/blob/de4baa10024f2cb62d438596274b9b710e01c59b/crates/cli/src/decompress.rs#L517

Dylan-Gallagher:
Seems gzip doesn't like those extensions either.

```bash
$ echo "hello world" > hello.txt
$ gzip hello.txt
$ mv hello.txt.gz hello.txt.gzip
$ gzip -d hello.txt.gzip
gzip: hello.txt.gzip: unknown suffix -- ignored
```

BurntSushi:
If gzip doesn't recognize the suffix itself, then I don't think ripgrep should either. The log appliances or whatever that's generating incorrect extensions should be fixed instead.

PR #3021, opened 2025-04-02
Update decompress.rs

bcling:
gzip suffix support added. Seems to work for me.

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -97,7 +97,7 @@ pub(crate) const DEFAULT_TYPES: &[(&[&str], &[&str])] = &[
     ]),
     (&["graphql"], &["*.graphql", "*.graphqls"]),
     (&["groovy"], &["*.groovy", "*.gradle"]),
-    (&["gzip"], &["*.gz", ".gzip", "*.tgz"]),
+    (&["gzip"], &["*.gz", "*.gzip", "*.tgz"]),
     (&["h"], &["*.h", "*.hh", "*.hpp"]),
     (&["haml"], &["*.haml"]),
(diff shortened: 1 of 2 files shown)


ltrzesniewski:
Shouldn't that also be added to `default_types.rs` to make `-tgzip` work?

bcling:
> Shouldn't that also be added to `default_types.rs` to make `-tgzip` work?

It was already in there:

```
(&["gzip"], &["*.gz", "*.tgz"]),
```

ltrzesniewski:
It's in the key, I meant to add it to the value:

```
(&["gzip"], &["*.gz", "*.gzip", "*.tgz"]),
```

bcling:
You were very clear, I failed my code reading comprehension. Added as an additional value as noted above.

crates/ignore/src/default_types.rs
```diff
@@ -97,7 +97,7 @@ pub(crate) const DEFAULT_TYPES: &[(&[&str], &[&str])] = &[
     ]),
     (&["graphql"], &["*.graphql", "*.graphqls"]),
     (&["groovy"], &["*.groovy", "*.gradle"]),
-    (&["gzip"], &["*.gz", "*.tgz"]),
+    (&["gzip"], &["*.gz", ".gzip", "*.tgz"]),
```
ltrzesniewski, line 100: The `*` is missing:

```rust
(&["gzip"], &["*.gz", "*.gzip", "*.tgz"]),
```

bcling:
Ugh. Apologies for making you proof what I should have caught. Thanks for the gentle corrections.

BurntSushi:
See https://github.com/BurntSushi/ripgrep/issues/3020#issuecomment-2786892024

PR #3025, opened 2025-04-09
Implement `IntoIterator` for `GlobSetBuilder`

squidfunk:
Originally reported here: https://github.com/BurntSushi/globset/issues/8

This PR fixes the linked issue. See below:

---

The `Debug` implementation of `GlobSetBuilder` is quite noisy. For better introspectability in consuming packages, it would be perfect if one of two things could be implemented, so the `Debug` implementation can be adjusted (favorite) or is more readable:

1. An `IntoIterator<Item = &Glob>` implementation for `&GlobSetBuilder`, so we can use the `Display` implementation of `Glob` where applicable. This would be a non-invasive addition, as the consumer can decide what style they want.
2. A `Debug` implementation for `GlobSetBuilder` that does 1., i.e., prints `Glob` via `Display`.

I'm happy to send a PR for either of which, if this is a good match.

### Example

``` rust 
#[cfg(test)]
mod tests {
    use globset::{Glob, GlobSetBuilder};

    #[test]
    fn test() -> Result<(), Box<dyn std::error::Error>> {
        let mut builder = GlobSetBuilder::new();
        builder.add(Glob::new("src/**/*.rs")?);
        builder.add(Glob::new("*.txt")?);
        builder.add(Glob::new("docs/**/*.md")?);
        println!("{:#?}", builder);
        Ok(())
    }
}
```

### Current Result

``` rust
GlobSetBuilder {
    pats: [
        Glob {
            glob: "src/**/*.rs",
            re: "(?-u)^src(?:/|/.*/).*\\.rs$",
            opts: GlobOptions {
                case_insensitive: false,
                literal_separator: false,
                backslash_escape: true,
            },
            tokens: Tokens(
                [
                    Literal(
                        's',
                    ),
                    Literal(
                        'r',
                    ),
                    Literal(
                        'c',
                    ),
                    RecursiveZeroOrMore,
                    ZeroOrMore,
                    Literal(
                        '.',
                    ),
                    Literal(
                        'r',
                    ),
                    Literal(
                        's',
                    ),
                ],
            ),
        },
        Glob {
            glob: "*.txt",
            re: "(?-u)^.*\\.txt$",
            opts: GlobOptions {
                case_insensitive: false,
                literal_separator: false,
                backslash_escape: true,
            },
            tokens: Tokens(
                [
                    ZeroOrMore,
                    Literal(
                        '.',
                    ),
                    Literal(
                        't',
                    ),
                    Literal(
                        'x',
                    ),
                    Literal(
                        't',
[...]

PR #3026, opened 2025-04-09
globset: compact Debug impl for GlobSetBuilder

squidfunk:
Supersedes #3025, as discussed in https://github.com/BurntSushi/ripgrep/pull/3025#issuecomment-2790663355 and https://github.com/BurntSushi/ripgrep/pull/3025#issuecomment-2790741277.

### DIFF
--- a/crates/globset/src/lib.rs
+++ b/crates/globset/src/lib.rs
@@ -487,7 +487,7 @@ impl Default for GlobSet {
 
 /// GlobSetBuilder builds a group of patterns that can be used to
 /// simultaneously match a file path.
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct GlobSetBuilder {
     pats: Vec<Glob>,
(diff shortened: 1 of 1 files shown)


BurntSushi:
Thanks! This wasn't quite what I would like. The allocs in the `Debug` impl are somewhat unfortunate. Instead, I implemented a more concise `Debug` impl on `Glob` and then used that in `GlobSetBuilder`.

Ideally we'd have a concise `Debug` impl for `GlobSet` too, but that's harder. And involves some design trade-offs since the original glob patterns are dropped by the time it is constructed.

squidfunk:
Thanks for coming back to this PR! I don't really mind how things are implemented, but more concise debug output would be helpful, alternatively implementing `IntoIterator` as I proposed in https://github.com/BurntSushi/ripgrep/pull/3025, which would allow to customize debug outputs.

PR #3037, opened 2025-04-27
Ensure security of dependency graph

janvhs:
Hi, for a while now I'm a bit skeptical of the steep dependency graphs that a lot of Rust tools have. As a counter measure I decided to "validate" the ones I installed on my system and use regularly. ripgrep is the first one in this series.

This PR would add [cargo-vet](https://github.com/mozilla/cargo-vet), the tool that Mozilla develops and Google, Mozilla and several other projects use to ensure the code they deploy is not containing any malicious code by third parties. It's a proactive tool used for code auditing dependencies incrementally. You can import audits from people or organisations you trust, which leads to a smaller "trust-graph". I am in the process of introducing this tool at my work at SUSE and it seems to become slowly more popular in the Rust community.

This PR adds a few audits and trusts the Windows crates as well as some crates by David Tolnay. I left `jemallocator` and `crossbeam-deque` without any decision, because I am unfamiliar with both crates and their authors and did not have the capacity to audit them. Especially, `jemalloc-sys` has a footprint of 127106 lines and I didn't want to deal with it before getting feedback to this PR. Information of each decision is in the corresponding commit

If you would be interested into improving the auditability further, [cargo-auditable](https://github.com/rust-secure-code/cargo-auditable) seems very nice for embedding a dependency list into the built binary without effecting reproducibility. Embedding this info allows various tools like [OWASP's blint](https://github.com/owasp-dep-scan/blint), [cargo-auditable](https://github.com/rust-secure-code/cargo-auditable), [Google's osv-scanner](https://github.com/google/osv-scanner), [Anchore's syft](https://github.com/anchore/syft) and [Aqua Security's trivy](https://github.com/aquasecurity/trivy) to pick it up from the binary and create SBOMs for a project, whole system or use them in different ways like vulnerability scanning via [grype](https://github.com/anchore/grype). In fact, Go already does this by default and it is a great value add. I actually use part of this info automatically for updating my manually installed Go binaries to the newest version.

### DIFF
--- a/supply-chain/audits.toml
+++ b/supply-chain/audits.toml
@@ -1,7 +1,44 @@
 
 # cargo-vet audits file
 
-[audits]
+[[audits.anyhow]]
+who = "Jan Fooken <jan@faulty.computer>"
+criteria = "safe-to-deploy"
+delta = "1.0.95 -> 1.0.87"
+notes = "Mostly refactoring, because some APIs are now stable"
+
+[[audits.cc]]
+who = "Jan Fooken <jan@faulty.computer>"
+criteria = "safe-to-deploy"
+delta = "1.1.6 -> 1.1.18"
+
+[[audits.lexopt]]
+who = "Jan Fooken <jan@faulty.computer>"
+criteria = "safe-to-deploy"
+version = "0.3.0"
+notes = """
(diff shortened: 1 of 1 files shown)


BurntSushi:
Yes! Thanks for understanding. It has been a lot of work to keep the dependency tree "small."

Issue #1332, opened 2019-07-29
No matches with empty patterns file and invert match


wrodis:
#### What version of ripgrep are you using?

ripgrep 11.0.1
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)

#### How did you install ripgrep?

Compiled from repo.

#### What operating system are you using ripgrep on?

CentOS 7.

#### Describe your question, feature request, or bug.

ripgrep's behaviour seems inconsistent with grep's when using an empty file with --file and inverting match with -v.

```
$ echo text > /tmp/file; touch /tmp/empty_patterns; grep --file /tmp/empty_patterns -v /tmp/file; echo $?; rg --debug --file /tmp/empty_patterns -v /tmp/file; echo $?
text
0
1
```
Not 100% sure if it's a bug or a feature.

BurntSushi:
Interesting. This is _probably_ a bug. For corner cases like these, I've generally deferred to `grep`'s behavior, but this probably should get some careful thought. With that said, just because ripgrep and grep differ, that doesn't mean there's a bug.

One thing that is making this a bit harder to make sense of, is that if ripgrep detects that a match can never occur, then it exits immediately. One of those cases, on master right now, is when the number of patterns given to ripgrep is zero. The thinking there is that if you have nothing to search for, then you can't get any results. So, right now, this doesn't check whether `-v` is given or not. Whether `-v` should mean "match everything" even when there are no patterns isn't quite clear to me. Do you have a specific use case in mind?

Interestingly, if I _remove_ the optimization that causes ripgrep to bail early, then an empty pattern file gets compiled to an empty regex, and an [empty regex correctly matches at every position](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=d80819c8a984c4c07b17fc48123f6c98). So the question here, to me, is how an empty pattern file should be interpreted. Is it a single empty regex? Or is it zero regexes? If it's an empty regex, then `-v` should turn up nothing, but if it's zero regexes, then `-v` should turn up everything. Presumably, the latter is the choice grep made. It is possible to provide a single empty regex (provide a pattern file with a single new line), so it does kind of seem like an empty pattern file should be treated as providing no regexes. In which case, grep's behavior here is probably correct.

Do you have a particular use case in mind here, or did you just happen to notice a curious inconsistency?

wrodis:
[...]

PR #3041, opened 2025-05-08
Fix the "No matches with empty patterns file and invert match" bug

xenoninja:
This PR addresses https://github.com/BurntSushi/ripgrep/issues/1332.
The problem is when an empty pattern file is given, ripgrep sees the pattern as "an empty regex" not "zero regexes".

> | pattern file | grep's output | rg's output |
> 
> |:-|:-:|:-:|
> 
> | **empty** | everything |  nothing |
> 
> | **newline** | nothing | nothing |
> 

Now the fix works out as below

| pattern file | grep's output | rg's output |

|:-|:-:|:-:|

| **empty** | everything |  **everything** |

| **newline** | nothing | nothing |


BurntSushi:
Thank you for this PR. Unfortunately, this was not quite correct. The optimization needs to be a hair smarter and the PCRE2 regex builder had to be fixed to correctly construct a pattern that can't match when there are zero patterns. And this should also come with a regression test.

I have this fixed in a branch that I'll release at some point soon.

xenoninja:
Oh, sorry about the mistaken PR. I've seen the fix commit. Thanks for your effort!

PR #3048, opened 2025-05-12
Add support for nested alternates

lukesandberg:
This adds support for nesting alternates.

e.g. `**/{node_modules/**/*/{ts,js},crates/**/*.{rs,toml}}` would select all ts/js files found under `node_modules` and all `rs/toml` files found under `crates` in a single glob pattern.


BurntSushi (approved):
Nice, thank you!!!

PR #3057, opened 2025-05-29
build: emit warning if git is missing during build

kevichi7:
Adds a cargo:warning if git is not available during build.
This makes it easier to understand why RIPGREP_BUILD_GIT_HASH might be missing, without breaking the build.


kevichi7:
Thanks so much for taking a look and for the fix! Learned something new about stdout vs stderr here.

PR #3077, opened 2025-06-26
Fix semantic versioning in README.md

VlxtIykg:


VlxtIykg:
I ran it myself on wsl (Ubuntu 24.04.2 LTS).
Could test on other distro/versions but I would assume they work universally so long as it's debian based.

BurntSushi:
Duplicate of #3058

Issue #3089, opened 2025-07-04
Add type list support for ColdFusion/CFML and BoxLang


JamoCA:
Please consider adding built-in "Type List" support for [ColdFusion](https://www.adobe.com/products/coldfusion-family.html)/[CFML](https://www.lucee.org/) and [BoxLang](https://boxlang.io/) file types.

```
cfml: *.cfc, *.cfm

boxlang: *.bx, *.bxm, *.bxs
```

If "case" is important, please update it to accept either character (ie, `[Cc][Ff][Cc]`) to support older code (as some CFML files could be ~30 years old).

*I considered using `--type-add`, but it's not persistent and the flag only applies to the current command.*

Thanks.

BurntSushi:
I don't keep issues tracking the default file type list. It's too much and not worth it. If you want things added to the default type list, then please submit a PR.

PR #3090, opened 2025-07-05
Add Type List Support for ColdFusion/CFML & BoxLang

JamoCA:
Requested here and now submitting a PR.
https://github.com/BurntSushi/ripgrep/issues/3089#issuecomment-3037430356


crates/ignore/src/default_types.rs
```diff
@@ -39,6 +40,7 @@ pub(crate) const DEFAULT_TYPES: &[(&[&str], &[&str])] = &[
     (&["carp"], &["*.carp"]),
     (&["cbor"], &["*.cbor"]),
     (&["ceylon"], &["*.ceylon"]),
+    (&["cfml"], &["*.[Cc][Ff][Cc]", "*.[Cc][Ff][Mm]"]),
```
BurntSushi, line 43: I've changed this to `*.cfc` and `*.cfm`. It doesn't make sense for everyone using this file type to pay for case insensitive matching all of the time. For users where this matters, they can use the `--glob-case-insensitive` flag.

Issue #3081, opened 2025-07-01
Feature: Add `id` to Glob to allow GlobSet identifying/reporting which one matched


reneleonhardt:
#### Describe your feature request

`GlobSet` is very useful, but at the moment it can't report which specific Glob matched, it only reports true/false after iterating patterns.

Current function:
```rust
pub fn is_match_candidate(&self, path: &Candidate<'_>) -> bool {
```

New function:
```rust
pub fn get_match_id(&self, path: &Candidate<'_>) -> Option<String> {
```

```rust
pub struct Glob {
    id: String, // Identifier
    glob: String,
    re: String,
    opts: GlobOptions,
    tokens: Tokens,
}
```

This would allow to process even hundreds of Globs (maybe even in parallel with futures-concurrency) while still being able to identify which exact pattern matched and process a corresponding mapping accordingly (i.e. ".gitignore discarded Cargo.lock").

BurntSushi:
I'm open to this, but as mentioned in #3092, the IDs should be auto-assigned as monotonically increasing integers starting from `0`. Similar to how `regex-automata` and `aho-corasick` work. Attaching `String` values like suggested is bad juju and imposes extra costs that don't need to be paid for all such use cases.

PR #3092, opened 2025-07-05
add  the ID field

melroy12:
solves the #3081


BurntSushi:
This isn't how I'd want to go about this unfortunately. What I'm willing to do is attached a monotonically increasing integer to each glob (starting at `0`), similar to how `regex-automata` treats multi-pattern regexes. But attaching arbitrary strings imposes extra costs for everyone _and_ requires users of the library to opt into it.

Issue #3076, opened 2025-06-24
`--max-count` not working properly with `--multiline` if matches are on consecutive lines


underflow00:
### Please tick this box to confirm you have reviewed the above.

- [x] I have a different issue.

### What version of ripgrep are you using?

ripgrep 14.1.1 (rev 4649aa9700)

features:+pcre2

PCRE2 10.43 is available (JIT is unavailable)

### How did you install ripgrep?

Download from https://github.com/BurntSushi/ripgrep/releases/tag/14.1.1

### What operating system are you using ripgrep on?

Windows 11

### Describe your bug.

For the purpose of limiting count by `--max-count`, with the mutiline flag, the program seems to be counting matches on consecutive lines as a single match, resulting in extra matches.

### What are the steps to reproduce the behavior?

1. Create a file with contents
```
line 2
line 3 x
line 2
line 3
```

2. Run `rg --max-count=1 -U "line 2\nline 3" <file> --stats --debug`.

### What is the actual behavior?

```
rg: DEBUG|rg::flags::parse|crates/core\flags\parse.rs:97: no extra arguments found from configuration file
rg: DEBUG|rg::flags::hiargs|crates/core\flags\hiargs.rs:1083: number of paths given to search: 1
rg: DEBUG|rg::flags::hiargs|crates/core\flags\hiargs.rs:1094: is_one_file? true
rg: DEBUG|rg::flags::hiargs|crates/core\flags\hiargs.rs:1269: found hostname for hyperlink configuration: d1
rg: DEBUG|rg::flags::hiargs|crates/core\flags\hiargs.rs:1279: hyperlink format: ""
rg: DEBUG|rg::flags::hiargs|crates/core\flags\hiargs.rs:174: using 1 thread(s)
rg: DEBUG|grep_cli::decompress|crates\cli\src\decompress.rs:502: gzip: could not find executable in PATH
rg: DEBUG|grep_cli::decompress|crates\cli\src\decompress.rs:502: gzip: could not find executable in PATH
rg: DEBUG|grep_cli::decompress|crates\cli\src\decompress.rs:502: bzip2: could not find executable in PATH
rg: DEBUG|grep_cli::decompress|crates\cli\src\decompress.rs:502: bzip2: could not find executable in PATH
rg: DEBUG|grep_cli::decompress|crates\cli\src\decompress.rs:502: xz: could not find executable in PATH
[...]

Issue #1311, opened 2019-06-23
Unexpected (or incorrect?) multiline replace behavior


mqudsi:
#### What version of ripgrep are you using?

ripgrep 11.0.1 (rev 34677d2622)
+SIMD -AVX (compiled)
+SIMD +AVX (runtime)

#### How did you install ripgrep?

cargo from git

#### What operating system are you using ripgrep on?

Distributor ID:	Debian
Description:	Debian GNU/Linux 10 (buster)
Release:	10
Codename:	buster

#### Describe your question, feature request, or bug.

I was trying to replace some of my usages of other commands (`sed`, `tr`, `ag`) with ripgrep now that it has proper multiline and pcre2 support features, and ran into the following odd behavior:

```
> printf "hello\nworld\n" | rg -U "\n" --color always 
hello
world 
```
This works fine, everything matches, nothing is highlighted because the newline can't be colored.

But when I try to replace, I get the following behavior:

```
> printf "hello\nworld\n" | rg -U "\n" -r "?" 
hello?
world?
```

which isn't correct because the newline has been correctly matched but incorrectly retained in the output.

Whereas the following does work:

```
> printf "hello\nworld\n" | rg --null-data "\n" -r "?"
hello?world?
```

But I don't see any documented reason why the former shouldn't give the same output as well. I can understand technically why this is happening (rg internally preserves the concept of lines but merely glosses over them for search purposes with `-U`), but the confluence of both `-U` and `-r` results in incorrect behavior: the output should either include the replacement character and no new lines (desirable) or the newlines and no replacement character (unfortunate but understandable limitation). I don't think including both can be considered correct, however.

mqudsi:
Actually, this has to be a mere oversight because the following *does* work:

```
> printf "hello\nworld\n" | rg -U o\nw -r '?'
hell?orld
```

learnbyexample:
My understanding is that `-U` doesn't strictly behave as if entire input is single input string. Rather, it allows the match to cross multiple lines when necessary, like the `printf "hello\nworld\n" | rg -U 'o\nw' -r '?'` example.

You might feel `--null-data` is working, but that is simply adding NUL character for each block of output, instead of newline. It so happens you have only one block in your example as there is no NUL character in input. `printf "hello\nworld\n" | perl -0777 -pe 's/\n/?/g'` is an example for slurping entire input as single string.

mqudsi:
[...]

PR #3094, opened 2025-07-06
fix max count for multiline matches on adjacent lines

fspv:
Fixes https://github.com/BurntSushi/ripgrep/issues/3076

### DIFF
--- a/crates/core/flags/hiargs.rs
+++ b/crates/core/flags/hiargs.rs
@@ -714,7 +714,8 @@ impl HiArgs {
             .line_number(self.line_number)
             .multi_line(self.multiline)
             .memory_map(self.mmap_choice.clone())
-            .stop_on_nonmatch(self.stop_on_nonmatch);
+            .stop_on_nonmatch(self.stop_on_nonmatch)
+            .max_matches(self.max_count);
         match self.context {
             ContextMode::Passthru => {
                 builder.passthru(true);
--- a/crates/printer/src/standard.rs
+++ b/crates/printer/src/standard.rs
@@ -2890,6 +2890,33 @@ Holmeses, success in the province of detective work must always
     }
 
     #[test]
+    fn max_matches_multi_line3() {
+        let matcher =
+            RegexMatcher::new(r"line 2\nline 3").unwrap();
+        let mut printer = StandardBuilder::new()
+            .max_matches(Some(1))
+            .build(NoColor::new(vec![]));
+        SearcherBuilder::new()
+            .line_number(false)
+            .multi_line(true)
+            .max_matches(Some(1))
+            .build()
+            .search_reader(
+                &matcher,
+                "line 2\nline 3 x\nline 2\nline 3\n".as_bytes(),
+                printer.sink(&matcher),
+            )
+            .unwrap();
+
+        let got = printer_contents(&mut printer);
+        let expected = "\
+line 2
+line 3 x
+";
+        assert_eq_printed!(expected, got);
+    }
+
+    #[test]
     fn only_matching() {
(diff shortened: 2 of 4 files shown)


fspv:
Hey @BurntSushi. Just getting back to this, don't want to waste the effort. Any chance you can take a look at this? (see more context in https://github.com/BurntSushi/ripgrep/issues/3076)

BurntSushi:
Thank you for working on this! Unfortunately, this is a rather difficult bug to fix. Even if #1311 were reverted, then using a pattern like `line 2\nline 3|x\nline 2\n` will still given an unexpected result:

```rust
    #[test]
    fn max_matches_multi_line4() {
        let matcher =
            RegexMatcher::new(r"line 2\nline 3|x\nline 2\n").unwrap();
        let mut printer = StandardBuilder::new()
            .max_matches(Some(1))
            .build(NoColor::new(vec![]));
        SearcherBuilder::new()
            .line_number(false)
            .multi_line(true)
            .build()
            .search_reader(
                &matcher,
                "line 2\nline 3 x\nline 2\nline 3 x\n".as_bytes(),
                printer.sink(&matcher),
            )
            .unwrap();

        let got = printer_contents(&mut printer);
        let expected = "\
line 2
line 3 x
";
        assert_eq_printed!(expected, got);
    }
```

gives

```
expected:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
line 2
line 3 x

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

got:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
line 2
line 3 x
line 2

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

It provokes the same sort of underlying problem. Indeed, this test fails with this PR too.

I'm going to try and think more about this problem. I wonder if perhaps #1311 needs to be fixed in a different way.

BurntSushi:
I think your instinct to put this in the `Searcher` is probably correct. I don't really see how this can be fixed in the printer easily. The problem is that the `SinkMatch` passed to the printer is fundamentally violating some implicit contracts here. e.g., That every match found in `SinkMatch::buffer` is a valid match that should be printed and handled. When multi-line mode is disabled, this is fine, because `--max-count` is specifically defined to be at the granularity of a _line_. But with multi-line mode, a single match can extend over multiple lines and that should only be counted once.

The other source of tension here is that the `Searcher` tries very hard to avoid not only finding literally every match (since one line may contain multiple matches and the most basic functionality of a grep does not require finding all matches on each line, just that a line matches _somewhere_, once or more) but also to avoid finding the starting position of a match if it isn't needed (because doing so is extra work). This is why a lot of stuff is pushed down in the printer and the _printer_ decides if all of the matches need to be found, not the searcher. But in order to implement `--max-count` when matches can span multiple lines, the searcher does need to find all matches.

I think in the multi-line case this is okay...

But it is not ideal to have a `max_matches` option on both the printer and the searcher. So I'm going to look into adding it to the searcher and removing it from the printer. And also having it only work for multi-line searches is very weird.

If anyone ever wonders why "just split logic out into their own crates" is hard, this is yet another example of it. I clearly fucked up the abstraction boundaries, and because there's a semver boundary at those abstraction boundaries, introducing coupling or tweaking those boundaries is annoyingly difficult.

PR #3102, opened 2025-07-13
complete/zsh: improve --hyperlink-format completion

okdana:
here's better completion for `--hyperlink-format`

three known annoyances:

* zsh doesn't perform brace expansion on single words, so users might be inclined to leave the variable braces unquoted, as in `--hyperlink-format=foo://{path}`. but i guess it generally assumes that it might and handles it at an earlier stage, which means our function doesn't even know the `{` is there. this causes unexpected behaviour. there is also a bug (i assume zsh's rather than mine but idk) where it re-inserts the `{` in a weird place, e.g. `\{ho{st\}`

* zsh converts the the argument in `--opt=arg` style options to the `\`-escaped (rather than quoted) form when completing it, which is not pleasant with the syntax used here and exacerbates the above issue

* if you have like `--hyperlink-format '{}`, press tab between the `{}`, and then press `}` after adding one of the matches, it won't swallow the extra `}`

off the top of my head idk if there's anything to be done about the first two. the last one can probably be fixed with `compadd -R` or sth. but tbh i got bored of the problem

if these limitations bother you i can simplify it so it only tries to complete the aliases i guess

i also made it so helper functions aren't re-defined if they exist. this is a convention of many completion functions that allows end-users to more easily define their own helpers if they want them to behave differently. i just forgot to do it before

### DIFF
--- a/crates/core/flags/complete/rg.zsh
+++ b/crates/core/flags/complete/rg.zsh
@@ -319,7 +319,7 @@ _rg() {
     '--field-context-separator[set string to delimit fields in context lines]'
     '--field-match-separator[set string to delimit fields in matching lines]'
     '--hostname-bin=[executable for getting system hostname]:hostname executable:_command_names -e'
-    '--hyperlink-format=[specify pattern for hyperlinks]:pattern'
+    '--hyperlink-format=[specify pattern for hyperlinks]: :_rg_hyperlink_formats'
     '--trace[show more verbose debug messages]'
     '--dfa-size-limit=[specify upper size limit of generated DFA]:DFA size (bytes)'
     "(1 stats)--files[show each file that would be searched (but don't search)]"
@@ -409,6 +409,7 @@ _rg() {
 }
 
 # Complete encodings
+(( $+functions[_rg_encodings] )) ||
 _rg_encodings() {
   local -a expl
   local -aU _encodings
@@ -421,6 +422,7 @@ _rg_encodings() {
 }
 
 # Complete file types
+(( $+functions[_rg_types] )) ||
 _rg_types() {
   local -a expl
   local -aU _types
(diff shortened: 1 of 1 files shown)


ltrzesniewski:
We were discussing with @ilyagr about the removal of duplicate hyperlink alias names in #3096  ([here's my draft](https://github.com/BurntSushi/ripgrep/compare/3b7fd442a6f3aa73f650e763d7cbb902c03d700e...ltrzesniewski:ripgrep:41a7d17349b87555f3bc2adb6fe7dd5804ba7a80) of this idea if you're interested).

If this would end up being accepted, what would you think about adding a `!HYPERLINK_ALIAS_NAMES!` variable (or similar) to be replaced, just like `!ENCODINGS!` is today?

okdana:
oh. i was semi-conscious of the fact that i was getting e-mails about that, which is why i thought to do this. but i must've forgot about it already by the time i went to actually work on it, i never even looked at it

yes, that would be convenient

though i think, in contrast to the encodings, it makes sense for these to have descriptions for what they actually do. since they're auto-generated it could be as simple as the strings they expand to. so something like `file:file://{host}{path}` (or whatever delimiter)? if that's too cute just the names will work ofc, it's just not quite as helpful

then again the man page isn't explicit about them either so w/e

ltrzesniewski:
Either having the strings they expand to or adding a custom description to `HYPERLINK_PATTERN_ALIASES` would be possible. I think I'd also prefer having a description, especially since I like the ones you provided.

Lots of variants are possible here, and also for the main page.

ltrzesniewski:
I made #3103 to be used as a baseline, but didn't add descriptions yet since they wouldn't be used. Adding them won't be a problem if needed.

BurntSushi:
Thank you! I took this as-is, except I added descriptions to the new `HyperlinkAlias` and auto-generated that part of the completion script.

PR #3107, opened 2025-07-24
Add line highlighting support for matching lines

### LINKED ISSUE #3024: Please allow setting color for matching line
First of all - thank you for amazing tool. It's beyond great.

Sometimes I'm searching for short/small matches, and need to see context. While the match can be highlighted, it's not always obvious where it is.

For example, using gnu grep, I did set:

`export GREP_COLORS="sl=48;5;240:mt=38;5;226;48;5;124"`

and this lets me see matches like this:

Image

As far as I understand `rg --help` - I can't have this configured like this in ripgrep.

I think it would be worthy addition.

emrebengue:
Implement a new "highlight" feature for the '--colors' flag that allows users to highlight entire matching lines with background color.

 It is important to note that this feature gets disabled if '--colors' flag is used with 'match:none' as it depends on matching detection.

Documentation in crates/core/flags/defs.rs is updated.

Fixes #3024

### DIFF
--- a/crates/printer/src/standard.rs
+++ b/crates/printer/src/standard.rs
@@ -1322,6 +1322,7 @@ impl<'a, M: Matcher, W: WriteColor> StandardImpl<'a, M, W> {
             self.write(&bytes[line])?;
             return Ok(());
         }
+        self.start_line_highlight()?;
         while !line.is_empty() {
             if matches[*match_index].end() <= line.start() {
                 if *match_index + 1 < matches.len() {
@@ -1348,6 +1349,7 @@ impl<'a, M: Matcher, W: WriteColor> StandardImpl<'a, M, W> {
             }
         }
         self.end_color_match()?;
+        self.end_line_highlight()?;
         Ok(())
     }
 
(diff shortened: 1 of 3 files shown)


crates/printer/src/standard.rs
```diff
@@ -1462,7 +1464,7 @@ impl<'a, M: Matcher, W: WriteColor> StandardImpl<'a, M, W> {
             }
             let remainder = format!(
                 "WARNING: stopped searching binary file after match \
-                 (found {:?} byte around offset {})\n",
+                (found {:?} byte around offset {})\n",
```
BurntSushi, line 1467: These sorts of unrelated changes are really annoying IMO.

crates/printer/src/standard.rs
```diff
@@ -1540,7 +1542,12 @@ impl<'a, M: Matcher, W: WriteColor> StandardImpl<'a, M, W> {
         if self.in_color_match.get() {
             return Ok(());
         }
-        self.wtr().borrow_mut().set_color(self.config().colors.matched())?;
+        let color_spec = if self.highlight_on() {
+            self.match_spec_with_highlight()
+        } else {
+            self.config().colors.matched().clone()
+        };
```
BurntSushi, line 1549: This is FUBAR. In `match_spec_with_highlight`, you're doing shenanigans based on whether specific stylings are set. That's no good and it assumes too much about how the user's _intention_ when using `highlight`. If the user wants the background color to be consistent, for example, then they'll need to set it for both the `matched` and `highlight` types.

PR #3111, opened 2025-07-28
ignore: look for .gitignore's in Jujutsu repos

atalii:
In the context of a non-colocated Jujutsu (jj) repository, .gitignore's retain their meaning, but .git directories aren't present. This commit amends ignore to look for jj repos when deciding whether or not to also acknowledge a .gitignore.

I've of course tested this patch and everything that I can see works fine, but do let me know if I've missed anything. Thanks a ton!


BurntSushi:
Duplicate of #2842

atalii:
Oh, looks like I missed that. Thanks for the quick response and sorry for the duplicate PR.

PR #3112, opened 2025-07-28
globset: allow nested alternatives

senekor:
Hello. I have a use case where it would be great if nested alternatives were supported in globset. Namely, the [Helix editor](github.com/helix-editor/helix) uses globset for [Editorconfig](https://spec.editorconfig.org/), which allows nested alternatives. It is used in the real world by the Linux kernel (see the [Linux editorconfig](https://github.com/torvalds/linux/blob/master/.editorconfig#L5) with nested alternatives).

Unix shells like Bash and Fish support this kind of syntax and the documentation of globset states that "Nesting `{...}` is not **currently** allowed" (emphasis mine)

So I assume this feature is in-scope for the globset crate. I'm happy to improve the implementation if there is any feedback. Thank you for taking a look!

### DIFF
--- a/crates/globset/src/glob.rs
+++ b/crates/globset/src/glob.rs
@@ -564,7 +564,7 @@ impl<'a> GlobBuilder<'a> {
     pub fn build(&self) -> Result<Glob, Error> {
         let mut p = Parser {
             glob: &self.glob,
-            stack: vec![Tokens::default()],
+            stack: vec![vec![Tokens::default()]],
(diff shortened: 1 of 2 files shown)


BurntSushi:
Looks like a duplicate of #3048? (I haven't yet looked at that PR or this one, so I don't know the differences between them.)

senekor:
Oh yeah, that seems to be the case. Apologies, I only searched issues and not PRs before posting.

Issue #3108, opened 2025-07-24
`-q` inverts status code from `--files-without-match`


BatmanAoD:
### Please tick this box to confirm you have reviewed the above.

- [x] I have a different issue.

### What version of ripgrep are you using?

ripgrep 14.1.0 (rev e50df40a19)

features:-simd-accel,+pcre2
simd(compile):+NEON
simd(runtime):+NEON

PCRE2 10.42 is available (JIT is available)

### How did you install ripgrep?

`cargo binstall`

### What operating system are you using ripgrep on?

macOS Seqouia 15.5, ARM

### Describe your bug.

The exit code from `rg -q --files-without-match` is inverted from the exit code from `rg --files-without-match`. As far as I can tell, this is not documented, so I don't believe it's intentional. There is a parenthetical in the `-q` helptext that says the exit code "will be an error code if no matches are found", but since it's a parenthetical, I interpret that as explaining the default exit code behavior.

If this is the intended behavior, I think that should be more explicit in the helptext; perhaps something like:

> The exit code will be an error code if no matches are found. This overrides other flags that may change the exit code, such as --files-without-match.

### What are the steps to reproduce the behavior?

```sh
$> echo haystack ...... | rg -q needle; echo $?
1
$> echo haystack needle | rg -q needle; echo $?
0
$> echo haystack ...... | rg --files-without-match needle > /dev/null; echo $?
0
$> echo haystack needle | rg --files-without-match needle > /dev/null; echo $?
1
$> echo haystack ...... | rg -q --files-without-match needle > /dev/null; echo $?
1
$> echo haystack needle | rg -q --files-without-match needle > /dev/null; echo $?
0
```

### What is the actual behavior?

`-q` inverts the exit code, effectively ignoring `--files-without-match`.

### What is the expected behavior?

I would expect `-q` not to change the exit code.

BurntSushi:
This is definitely a bug.

emrebengue:
I think the issue is the ExitCode logic in run function inside `crates/core/main.rs`. 

I do not quite understand why this behaviour happens. When you look at `ripgrep/crates/printer/src/summary.rs` you will see that has_match() correctly returns true IF config.kind is SummaryKind::PathWithoutMatch and match_count is 0. Which is the correct logic because it is expected to return true if the path/file does not contain a match. 

But this creates an issue when it comes to ExitCode logic.
Adding the following logic to crates/core/main.rs inside `run` function seems to fix the bug for now.

```
[...]

PR #3118, opened 2025-08-02
fix: exit code for -q --files-without-match combination

wisarootl:
This is to resolve issue #3108 

Exit codes were inverted when using -q with --files-without-match. Quiet mode was overriding the search mode logic.

Before:
```
echo "no match" | rg -q --files-without-match pattern  # returned 1 ❌
echo "has match" | rg -q --files-without-match pattern # returned 0 ❌
```

After:
```
echo "no match" | rg -q --files-without-match pattern  # returns 0 ✅
echo "has match" | rg -q --files-without-match pattern # returns 1 ✅
```

Fix: Don't let quiet mode override FilesWithoutMatch summary kind.

One line change in hiargs.rs printer logic.

### DIFF
--- a/crates/core/flags/hiargs.rs
+++ b/crates/core/flags/hiargs.rs
@@ -561,7 +561,7 @@ impl HiArgs {
         search_mode: SearchMode,
         wtr: W,
     ) -> Printer<W> {
-        let summary_kind = if self.quiet {
+        let summary_kind = if self.quiet && !matches!(search_mode, SearchMode::FilesWithoutMatch) {
             SummaryKind::Quiet
         } else {
             match search_mode {


BurntSushi:
This change is bunk unfortunately. ripgrep's test suite doesn't check that `-q --files-without-match` emits nothing on stdout, but this patch makes it... not be quiet any more.

The problem is that the summary printer doesn't distinguish "quiet" mode with "quiet inverted match" mode. This is an API design bug in ripgrep's internal libraries. Sigh.

Nice find though.

BurntSushi:
Specifically, with this PR, this test fails:

```
// See: https://github.com/BurntSushi/ripgrep/issues/3108
rgtest!(r3108_files_without_match_quiet_exit, |dir: Dir, _: TestCommand| {
    dir.create("yes-match", "abc");
    dir.create("non-match", "xyz");

    dir.command().args(&["-q", "abc", "non-match"]).assert_exit_code(1);
    dir.command().args(&["-q", "abc", "yes-match"]).assert_exit_code(0);
    dir.command()
        .args(&["--files-with-matches", "-q", "abc", "non-match"])
        .assert_exit_code(1);
    dir.command()
        .args(&["--files-with-matches", "-q", "abc", "yes-match"])
        .assert_exit_code(0);

    dir.command()
        .args(&["--files-without-match", "abc", "non-match"])
        .assert_exit_code(0);
    dir.command()
        .args(&["--files-without-match", "abc", "yes-match"])
        .assert_exit_code(1);

    let got = dir
        .command()
        .args(&["--files-without-match", "abc", "non-match"])
        .stdout();
    eqnice!("non-match\n", got);

    dir.command()
        .args(&["--files-without-match", "-q", "abc", "non-match"])
        .assert_exit_code(0);
    dir.command()
        .args(&["--files-without-match", "-q", "abc", "yes-match"])
        .assert_exit_code(1);

    let got = dir
        .command()
        .args(&["--files-without-match", "-q", "abc", "non-match"])
        .stdout();
    eqnice!("\n", got);
});
```

with:

```
running 1 test
test regression::r3108_files_without_match_quiet_exit ... FAILED

failures:

---- regression::r3108_files_without_match_quiet_exit stdout ----

thread 'regression::r3108_files_without_match_quiet_exit' (1875804) panicked at tests/regression.rs:1530:5:

printed outputs differ!

expected:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~


~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

got:
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
non-match

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

BurntSushi:
This part in the code is the problem:

https://github.com/BurntSushi/ripgrep.git/blob/119a58a400ea948c2d2b0cd4ec58361e74478641/crates/printer/src/summary.rs#L506-L511

For `SummaryKind::PathWithoutMatch`, the case of whether a match has occurred or not is correctly inverted. But for `SummaryKind::Quiet`, the printer doesn't know we want to invert the match.

PR #2981, opened 2025-01-30
Update README.md with working instructions for RHEL/CentOS/Rocky Linux 9

wackget:
Resolves #2924 

Updated/working instructions for installing ripgrep via EPEL repos on RHEL/CentOS/Rocky 9

PR #3124, opened 2025-08-13
doc: update CentOS/RHEL instructions

carlwgeorge:
The copr repo is no longer needed as ripgrep has been available from EPEL for several years now.


BurntSushi:
Oh, this is much better than #2981! I'll take this one instead.

BurntSushi:
These instructions are apparently insufficient according to #2981.

PR #3126, opened 2025-08-15
docs: Add installation method with gah

marverix:
Hello!
This PR adds to the documentation installation method using [gah](https://github.com/marverix/gah):
> gah is an GitHub Releases app installer, that DOES NOT REQUIRE SUDO! It is a simple bash script that downloads the latest release of an app from GitHub and installs it in ~/.local/bin. It is designed to be used with apps that are distributed as a single binary file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,6 +324,12 @@ If you're a **Flox** user, you can install ripgrep as follows:
 $ flox install ripgrep
 ```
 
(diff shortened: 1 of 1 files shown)


BurntSushi:
I appreciate the contribution, but I don't think `gah` is popular enough for me to include in an already long list.

PR #3136, opened 2025-08-31
Emacs hyperlink alias

purplefishies:
Enables emacs users to have hyperlink support with named emacs hyperlink type


ltrzesniewski:
Thanks, but is the `emacs://` scheme something official? I couldn't find it being defined anywhere.

BurntSushi:
Yes, please include a link to some authoritative source supporting this. Because once it's added to ripgrep, removing or changing could be difficult. So it's important to be sure that this is correct.

BurntSushi:
Feel free to re-open this when there is a good source for this alias.

purplefishies:
Hi Andrew,

Maybe we could suspend this. I'm working with the emacs community to
discuss adding this uri handler.


Sorry about the delay letting you know.

Sincerely,

Jimi

PR #3137, opened 2025-09-01
fix example .ripgreprc in GUIDE.md

turkishmaid:
`!.git/*` does not do what the example suggests, i.e. include general hidden files and directories, _except_ `.git/` folder contents. Matches inside `.git/` folders will instead be returned. Without the `*` it works as expected. Verified with `zsh`.

### DIFF
--- a/GUIDE.md
+++ b/GUIDE.md
@@ -579,7 +579,7 @@ web:*.{html,css,js}*
 
 # or
 --glob
-!.git/*
+!.git/
 
 # Set the colors.
 --colors=line:none


BurntSushi:
This change doesn't make any sense.

turkishmaid:
@BurntSushi – Well, I thought the same, but see this transcript, verbatim from my console:


```
% export -p | grep RIPGREP_CONFIG_PATH
export RIPGREP_CONFIG_PATH=/Users/sara/.config/ripgrep/rc



% cat $RIPGREP_CONFIG_PATH
# Don't let ripgrep vomit really long lines to my terminal, and show a preview.
--max-columns=150
--max-columns-preview

# Search hidden files / directories (e.g. dotfiles) by default
--hidden
--glob=!.git/*

# who cares about case!?
--smart-case



% rg remote
the_stow/.git/hooks/push-to-checkout.sample
12:# of the remote repository has any difference from the currently

the_stow/.git/hooks/pre-push.sample
4:# push" after it has checked the remote status, but before anything has been
9:# $1 -- Name of the remote to which the push is being done
12:# If pushing without using a named remote those arguments will be equal.
17:#   <local ref> <local oid> <remote ref> <remote oid>
22:remote="$1"
27:while read local_ref local_oid remote_ref remote_oid
34:		if test "$remote_oid" = "$zero"
40:			range="$remote_oid..$local_oid"

the_stow/.git/hooks/update.sample
23:#   This boolean sets whether remotely creating branches will be denied
110:	refs/remotes/*,commit)
113:	refs/remotes/*,delete)

the_stow/root/subdir/.stowed3
3:remotes



% vi $RIPGREP_CONFIG_PATH



% cat $RIPGREP_CONFIG_PATH
# Don't let ripgrep vomit really long lines to my terminal, and show a preview.
--max-columns=150
--max-columns-preview

# Search hidden files / directories (e.g. dotfiles) by default
--hidden
--glob=!.git/

# who cares about case!?
--smart-case



% rg remote               
the_stow/root/subdir/.stowed3
3:remotes
```

BurntSushi:
This example is just demonstrating the syntax of the config file.

The only thing that is said here about `--glob=!.git/*` or `--glob\n!.git/*` is the comment in the example itself:

```
# Using glob patterns to include/exclude files or folders 
```

Yet your characterization of what the example is showing is:

> i.e. include general hidden files and directories, except .git/ folder contents.

Which is not at all what is actually said.

Hence this change makes no sense.

PR #3141, opened 2025-09-07
Add Qt types for resource files and ui declaration

cmaureir:
qrc are the resource files for data related to user interfaces, and ui is the extension that the Qt Designer generates, for Widget based projects.


cmaureir:
Two references if needed:
* https://doc.qt.io/qt-6/resources.html
* https://doc.qt.io/qt-6/uic.html

BurntSushi:
Thanks! The `ui` name is way too general for something like this, so I've used `qui` instead.

cmaureir:
Cool, thanks!

PR #3142, opened 2025-09-08
Add support for GDScript files (*.gd) of Godot Engine

ttencate:


BurntSushi:
Thanks! Note that I've changed the name here from `gd` to `gdscript`. I'm not sure `gd` is recognizable enough.

PR #3145, opened 2025-09-12
globalset : Add literal_unclosed_class option to treat unclosed class as Literals

mostafa630:
## Changes
- Added `literal_unclosed_class: bool` to `GlobOptions` (defaults to `false`)
- Added `literal_unclosed_class()` method to `GlobBuilder`
- Modified `parse_class()` to save parser state and rollback on `UnclosedClass` errors when option is enabled
- Added comprehensive tests for patterns like `[abc`, `[]`, `[!]`, `[!`

## Behavior
When enabled, patterns that would throw `ErrorKind::UnclosedClass` are treated as literal characters instead:
- `[abc`  : matches literal `[abc`
- `[]`  : matches literal `[]`
- `[!]` :  matches literal `[!]`
- ` [! ` : matches literal `[!`

### DIFF
--- a/crates/globset/src/glob.rs
+++ b/crates/globset/src/glob.rs
@@ -206,6 +206,10 @@ struct GlobOptions {
     /// Whether or not an empty case in an alternate will be removed.
     /// e.g., when enabled, `{,a}` will match "" and "a".
     empty_alternates: bool,
+
+    /// Whether or not an unclosed character class will be treated as literals.
+    /// e.g., when enabled, `[abc` will be parsed as the literals "[abc".
+    literal_unclosed_class: bool,
 }
 
 impl GlobOptions {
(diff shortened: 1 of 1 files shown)


BurntSushi:
This needed a fair bit more work:

1. Your approach to parsing is quadratic. I fixed this by observing that if EOF is hit without finding a `]`, then there can never be another character class in the pattern. So when this happens, a toggle is set and prevents the parser from ever calling `parse_class` again.
2. This didn't wire the change up to `ignore`, which was a bit subtle. Namely, `Gitignore` now enables this by default but `Override` does not.
3. I added more tests.
4. I cleaned up the docs.
5. I renamed this to `allow_unclosed_class`.

mostafa630:
Thanks a lot for including this in the rollup,
and I really appreciate the detailed feedback you shared

PR #3147, opened 2025-09-13
cli: --iglob filter paths in argv

phanen:
`--{,i}glob` don't filter path in argv

can we add an option to support it or (this pr make it default now):
```sh
$ cargo run -- -e r LICENSE-MIT README.md --iglob '!*.md' --files
LICENSE-MIT
$ cargo run -- -e r LICENSE-MIT README.md --iglob '*.md' --files
README.md
```

For me paths in argv are built from fzf output, then I can use `--iglob` to filter the result in fzf live query again.

### DIFF
--- a/crates/ignore/src/dir.rs
+++ b/crates/ignore/src/dir.rs
@@ -362,7 +362,7 @@ impl Ignore {
     /// ignored or not.
     ///
     /// The match contains information about its origin.
-    fn matched<'a, P: AsRef<Path>>(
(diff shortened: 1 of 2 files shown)


BurntSushi:
No, this is an intentional design. It is very purposeful that if you pass a path to ripgrep, then ripgrep will always search it _regardless_ of any kind of filtering. To do otherwise would be madness. It is a useful property to be able to say, "if you give a file to ripgrep, it will get searched, no matter what."

This means that if you want filtering like this, it must be done before passing the file paths to ripgrep.

Issue #3115, opened 2025-07-31
The "copy" links in the README copy the command AND the leading $


lauralindzey:
First, thank you for including links to copy the commands! Not having to use the mouse to select text is a lot happier for my workflow.

However, when I paste the result, I get e.g. `$ brew install ripgrep`, which isn't a valid command :-)

BurntSushi:
Dupe of #2599, #2784, #2949

Issue #2949, opened 2024-12-17
dollar ('$') sign in readme installation code makes it harder to copy and past


nakul-krishnakumar:
### Please tick this box to confirm you have reviewed the above.

- [X] I have a different issue.

### What version of ripgrep are you using?

ripgrep 10.42

### How did you install ripgrep?

sudo apt-get install ripgrep

### What operating system are you using ripgrep on?

Ubuntu LTS 24.04

### Describe your bug.

Not a bug, but documentation error

### What are the steps to reproduce the behavior?

-

### What is the actual behavior?

when i copy paste the code to shell, $ comes with it, so I have to remove the dollar which according to me is a bad user experience

### What is the expected behavior?

direct copy paste should be possible

BurntSushi:
Duplicate of #2599, Duplicate of #2784

PR #3148, opened 2025-09-17
Make install commands works when pasted into terminal

hey-leon:
## Why The PR?
Github has a really nice feature where you can copy markdown code blocks to your clipboard. The README.md for this repo also has all the common commands specified to install this package depending on your environment. 

However, the commands include a prompt character. So when you copy and paste the commands they do not work without modification.

This PR aims to remedy this!


BurntSushi:
Dupe of #2599, #2784, #2949, #3115

hey-leon:
Yikes. My apologies @BurntSushi. I should have had a more thorough look before opening this!

PR #3152, opened 2025-09-21
cargo: set MSRV to 1.85

BurntSushi:
I believe the current stable version of Debian packages 1.85 rustc. So
if the next release of ripgrep uses a higher MSRV, then I think Debian
won't be able to package it.

It also turned out that I wasn't using anything from beyond Rust 1.85
anyway.

It's likely that I could make use of let-chains in various places, but I
don't think it's worth combing through the code to switch to them at
this point.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
         include:
         - build: pinned
           os: ubuntu-latest
-          rust: 1.88.0
+          rust: 1.85.0
         - build: stable
           os: ubuntu-latest
           rust: stable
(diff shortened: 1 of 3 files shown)


ltrzesniewski:
Quick question: why not add a `rust-toolchain.toml` file in order to avoid accidentally using the new features?

Is it to benefit from the newer compiler perf improvements for instance?

BurntSushi:
I always develop with ~latest nightly. I would never restrain my own local development environment just because Debian wants to lag behind by multiple years.

BurntSushi:
> Is it to benefit from the newer compiler perf improvements for instance?

But yes, this is one reason why I always use the latest nightly.

PR #3162, opened 2025-09-28
Implement min depth on `ignore`

### LINKED ISSUE #3158: Add `min_depth` to `ignore::WalkBuilder`

I am migrating from `walkdir` to `ignore` so I can support `.dockerignore` files, and we use the `walkdir::WalkDir::min_depth` method which has no equivalent in `ignore::WalkBuilder`.

I'd be willing to add the method.

Also, thanks for all the great crates!

AlvaroParker:
Closes #3158 

This replicates the behavior of `walkdir`, where passing `.min_depth(1)`, `.min_depth(0)` or no `min_depth` will have the same behavior.

Passing `min_depth(Some(2)` to the builder to a directory that looks like this:

```
.
├── a
│   ├── b
│   │   ├── c
│   │   │   └── foo
│   │   └── foo
│   └── foo
└── foo
```

Will returns something like this:

```
["a/b", "a/b/c", "a/b/c/foo", "a/b/foo", "a/foo"]
```

While passing a `min_depth` of `0`, `1` or `None`: 

```
["a", "a/b", "a/b/c", "foo", "a/foo", "a/b/foo", "a/b/c/foo"]
```

If the user passes a `max_depth` < `min_depth` then we set `max_depth = min_depth` as `walkdir` does.

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -484,6 +484,7 @@ pub struct WalkBuilder {
     paths: Vec<PathBuf>,
     ig_builder: IgnoreBuilder,
     max_depth: Option<usize>,
+    min_depth: Option<usize>,
     max_filesize: Option<u64>,
     follow_links: bool,
     same_file_system: bool,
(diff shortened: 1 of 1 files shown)


jalil-salame:
Thanks for doing this c:

Issue #1764, opened 2020-12-19
Issue with -g "!dirname/" when the CWD is "dirname"


thisch:
#### What version of ripgrep are you using?

```
ripgrep 12.1.1 (rev 7cb211378a)
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

#### How did you install ripgrep?

`sudo dpkg -i ripgrep_12*.deb `

#### What operating system are you using ripgrep on?

`Ubuntu 19.10`

#### Describe your bug.

If a glob pattern and a root directory is passed to `rg`, the ignored files depend on the current working directory.

#### What are the steps to reproduce the behavior?

```bash
#!/bin/bash
set -x

rm -rf /tmp/foo

mkdir -p /tmp/foo/to-be-ignored
mkdir /tmp/foo/dir1

touch /tmp/foo/to-be-ignored/a
touch /tmp/foo/dir1/d

#### setup done ####

cd /tmp/foo/to-be-ignored
# no output expected:  BUG
rg --files -g '!to-be-ignored/' /tmp/foo | grep to-be-ignored

cd /tmp/foo/dir1
# no output expected:  OK
rg --files -g '!to-be-ignored/' /tmp/foo | grep to-be-ignored

cd /tmp/foo
# no output expected:  OK
rg --files -g '!to-be-ignored/' /tmp/foo | grep to-be-ignored


cd /tmp/foo/to-be-ignored
# no output expected:  OK
rg --files -g '!dir1/' /tmp/foo | grep dir1

cd /tmp/foo/dir1
# no output expected:  BUG
rg --files -g '!dir1/' /tmp/foo | grep dir1

cd /tmp/foo
# no output expected:  OK
rg --files -g '!dir1/' /tmp/foo | grep dir1
```

#### What is the actual behavior?

```
+ mkdir -p /tmp/foo/to-be-ignored
+ mkdir /tmp/foo/dir1
+ touch /tmp/foo/to-be-ignored/a
+ touch /tmp/foo/dir1/d
+ cd /tmp/foo/to-be-ignored
+ rg --files -g '!to-be-ignored/' /tmp/foo
+ grep to-be-ignored
/tmp/foo/to-be-ignored/a
+ cd /tmp/foo/dir1
+ rg --files -g '!to-be-ignored/' /tmp/foo
+ grep to-be-ignored
+ cd /tmp/foo
+ rg --files -g '!to-be-ignored/' /tmp/foo
+ grep to-be-ignored
[...]

PR #3163, opened 2025-10-03
#1764

shubhamDP-arch:
fixed issue at #1764  i am new to this it got fixed when i ran it on my machine if valid please merge .

### DIFF
--- a/crates/ignore/src/gitignore.rs
+++ b/crates/ignore/src/gitignore.rs
@@ -8,10 +8,7 @@ the `git` command line tool.
 */
 
 use std::{
-    fs::File,
-    io::{BufRead, BufReader, Read},
-    path::{Path, PathBuf},
-    sync::Arc,
+    env, fs::File, io::{BufRead, BufReader, Read}, path::{Path, PathBuf}, sync::Arc
 };
 
 use {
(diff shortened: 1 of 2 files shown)


crates/ignore/src/gitignore.rs
```diff
@@ -250,7 +251,12 @@ impl Gitignore {
         if self.is_empty() {
             return Match::None;
         }
-        let path = path.as_ref();
+        let path_buf: PathBuf = if path.as_ref().to_str().unwrap().is_empty() {
+            env::current_dir().unwrap()
+        } else {
+            path.as_ref().to_path_buf()
```
BurntSushi, line 257: This is doing an alloc and a copy just to make the types unify. And this is called for _every_ match. Absolutely cannot do that.

crates/ignore/src/gitignore.rs
```diff
@@ -250,7 +251,12 @@ impl Gitignore {
         if self.is_empty() {
             return Match::None;
         }
-        let path = path.as_ref();
+        let path_buf: PathBuf = if path.as_ref().to_str().unwrap().is_empty() {
+            env::current_dir().unwrap()
```
BurntSushi, line 255: This doesn't make any sense to me.

crates/ignore/src/gitignore.rs
```diff
@@ -250,7 +251,12 @@ impl Gitignore {
         if self.is_empty() {
             return Match::None;
         }
-        let path = path.as_ref();
+        let path_buf: PathBuf = if path.as_ref().to_str().unwrap().is_empty() {
```
BurntSushi, line 254: Doing a UTF-8 check just to see if the path is empty doesn't make sense.

PR #3186, opened 2025-10-14
ignore: add lowercase r r_extensions

luhann:
Add lowercase versions of *.Rmd and *.Rnw to the R type. 

These extensions are less common, but useful given that the *.r extension is also added as the R type.

Thanks for all the work you do on rg, if this isn't wanted I'm happy to keep using my custom r type.

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -219,7 +219,7 @@ pub(crate) const DEFAULT_TYPES: &[(&[&str], &[&str])] = &[
     (&["qml"], &["*.qml"]),
     (&["qrc"], &["*.qrc"]),
     (&["qui"], &["*.ui"]),
-    (&["r"], &["*.R", "*.r", "*.Rmd", "*.Rnw"]),
+    (&["r"], &["*.R", "*.r", "*.Rmd", "*.rmd", "*.Rnw", "*.rnw"]),
     (&["racket"], &["*.rkt"]),
     (&["raku"], &[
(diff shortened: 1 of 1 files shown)


BurntSushi:
Can you show me some code/repo that is using the `rmd` and `rnw` extensions?

luhann:
Sure, I use [R.nvim](https://github.com/R-nvim/R.nvim) for my development and it recognises both *.rmd and *.rnw as R files. [Knitr](https://github.com/yihui/knitr) also recognises both *.rmd and *.rnw as R files, although the actual files in the repo are .R and .Rnw/.Rmd due to tidyverse style guides. [Rmarkdown](https://github.com/rstudio/rmarkdown/blob/d01420eea96019cbb544f12eefd8a80e0c458114/NEWS.md?plain=1#L160) will also recognise .rmd as valid.

BurntSushi:
I meant code repositories that _contain_ `rmd` and `rnw` files. But I suppose other tools recognizing those extensions is a fine approximation too.

luhann:
I don't mind continuing to take a look, but most R based repositories follow the tidyverse style guide which calls for .R, .Rnw, and .Rmd. Even finding examples for .r is challenging. 

I also did some checking for other file types that could possibly be confused for *.rmd and *.rnw, just so this change doesn't cause unexpected behaviour for anyone and I couldn't find any.

Thanks again for all the work you do.

BurntSushi:
Thanks!

Issue #3198, opened 2025-10-21
manpage mistake regarding word boundary flag?


noughtnaut:
### Please tick this box to confirm you have reviewed the above.

- [x] I have a different issue.

### What version of ripgrep are you using?

The man page states:

       -w, --word-regexp
           When enabled, ripgrep will only show matches surrounded by word boundaries.  This is equiv‐
           alent to surrounding every pattern with \b{start-half} and \b{end-half}.

but should the last bit not be `{end-half}\b` to indicate the pattern _ends_ at a word boundary?

### How did you install ripgrep?

apt

### What operating system are you using ripgrep on?

WSL on Windows 11

### Describe your bug.

Manpage seems inaccurate. The program works fine.

### What are the steps to reproduce the behavior?

`man rg | grep -B 3 end-half`

### What is the actual behavior?

man page is shown but perhaps misleading

### What is the expected behavior?

The man page should state:

       -w, --word-regexp
           When enabled, ripgrep will only show matches surrounded by word boundaries.  This is equiv‐
           alent to surrounding every pattern with \b{start-half} and {end-half}\b.

BurntSushi:
`{end-half}\b` isn't a valid pattern. Why not try it?

The man page is correct.

PR #3200, opened 2025-10-22
Fix mistake in man page for -w/--word-regexp

KlaymenDK:
This is a fix for the documentation for the `-w` flag. The feature itself works as intended.

The output of `man rg | grep -B 3 end-half` states that:

       -w, --word-regexp
           When  enabled,  ripgrep will only show matches surrounded by word boundaries.  This is equiva‐
           lent to surrounding every pattern with \b{start-half} and \b{end-half}.

in which the last bit seems incorrect -- `\b{end-half}` should be `{end-half}\b` meaning the END of the pattern match aligns with a word boundary. This seems like a typo or simple mistake.


BurntSushi:
Duplicate of #3198.

`{end-half}\b` isn't a valid pattern. Why would you change a valid pattern to an invalid one?

PR #3207, opened 2025-10-23
ci: support aarch64-musl statically-linked binary

chrisdone-artificial:
Curious whether this will generate the arm64 static artifact, as I need this.


chrisdone-artificial:
Nope, it did not. Nevermind!

PR #3212, opened 2025-10-30
ignore: only stat .jj if we actually care

ianloic:
I was comparing the work being done by fd and find and noticed (with strace -f -c -S calls) that fd was doing a ton of failed statx calls. Upon closer inspection it was stating .jj even though I was passing --no-ignore. Eventually I turned up this check in Ignore::add_child_path that was doing stat on .jj regardless of whether the options request it.

With this patch it'll only stat .jj if that's relevant to the query.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@ TBD
 ===
 Unreleased changes. Release notes have not yet been written.
 
+Bug fixes:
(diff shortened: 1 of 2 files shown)


BurntSushi:
This PR is on crates.io in `ignore 0.4.25`.

Issue #3214, opened 2025-11-01
Feature: please add support for -mtime and -ctime similar to those from find


adrimt:
RIPGREP is already a very useful tool for replacing find and grep, however I encountered some cases when mtime or ctime options where needed.
Please add support for -mtime, -ctime or some time related options similar to those on find.

BurntSushi:
I've been thinking about doing this lately. We can use `jiff` to accept pretty fleixlbe inputs. IDK if I'll call them `-mtime` or `-ctime` though.

PR #3220, opened 2025-11-16
[FEATURE] Added --mtime flag and logic

gege42o:
## Feature Description

The `--mtime` flag allows users to filter files based on their modification time before searching them. 
### Syntax

The flag accepts three formats, matching the behavior of `find -mtime`:

| Format | Description | Example |
|--------|-------------|---------|
| `--mtime N` | Files modified exactly N days ago | `--mtime 3` |
| `--mtime -N` | Files modified less than N days ago (within last N days) | `--mtime -7` |
| `--mtime +N` | Files modified more than N days ago (older than N days) | `--mtime +30` |

### Behavior Characteristics

1. **Fractional Day Handling**: When calculating days ago, fractional parts are **truncated** (not rounded)
   - A file modified 1.9 days ago is considered "1 day old"
   - `--mtime +1` requires at least 2 full days

2. **24-Hour Periods**: Each day is exactly 24×60×60 = 86,400 seconds

3. **Filter-Only**: This flag only affects which files are searched, not the output format

4. **Metadata Required**: Files without accessible metadata are skipped with debug logging

---

### Data Flow

1. **Parsing Phase**: User provides `--mtime VALUE`
   - Value is parsed to determine format (`+`, `-`, or plain number)
   - Converted to `MtimeFilter` enum variant
   - Stored in `LowArgs.mtime`

2. **Configuration Phase**: Arguments converted from low to high level
   - `LowArgs.mtime` → `HiArgs.mtime`
   - No transformation, direct copy

3. **Execution Phase**: Files are filtered during traversal
   - `HiArgs.mtime` → `HaystackBuilder.mtime_filter`
   - For each file encountered:
     - Get file metadata
     - Extract modification time
     - Compare against filter
     - Include/exclude file

### DIFF
--- a/crates/core/flags/complete/rg.zsh
+++ b/crates/core/flags/complete/rg.zsh
@@ -216,6 +216,9 @@ _rg() {
     '(--no-multiline)--multiline-dotall[allow "." to match newline (with -U)]'
     $no"(--no-multiline)--no-multiline-dotall[don't allow \".\" to match newline (with -U)]"
 
+    + '(mtime)' # Mtime options
+    '--mtime+[filter files by modification time]:days (N, -N, or +N)'
+
     + '(only)' # Only-match options
     {-o,--only-matching}'[show only matching part of each line]'
 


gege42o:
It appears that the feature works and is not flagged by rust-fmt. The errors are related to the help message but I don't know exactly where it is wrong. A little help would be appreciated and if needed I will provide further explanations about the code, idea and execution.

BurntSushi:
If someone other than me is going to add this to ripgrep, then it needs design first before going to a PR. See #3214.

There are a number of issues with the approach here. Like being limited to days. Or always assuming 1 day is 24 hours (some are longer, some are shorter). And I'm not a fan of the name "mtime."

gege42o:
I understand, I saw the same issue before i started designing this. Wanted to do a 1:1 copy to find's --mtime that works only for days because thats what the person who wrote the original issue wanted. What names would be suitable for this option? Maybe I can try again while using jiff.

PR #3224, opened 2025-11-20
Separate flags from paths in decompression tool invocations

jafd:
This fixes the bug when the filename of a compressed file starts with the `-` character and it gets passed to the decompression tool, but gets treated as an option instead of file.

In the best case, the option would be invalid and the decompression tool would finish with an error. In the worst case, the filename that looks like a valid option could potentially overwrite or remove data as a side effect.

All decompression tools in use here follow the convention that everything after `--` is treated as a filename.

### DIFF
--- a/crates/cli/src/decompress.rs
+++ b/crates/cli/src/decompress.rs
@@ -488,14 +488,14 @@ fn try_resolve_binary<P: AsRef<Path>>(
 }
 
 fn default_decompression_commands() -> Vec<DecompressionCommand> {
-    const ARGS_GZIP: &[&str] = &["gzip", "-d", "-c"];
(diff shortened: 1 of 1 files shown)


jafd:
The reason I'm adding the `--` for each tool explicitly is so that it all still works when some other tool appears that doesn't follow this convention and says "Unknown option -", or expects some nonstandard flag like `-input`. If you just add `--` before the path, you're making an assumption that may or may not hold true.

blyxxyz:
Another option is to prefix relative paths with `./` so that `gzip` sees `./-10.txt`. That'd also help `--pre` (though it could thereby be more disruptive).

PR #3239, opened 2025-12-06
Add --suppress-error flag for permission-denied handling

M1tsumi:
This PR introduces a new `--suppress-error` flag that lets users control how ripgrep handles permission-denied errors during file traversal and searching.

The new flag has two modes:
- `--suppress-error=permission` - suppresses the error output but still exits with code 2
- `--suppress-error=nopermission` - suppresses output AND treats these files as skipped (exit code unaffected)

## Changes

**Flag and enum** (`lowargs.rs`, `defs.rs`, `hiargs.rs`)
- Added `SuppressErrorMode` enum with `PermissionSilence` and `PermissionSkip` variants
- Hooked it up through the low-args → hi-args pipeline

- Added `handle_search_error()` helper used by both single-threaded and parallel search
- Checks error kind against suppression mode to decide whether to print/set error flag

**Tests**
- Flag parsing tests in `defs.rs`
- Unit tests for the error handler covering mode × error type combinations
- Existing tests pass

This leaves the Default behavior unchanged - new behavior only kicks in if you pass the flag.

### DIFF
--- a/crates/core/flags/complete/rg.zsh
+++ b/crates/core/flags/complete/rg.zsh
@@ -199,6 +199,7 @@ _rg() {
     + '(messages)' # Error-message options
     '(--no-ignore-messages)--no-messages[suppress some error messages]'
     $no"--messages[don't suppress error messages affected by --no-messages]"
+    '--suppress-error=[control how permission-denied errors affect messages and exit status]:kind'
 
     + '(messages-ignore)' # Ignore-error message options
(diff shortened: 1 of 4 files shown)


BurntSushi:
ripgrep already has a `--no-messages` flag.

ltrzesniewski:
Wow, you even left the prompt in the PR /s

Issue #36, opened 2016-09-23
don't use memory maps ever on mac


BurntSushi:
See this thread for justification: https://news.ycombinator.com/item?id=12567326

We already do custom memory map logic for Windows, so it shoul dbe no problem to do it for Mac too.

PR #3246, opened 2025-12-15
perf(searcher/mmap): advise sequential access for file-backed mmaps on Unix

howmanysmall:
This applies `memmap2::Mmap::advise(Advice::Sequential)` after mapping a file on Unix. This is particularly important on macOS where mmap performance has historically been inconsistent (see #36). We also log `madvise` failures at debug level (including the path when available).

## Benchmark (macOS, 4GB file)
Command (as-run):

```sh
hyperfine --warmup 5 \
  'rg -a "foo" test_large_file || true' \
  './target/release/rg -a "foo" test_large_file --no-mmap || true' \
  './target/release/rg -a "foo" test_large_file --mmap || true'
```

Results:

- released `rg` (PATH), default behavior: 2.569 s ± 0.529 s
- local `./target/release/rg --no-mmap`: 2.548 s ± 0.315 s
- local `./target/release/rg --mmap`: 530.1 ms ± 362.8 ms

(So local `--mmap` is ~4.8× faster than local `--no-mmap` on this workload.)

## Safety / escape hatch
File-backed mmaps can SIGBUS if the underlying file is truncated while being read. Users can opt out with `--no-mmap` (already documented in the CAVEATS).

## Testing
- cargo fmt --all --check
- cargo test --workspace --features pcre2
- ci/test-complete

### DIFF
--- a/crates/searcher/src/searcher/mmap.rs
+++ b/crates/searcher/src/searcher/mmap.rs
@@ -76,7 +76,11 @@ impl MmapChoice {
         // assertion that using memory maps is safe.
         match unsafe { Mmap::map(file) } {
             Ok(mmap) => {
-                #[cfg(target_os = "macos")]
+                // Hint to the kernel that we'll read sequentially. This is
+                // particularly important on macOS where mmap was historically
+                // ~2x slower than regular reads without this hint.
+                // See: https://github.com/BurntSushi/ripgrep/issues/36
+                #[cfg(unix)]
                 if let Err(err) = mmap.advise(memmap::Advice::Sequential) {
                     if let Some(path) = path {
                         log::debug!(
--- a/crates/searcher/src/searcher/mmap.rs
+++ b/crates/searcher/src/searcher/mmap.rs
@@ -70,16 +70,16 @@ impl MmapChoice {
         if !self.is_enabled() {
             return None;
         }
-        if cfg!(target_os = "macos") {
-            // I guess memory maps on macOS aren't great. Should re-evaluate.
-            return None;
-        }
         // SAFETY: This is acceptable because the only way `MmapChoiceImpl` can
         // be `Auto` is if the caller invoked the `auto` constructor, which
         // is itself not safe. Thus, this is a propagation of the caller's
         // assertion that using memory maps is safe.
         match unsafe { Mmap::map(file) } {
-            Ok(mmap) => Some(mmap),
+            Ok(mmap) => {
+                #[cfg(target_os = "macos")]
+                let _ = mmap.advise(memmap::Advice::Sequential);
+                Some(mmap)
+            }
             Err(err) => {
                 if let Some(path) = path {
                     log::debug!(
--- a/crates/searcher/src/searcher/mmap.rs
+++ b/crates/searcher/src/searcher/mmap.rs
@@ -77,7 +77,17 @@ impl MmapChoice {
         match unsafe { Mmap::map(file) } {
             Ok(mmap) => {
                 #[cfg(target_os = "macos")]
-                let _ = mmap.advise(memmap::Advice::Sequential);
+                if let Err(err) = mmap.advise(memmap::Advice::Sequential) {
+                    if let Some(path) = path {
+                        log::debug!(
+                            "{}: madvise failed: {}",
+                            path.display(),
+                            err
+                        );
+                    } else {
+                        log::debug!("madvise failed: {}", err);
+                    }
+                }
                 Some(mmap)
             }
             Err(err) => {


howmanysmall:
Should add some benchmarks as well. These were ran on my M4 Pro (14 core).

## 1 GB File

```
>mkfile 1g test_large_file
>hyperfine --warmup 5 'rg -a "foo" test_large_file || true' './target/release/rg -a "foo" test_large_file --no-mmap || true' './target/release/rg -a "foo" test_large_file --mmap || true' 'grep -a "foo" test_large_file || true'
Benchmark 1: rg -a "foo" test_large_file || true
  Time (mean ± σ):     195.7 ms ±  11.2 ms    [User: 53.9 ms, System: 136.8 ms]
  Range (min … max):   186.6 ms … 221.5 ms    13 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 2: ./target/release/rg -a "foo" test_large_file --no-mmap || true
  Time (mean ± σ):     195.6 ms ±  13.6 ms    [User: 55.0 ms, System: 136.6 ms]
  Range (min … max):   187.1 ms … 235.7 ms    15 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 3: ./target/release/rg -a "foo" test_large_file --mmap || true
  Time (mean ± σ):      90.6 ms ±   2.8 ms    [User: 56.9 ms, System: 32.9 ms]
  Range (min … max):    85.9 ms …  97.1 ms    31 runs

Benchmark 4: grep -a "foo" test_large_file || true
  Time (mean ± σ):     216.4 ms ±  11.9 ms    [User: 70.6 ms, System: 138.2 ms]
  Range (min … max):   201.1 ms … 235.5 ms    14 runs

Summary
  ./target/release/rg -a "foo" test_large_file --mmap || true ran
    2.16 ± 0.16 times faster than ./target/release/rg -a "foo" test_large_file --no-mmap || true
    2.16 ± 0.14 times faster than rg -a "foo" test_large_file || true
    2.39 ± 0.15 times faster than grep -a "foo" test_large_file || true
```

## 4 GB File

```
>mkfile 4g test_large_file
>hyperfine --warmup 5 'rg -a "foo" test_large_file || true' './target/release/rg -a "foo" test_large_file --no-mmap || true' './target/release/rg -a "foo" test_large_file --mmap || true' 'grep -a "foo" test_large_file || true'
Benchmark 1: rg -a "foo" test_large_file || true
  Time (mean ± σ):      3.291 s ±  0.678 s    [User: 0.425 s, System: 1.728 s]
  Range (min … max):    2.559 s …  4.702 s    10 runs

Benchmark 2: ./target/release/rg -a "foo" test_large_file --no-mmap || true
  Time (mean ± σ):      2.749 s ±  0.321 s    [User: 0.422 s, System: 1.476 s]
  Range (min … max):    2.280 s …  3.412 s    10 runs

Benchmark 3: ./target/release/rg -a "foo" test_large_file --mmap || true
  Time (mean ± σ):     412.7 ms ±  67.4 ms    [User: 222.7 ms, System: 169.0 ms]
  Range (min … max):   381.7 ms … 602.9 ms    10 runs

  Warning: Statistical outliers were detected. Consider re-running this benchmark on a quiet system without any interferences from other programs. It might help to use the '--warmup' or '--prepare' options.

Benchmark 4: grep -a "foo" test_large_file || true
  Time (mean ± σ):      1.240 s ±  0.091 s    [User: 0.187 s, System: 0.812 s]
  Range (min … max):    1.105 s …  1.408 s    10 runs

Summary
  ./target/release/rg -a "foo" test_large_file --mmap || true ran
    3.00 ± 0.54 times faster than grep -a "foo" test_large_file || true
    6.66 ± 1.34 times faster than ./target/release/rg -a "foo" test_large_file --no-mmap || true
    7.97 ± 2.10 times faster than rg -a "foo" test_large_file || true
```

BurntSushi:
Thanks! I'll take a look at this soon. I'll need to do my own benchmarks though. The benchmark you have here is quite pathological. It's just 4GB of NUL bytes. Is this something that you search realistically?

howmanysmall:
No, it was really just the easiest thing to benchmark. I should have probably done something more comprehensive but I couldn't be bothered.

BurntSushi:
OK, so I did some of my own benchmarking on a "normal" plain text file. And unfortunately, I don't see that much has changed. There _is_ a case where using memory maps on macOS (for this input) is slightly faster than not, but there is also a very significant regression once you start searching a larger file.

OK, first is acquiring the corpora. There are two files. The first is a smaller
sample of the second. The first is 278MB compressed. The second is 3.5GB compressed.

```
curl -LO https://burntsushi.net/stuff/OpenSubtitles2018.raw.sample.en.gz
curl -LO https://burntsushi.net/stuff/OpenSubtitles2018.raw.en.gz
gzip -d OpenSubtitles2018.raw.sample.en.gz
gzip -d OpenSubtitles2018.raw.en.gz
```

This first benchmark is on Linux that uses a pattern that never matches. We compare three cases: no memory maps, with memory maps and advice and with memory maps and no advice:

```
$ hyperfine --output pipe -i -w1 -m3 -M10 'rg-master --no-mmap ZQZQZQ OpenSubtitles2018.raw.sample.en' 'rg-mmap-advice ZQZQZQ OpenSubtitles2018.raw.sample.en' 'rg-mmap-noadvice ZQZQZQ OpenSubtitles2018.raw.sample.en'

Benchmark 1: rg-master --no-mmap ZQZQZQ OpenSubtitles2018.raw.sample.en
  Time (mean ± σ):      87.5 ms ±   1.4 ms    [User: 31.4 ms, System: 55.8 ms]
  Range (min … max):    85.1 ms …  89.2 ms    10 runs

  Warning: Ignoring non-zero exit code.

Benchmark 2: rg-mmap-advice ZQZQZQ OpenSubtitles2018.raw.sample.en
  Time (mean ± σ):      59.5 ms ±   2.7 ms    [User: 47.1 ms, System: 12.1 ms]
  Range (min … max):    55.5 ms …  65.5 ms    10 runs

  Warning: Ignoring non-zero exit code.

Benchmark 3: rg-mmap-noadvice ZQZQZQ OpenSubtitles2018.raw.sample.en
  Time (mean ± σ):      59.2 ms ±   2.6 ms    [User: 45.0 ms, System: 13.9 ms]
  Range (min … max):    54.9 ms …  62.4 ms    10 runs

  Warning: Ignoring non-zero exit code.

Summary
  rg-mmap-noadvice ZQZQZQ OpenSubtitles2018.raw.sample.en ran
    1.01 ± 0.06 times faster than rg-mmap-advice ZQZQZQ OpenSubtitles2018.raw.sample.en
    1.48 ± 0.07 times faster than rg-master --no-mmap ZQZQZQ OpenSubtitles2018.raw.sample.en
```

That was on the smaller "sample" file. Now let's do it again on the larger file:

```
$ hyperfine --output pipe -i -w1 -m3 -M3 'rg-master --no-mmap ZQZQZQ full.txt' 'rg-mmap-advice ZQZQZQ full.txt' 'rg-mmap-noadvice ZQZQZQ full.txt'
Benchmark 1: rg-master --no-mmap ZQZQZQ full.txt
  Time (mean ± σ):      1.215 s ±  0.110 s    [User: 0.380 s, System: 0.833 s]
  Range (min … max):    1.142 s …  1.342 s    3 runs

  Warning: Ignoring non-zero exit code.

Benchmark 2: rg-mmap-advice ZQZQZQ full.txt
  Time (mean ± σ):     803.6 ms ±  12.0 ms    [User: 589.3 ms, System: 212.4 ms]
  Range (min … max):   792.9 ms … 816.6 ms    3 runs

  Warning: Ignoring non-zero exit code.

Benchmark 3: rg-mmap-noadvice ZQZQZQ full.txt
  Time (mean ± σ):     798.9 ms ±  40.5 ms    [User: 592.7 ms, System: 204.2 ms]
  Range (min … max):   772.1 ms … 845.5 ms    3 runs

  Warning: Ignoring non-zero exit code.

Summary
  rg-mmap-noadvice ZQZQZQ full.txt ran
    1.01 ± 0.05 times faster than rg-mmap-advice ZQZQZQ full.txt
    1.52 ± 0.16 times faster than rg-master --no-mmap ZQZQZQ full.txt

```

This is consistent with my historical observation of mmap performance on Linux:

1. The sequential advice doesn't matter.
2. Using memory maps gives a modest improvement over not using memory maps.

Now let's do exactly the same benchmark, but on macOS. First is the smaller file:

```
$ hyperfine --output pipe -i -w1 -m3 -M10 'rg-master ZQZQZQ OpenSubtitles2018.raw.sample.en' 'rg-mmap-advice ZQZQZQ OpenSubtitles2018.raw.sample.en' 'rg-mmap-noadvice ZQZQZQ OpenSubtitles2018.raw.sample.en'
Benchmark 1: rg-master ZQZQZQ OpenSubtitles2018.raw.sample.en
  Time (mean ± σ):     107.0 ms ±   0.1 ms    [User: 42.0 ms, System: 64.6 ms]
  Range (min … max):   106.8 ms … 107.1 ms    10 runs

  Warning: Ignoring non-zero exit code.

Benchmark 2: rg-mmap-advice ZQZQZQ OpenSubtitles2018.raw.sample.en
  Time (mean ± σ):      86.4 ms ±   0.3 ms    [User: 54.0 ms, System: 32.0 ms]
  Range (min … max):    86.1 ms …  86.9 ms    10 runs

  Warning: Ignoring non-zero exit code.

Benchmark 3: rg-mmap-noadvice ZQZQZQ OpenSubtitles2018.raw.sample.en
  Time (mean ± σ):      86.3 ms ±   0.1 ms    [User: 54.0 ms, System: 31.9 ms]
  Range (min … max):    86.0 ms …  86.5 ms    10 runs

  Warning: Ignoring non-zero exit code.

Summary
  rg-mmap-noadvice ZQZQZQ OpenSubtitles2018.raw.sample.en ran
    1.00 ± 0.00 times faster than rg-mmap-advice ZQZQZQ OpenSubtitles2018.raw.sample.en
    1.24 ± 0.00 times faster than rg-master ZQZQZQ OpenSubtitles2018.raw.sample.en
```

OK, this is promising! There's no difference between memory maps with and without sequential advice, but there is a modest (even more modest) improvement over not using memory maps.

Now let's try with the bigger file:

```
$ hyperfine --output pipe -i -w1 -m3 -M3 'rg-master ZQZQZQ OpenSubtitles2018.raw.en' 'rg-mmap-advice ZQZQZQ OpenSubtitles2018.raw.en' 'rg-mmap-noadvice ZQZQZQ OpenSubtitles2018.raw.en'
Benchmark 1: rg-master ZQZQZQ OpenSubtitles2018.raw.en
  Time (mean ± σ):      7.497 s ±  0.006 s    [User: 0.649 s, System: 1.380 s]
  Range (min … max):    7.492 s …  7.503 s    3 runs

  Warning: Ignoring non-zero exit code.

Benchmark 2: rg-mmap-advice ZQZQZQ OpenSubtitles2018.raw.en
  Time (mean ± σ):     26.701 s ±  0.061 s    [User: 1.495 s, System: 2.147 s]
  Range (min … max):   26.639 s … 26.762 s    3 runs

  Warning: Ignoring non-zero exit code.

Benchmark 3: rg-mmap-noadvice ZQZQZQ OpenSubtitles2018.raw.en
  Time (mean ± σ):     31.401 s ±  0.070 s    [User: 1.611 s, System: 4.829 s]
  Range (min … max):   31.345 s … 31.479 s    3 runs

  Warning: Ignoring non-zero exit code.

Summary
  rg-master ZQZQZQ OpenSubtitles2018.raw.en ran
    3.56 ± 0.01 times faster than rg-mmap-advice ZQZQZQ OpenSubtitles2018.raw.en
    4.19 ± 0.01 times faster than rg-mmap-noadvice ZQZQZQ OpenSubtitles2018.raw.en
```

Wat. This to me says that memory maps on macOS still suck. The sequential advice *does* seem to matter here, but regardless of setting the advice, using memory maps is still significantly slower than not using them.

Now, let's go back to your benchmark. Let's start by confirming that I can roughly reproduce your result:

```
$ hyperfine -i --output pipe -w1 'rg-master -a "foo" test_large_file' 'rg-mmap-advice -a "foo" test_large_file' 'rg-mmap-noadvice -a "foo" test_large_file' 'grep -a "foo" test_large_file'
Benchmark 1: rg-master -a "foo" test_large_file
  Time (mean ± σ):      2.461 s ±  0.098 s    [User: 0.449 s, System: 1.223 s]
  Range (min … max):    2.391 s …  2.644 s    10 runs

  Warning: Ignoring non-zero exit code.

Benchmark 2: rg-mmap-advice -a "foo" test_large_file
  Time (mean ± σ):     418.9 ms ±   1.8 ms    [User: 263.7 ms, System: 154.7 ms]
  Range (min … max):   416.5 ms … 423.1 ms    10 runs

  Warning: Ignoring non-zero exit code.

Benchmark 3: rg-mmap-noadvice -a "foo" test_large_file
  Time (mean ± σ):     799.6 ms ±   3.8 ms    [User: 239.8 ms, System: 559.4 ms]
  Range (min … max):   797.0 ms … 810.1 ms    10 runs

  Warning: Ignoring non-zero exit code.

Benchmark 4: grep -a "foo" test_large_file
  Time (mean ± σ):     685.1 ms ±   8.1 ms    [User: 217.0 ms, System: 465.8 ms]
  Range (min … max):   677.2 ms … 704.9 ms    10 runs

  Warning: Ignoring non-zero exit code.

Summary
  rg-mmap-advice -a "foo" test_large_file ran
    1.64 ± 0.02 times faster than grep -a "foo" test_large_file
    1.91 ± 0.01 times faster than rg-mmap-noadvice -a "foo" test_large_file
    5.87 ± 0.24 times faster than rg-master -a "foo" test_large_file
```

That looks roughly consistent with what you observe: mmap is significantly better than non-mmap, and sequential advice gives another boost.

So... does file size matter here? If I create a 12GB file of all NUL bytes, will the mmap approach be slower? I just did an ad hoc test here since I didn't feel like waiting for `hyperfine` given the times:

```
$ time rg-master -a foo test_large_file_x3

real    18.033
user    2.469
sys     7.605
maxmem  7920528 MB
faults  14
$ time rg-master -a foo test_large_file_x3

real    18.395
user    2.487
sys     7.859
maxmem  8068928 MB
faults  177

$ time rg-mmap-advice -a foo test_large_file_x3

real    25.124
user    1.581
sys     2.033
maxmem  11026528 MB
faults  786645
$ time rg-mmap-advice -a foo test_large_file_x3

real    25.250
user    1.595
sys     2.198
maxmem  11002368 MB
faults  770135
$ time rg-mmap-advice -a foo test_large_file_x3

real    24.603
user    1.580
sys     2.273
maxmem  10994512 MB
faults  729232
```

... so... yeah... file size matters?

What about testing the original 4GB file on Linux? (I'm not going to bother with the 12GB file because I have no suspicion that Linux will crap out with larger files like macOS does.)

```
$ hyperfine --output pipe -i -w1 -m3 -M3 'rg-master --no-mmap -a foo /tmp/test_large_file' 'rg-mmap-advice -a foo /tmp/test_large_file' 'rg-mmap-noadvice -a foo /tmp/test_large_file'
Benchmark 1: rg-master --no-mmap -a foo /tmp/test_large_file
  Time (mean ± σ):      2.570 s ±  0.032 s    [User: 0.659 s, System: 1.906 s]
  Range (min … max):    2.550 s …  2.606 s    3 runs

  Warning: Ignoring non-zero exit code.

Benchmark 2: rg-mmap-advice -a foo /tmp/test_large_file
  Time (mean ± σ):     375.7 ms ±   5.1 ms    [User: 210.0 ms, System: 164.8 ms]
  Range (min … max):   371.4 ms … 381.4 ms    3 runs

  Warning: Ignoring non-zero exit code.

Benchmark 3: rg-mmap-noadvice -a foo /tmp/test_large_file
  Time (mean ± σ):     373.8 ms ±   2.6 ms    [User: 206.2 ms, System: 166.7 ms]
  Range (min … max):   370.8 ms … 375.8 ms    3 runs

  Warning: Ignoring non-zero exit code.

Summary
  rg-mmap-noadvice -a foo /tmp/test_large_file ran
    1.01 ± 0.02 times faster than rg-mmap-advice -a foo /tmp/test_large_file
    6.88 ± 0.10 times faster than rg-master --no-mmap -a foo /tmp/test_large_file
```

Yeah so both the mmap searches are faster, but now _substantially_ faster than the approach without memory mapping. So I think there is more going on here than just memory mapping. In particular, if we get the peak memory usage of the no-mmap search:

```
$ time rg-master --no-mmap -a foo /tmp/test_large_file

real    2.704
user    0.831
sys     1.867
maxmem  11077 MB
faults  0
```

That is quite a lot. It turns out that searching a file with all NUL bytes is pretty pathological. And when you use the `-a/--text` flag, you're specifically telling ripgrep to treat binary data as if it were text. But if you omit the `-a/--text` flag:

```
$ time rg-master --no-mmap foo /tmp/test_large_file

real    1.400
user    1.020
sys     0.376
maxmem  30 MB
faults  0
```

Which is still slower than the mmap searches, but faster than with `-a/--text` _and_ memory usage is far more reasonable.

What's going on here is that when you use `-a/--text` in this example, ripgrep searches by chunking up the input into lines. And the underlying regex engine(s) have a requirement that their haystack fit into contiguous memory. So this means every line has to be read onto the heap. But in this case, the file is just 4GB of NUL bytes. It's only one line. So the entire file gets read onto the heap as a single line. That's slow and causes higher memory usage.

But what about when `-a/--text` is _not_ used? In that case, ripgrep uses a heuristic (adopted from GNU grep) that "zaps" all of the NUL bytes in the haystack into line terminators. This heuristic is meant to avoid the "read GB of data onto the heap" by introducing "fake" lines where there previously were none.

Both of these cases, in other words, are doing extra work due to the pathological nature of the input. And this is where the mmap option really helps. Here, it isn't _just_ that mmap is itself faster than file I/O (at least, it is on Linux for a single large file by a modest amount), but it's also fundamentally changing the work that ripgrep has to do. When you memory map a file, the entire file is represented by a contiguous sequence of bytes (courtesy of the operating system). So all this "zap NUL" or "read entire file as a single line onto the heap" business completely goes away.

So this contributes to _this particular case_ being faster with memory maps on macOS. But as the file size grows, the suckiness of macOS memory maps seems to win out. **And also**, as soon as you use more "regular" input that isn't full of NUL bytes, then you're just left with whether memory maps on macOS are better than normal file I/O _all else being equal_. And while it seems it may be the case for smaller files, it becomes definitively _not_ the case for larger files.

TL;DR - File backed memory maps on macOS still suck and I am still unable to see a benchmark on Linux where the "sequential" advice matters.

So at least for now, I think I'm not seeing enough evidence here to merge this change. While it improves the pathological benchmark you've presented, it _significantly_ regresses non-pathological benchmarks.

Issue #3243, opened 2025-12-10
Add linkable headings to options in man pages


solonovamax:
#### Describe your feature request

This is a kinda random/niche request, though ideally should be rather trivial to add: it would be nice if there were linkable headings for each of the different options (in programs which support this), as there may be times when it's desireable to send a link to a specific option in a man page to someone, when using a website like the [arch linux man pages](https://man.archlinux.org/man/rg.1) which support links to headings.

An example of another man page which does this is `fd`. although I'm not *too* familiar with the man page syntax, however I *believe* the correct way to do this is to use `.TP`, although `.IP` might be better as it allows you to set the tag/label?

example link to a specific option in fd (albeit the names is not the best and could break if a new option starting with the same word is added, but it's better than nothing): https://man.archlinux.org/man/fd.1#no~2

BurntSushi:
Seems like a good idea.

ShivangiRay:
https://github.com/BurntSushi/ripgrep/pull/3521
Verified against the real generated rg.1 (15.2.0 release) by applying this transformation and rendering with mandoc: per-option anchors go from 0 to 104, all pre-existing rendered text is byte-for-byte unchanged (including the --color flag's nested bullet list, the one case that needed the original .RS 4/.RE body wrapper preserved), and no new mandoc lint warnings are introduced.

PR #3250, opened 2025-12-24
doc: add linkable headings for options

kanata9819:
This adds linkable headings for the option list in the generated man page.
Most options already work, but `--hidden` uses `.` as its short form (`-.`),
which prevents some man-to-HTML renderers from generating a stable anchor.
This change prefers the long option as the tag when the short option is not
ASCII alphanumeric, while preserving the usual short/long ordering for normal flags.

This addresses the request in #3243 to make option headings linkable.


wyf027:
Verified locally: `cargo test --locked` passes on this branch (320 tests).

The `.TP 12` + structured flag formatting approach looks good for linkable man page sections. Ready for review when you have time — thanks @kanata9819!

PR #3261, opened 2026-01-09
ignore/walk: add 'empty' and 'from_iter' methods, with tests

### LINKED ISSUE #1761: WalkBuilder: Improved API for adding multiple roots
In the creations I'm porting from Python or writing anew, the most common use I have for recursive directory traversal is to turn a mixture of file and directory paths given at the command line into a list of file paths to be processed.

While `ignore` and `walkdir` one-up Python's `os.walk` in one respect (I don't need to call `os.path.isdir` and bypass `os.walk` for non-directories), `WalkBuilder` makes taking a `Vec` of roots surprisingly un-ergonomic because the first one must be special-cased.

If I'm willing to mutate it, the simplest I've found is:

```rust
    if let Some(path1) = opts.inpath.pop() {
        let mut builder = WalkBuilder::new(path1);
        for path in opts.inpath {
            builder.add(path);
        }
        for result in builder.build() {
            // TODO: Process each file
        }

    }
```

If I'm not willing to mutate it, then it's worse.

It reminds me of the awkwardness of using `Iterator::fold` when what's really needed is `Iterator::fold_first`.

I'd like to propose one of the following:

* Add a `WalkBuilder::empty()` constructor so I can call `add` for the first argument too.
* Add a `WalkBuilder::from_iter()` constructor so I can just feed my `Vec` of roots directly to it.

jar-of-salt:
## Motivation

Closes #1761

Goal is to provide an expanded interface for creating `WalkBuilder` instances, specifically to avoid the pattern of having to call `add` repeatedly when you already know you have several paths you want to traverse.

E.g. courtesy of @ssokolow in the above issue
```rust
if let Some(path1) = opts.inpath.pop() {
    let mut builder = WalkBuilder::new(path1);
    for path in opts.inpath {
        builder.add(path);
    }
    for result in builder.build() {
        // TODO: Process each file
    }
}
```

Now you can just proceed as follows:
```rust
let builder = WalkBuilder::from_iter(opts.inpath);
```

## Details

I have added the following constructor methods for `WalkBuilder`:
- `empty`: produces an empty walk, which, when iterated before any calls to `add`, produces exactly zero items
- `from_iter`: produces a walk with the paths from the given iterator. If called with an empty iterator, produces the same result as `empty`

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -545,8 +545,18 @@ impl WalkBuilder {
     /// is better to call `add` on this builder than to create multiple
     /// `Walk` values.
     pub fn new<P: AsRef<Path>>(path: P) -> WalkBuilder {
+        let mut builder = WalkBuilder::empty();
+        builder.add(path);
+        builder
+    }
+
+    /// Create an empty builder to which paths can be added.
+    ///
+    /// Note that if you call `build` on this instance before calling `add`
+    /// on it, it will return exactly zero items during iteration.
+    pub fn empty() -> WalkBuilder {
         WalkBuilder {
-            paths: vec![path.as_ref().to_path_buf()],
+            paths: vec![],
             ig_builder: IgnoreBuilder::new(),
             max_depth: None,
             min_depth: None,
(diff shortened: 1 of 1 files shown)


jar-of-salt:
@BurntSushi this is ready for review, though I am unsure of how high priority this is given how old the related issue is.

ssokolow:
Embarrassingly, it's not how old the issue is that controls the urgency. The last few years have been a mess for me, so I haven't moved that far from the situations that the issue was opened under... the question is how rapidly my current efforts to pull myself together proceed.

BurntSushi:
LGTM. I've also added `Walk::from_iter`.

PR #3268, opened 2026-01-27
ci: replace cargo install with install-action for cargo-fuzz

xtqqczze:
Use [`taiki-e/install-action`](https://github.com/taiki-e/install-action) to install from a binary (from GitHub Releases) instead of compiling from source.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,9 +255,10 @@ jobs:
         with:
           toolchain: stable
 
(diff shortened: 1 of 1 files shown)


BurntSushi:
I would rather keep GitHub Action dependencies very lightweight. If the fuzzer action was taking a huge amount of time relative to the other actions, I might be more inclined to take this trade-off. But as of now, this just isn't a pain point worth solving.

PR #3271, opened 2026-02-03
ignore/types: add Containerfile

Pashugan:
This PR adds `Containerfile` as a synonym to `Dockerfile`.

`Containerfile` is the preferred file name in [Podman](https://docs.podman.io/en/latest/markdown/podman-build.1.html#file-f-containerfile) and [Buildah](https://github.com/containers/buildah/blob/main/docs/buildah-build.1.md).

In hindsight, `docker` could be not the best name for the file type, but my thinking is that it's best to leave it as it is than to add a new type, let alone rename it. Also, `container` would have been too generalized, and `containerfile` would have been too long to type.

### DIFF
--- a/crates/ignore/src/default_types.rs
+++ b/crates/ignore/src/default_types.rs
@@ -47,6 +47,7 @@ pub(crate) const DEFAULT_TYPES: &[(&[&str], &[&str])] = &[
     (&["cml"], &["*.cml"]),
     (&["coffeescript"], &["*.coffee"]),
     (&["config"], &["*.cfg", "*.conf", "*.config", "*.ini"]),
+    (&["container"], &["*Containerfile*", "*Dockerfile*"]),
     (&["coq"], &["*.v"]),
     (&["cpp"], &[
         "*.[ChH]", "*.cc", "*.[ch]pp", "*.[ch]xx", "*.hh",  "*.inl",


BurntSushi:
This seems a little odd to me. Podman and Buildah aren't Docker though right? So it seems weird for non-Docker things to show up in a search for Docker related files.

I agree the naming here is weird. `-t container` doesn't feel quite right.

Usually the ignore types are tied to a particular piece of software. Maybe we should just have `-t podman` and `-t buildah`?

Pashugan:
You are completely right that the situation here is a bit odd.

Yes, Podman and Buildah are alternatives to Docker tools, with similar functionality but having nothing to do with Docker Inc. as a vendor. As I understand it, the whole idea of using the name "Containerfile" was to emphasize vendor neutrality. Having that in mind, we can think about software that uses it as "container tools", e.g. Docker itself (though not by default but via `-f Containerfile`), Podman, Buildah etc.

I do agree that most people who historically used `-t docker` may not expect to suddenly find Containerfiles in the output. I am just not sure how serious this issue can be for them.

On the other hand, people who want Containerfiles likely expect to find Dockerfiles as well, because they are used interchangeably for backward compatibility with other tools than Docker. It can also be handy for many to have a file type that searches through both file names.

Hence, we perhaps should leave `-t docker` as it is.

The real question then is how to name the new type. I do not feel that naming after a tool/vendor is a good idea because of the reasons mentioned in the second paragraph. At the moment, I can't come up with a better idea than `-t containerfile`, so any help from the community is very much appreciated. It is long, but at least it's very specific.

Pashugan:
To make things more confusing, Apple's container tool [defaults](https://github.com/apple/container/blob/main/docs/command-reference.md#container-build) to Dockerfile and falls back to Containerfile. Either way, we cannot list all container tools and should go with something neutral.

BurntSushi:
I guess we should probably just call it `container`. And it should include both `Containerfile` and `Dockerfile`, right?

Pashugan:
I find `container` a bit odd, but eventually it's your call. We might have `containerfile` and `container` as a synonym/shortcut. If you're happy, I can push the updated version. Thanks.

BurntSushi:
I also find it odd. But `containerfile` is very long and perhaps suggests something more specific IMO that would exclude `Dockerfile`.

Pashugan:
I forgot to confirm that I pushed the change.

PR #3276, opened 2026-02-17
ignore: support `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` for `core.excludesFile`

### LINKED ISSUE #3275: Support `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM` environment variables for finding `core.excludesFile`
Git 2.32 (which was released in June 2021) introduced two environment variables for overriding git config file locations:

- GIT_CONFIG_GLOBAL  which overrides `~/.gitconfig` and `$XDG_CONFIG_HOME/git/config`
- GIT_CONFIG_SYSTEM which overrides `/etc/gitconfig`

The ignore crate's `gitconfig_excludes_path()` currently only checks the hardcoded paths (`$HOME/.gitconfig`, `$XDG_CONFIG_HOME/git/config`) when looking for `core.excludesFile`. It does not check these environment variables, so users who set `GIT_CONFIG_GLOBAL` to a non-standard path will have their global gitignore silently ignored by ripgrep while git itself respects it.

My use case is that my NixOS system manages system-wide git configuration at `/etc/git/config` and sets `GIT_CONFIG_GLOBAL` to `/etc/git/config`. The config contains `core.excludesFile = /etc/git/ignore`. Git commands like `git check-ignore` work correctly, but ripgrep never finds the global excludes file because it doesn't check `GIT_CONFIG_GLOBAL`.

Reproducer:

```sh
echo -e '[core]\n\texcludesFile = /tmp/test-ignore' > /tmp/test-gitconfig
echo '*.log' > /tmp/test-ignore
export GIT_CONFIG_GLOBAL=/tmp/test-gitconfig

mkdir /tmp/rg-test && cd /tmp/rg-test && git init
touch foo.txt bar.log

# git respects `GIT_CONFIG_GLOBAL`
git check-ignore bar.log    # outputs: bar.log

# ripgrep does not
rg --files                  # lists bar.log
rg --debug --files 2>&1 | grep 'opened gitignore'
# which only shows .git/info/exclude, meaning the global excludes file wasn't opened
```

I'd expect that when `GIT_CONFIG_GLOBAL` is set, ripgrep should read git config from that path (instead of `$HOME/.gitconfig` & `$XDG_CONFIG_HOME/git/config`) to find `core.excludesFile`, which would match git's own behavior. Similarly for `GIT_CONFIG_SYSTEM` overriding the system-level config path.

amaanq:
- Fixes #3275 

### Problem

`gitconfig_excludes_path()` resolves the global gitignore by reading
`core.excludesFile` from `$HOME/.gitconfig` and `$XDG_CONFIG_HOME/git/config`,
but git 2.32 (June 2021) introduced `GIT_CONFIG_GLOBAL` and `GIT_CONFIG_SYSTEM`
environment variables that override these locations. The `ignore` crate doesn't
check either of them, so users who point `GIT_CONFIG_GLOBAL` at a non-standard
path (e.g. NixOS managing git config at `/etc/git/config`) have their global
gitignore silently ignored by ripgrep while `git` itself respects it.

### Solution

This commit checks `GIT_CONFIG_GLOBAL` before the hardcoded paths, matching
git's semantics where it replaces both `$HOME/.gitconfig` and the XDG config.
It also adds system-level config (`GIT_CONFIG_SYSTEM`, falling back to
`/etc/gitconfig`) as a lowest-priority source for `core.excludesFile`, which
the crate previously didn't consult at all.

### DIFF
--- a/crates/ignore/src/gitignore.rs
+++ b/crates/ignore/src/gitignore.rs
@@ -568,19 +568,67 @@ impl GitignoreBuilder {
 ///
 /// Note that the file path returned may not exist.
 pub fn gitconfig_excludes_path() -> Option<PathBuf> {
-    // git supports $HOME/.gitconfig and $XDG_CONFIG_HOME/git/config. Notably,
-    // both can be active at the same time, where $HOME/.gitconfig takes
-    // precedent. So if $HOME/.gitconfig defines a `core.excludesFile`, then
-    // we're done.
-    match gitconfig_home_contents().and_then(|x| parse_excludes_file(&x)) {
-        Some(path) => return Some(path),
-        None => {}
+    // When GIT_CONFIG_GLOBAL is set, it replaces both $HOME/.gitconfig and
+    // $XDG_CONFIG_HOME/git/config (per git 2.32+). Otherwise, git supports
+    // both simultaneously, where $HOME/.gitconfig takes precedent.
+    if std::env::var_os("GIT_CONFIG_GLOBAL")
+        .as_deref()
+        .is_some_and(|v| !v.is_empty())
+    {
+        if let Some(path) = gitconfig_global_env_contents()
+            .and_then(|x| parse_excludes_file(&x))
+        {
+            return Some(path);
+        }
+    } else {
+        if let Some(path) =
+            gitconfig_home_contents().and_then(|x| parse_excludes_file(&x))
+        {
+            return Some(path);
+        }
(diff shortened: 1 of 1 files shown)


amaanq:
Friendly ping to look at this when you can @BurntSushi, thanks :)

PR #3280, opened 2026-02-27
doc: add FAQ entry for multi-file edits with ripgrep-edit

aagit:
Add documentation about using ripgrep-edit for multi-file edits, explaining how it complements ripgrep's search capabilities. This addresses the common need for efficient cross-file modifications in large codebases.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -27,6 +27,7 @@
 * [Can ripgrep replace grep?](#posix4ever)
 * [What does the "rip" in ripgrep mean?](#intentcountsforsomething)
 * [How can I donate to ripgrep or its maintainers?](#donations)
+* [How can I use ripgrep for multi-file edits?](#multifile-edits)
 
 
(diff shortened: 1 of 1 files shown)


BurntSushi:
If it's still being actively maintained in a year or so, I'd be happy to revisit this. Otherwise I'm not inclined to officially recommend it.

Moreover, this FAQ item seems redundant with [How can I search and replace with ripgrep?](https://github.com/BurntSushi/ripgrep/blob/master/FAQ.md#search-and-replace), which suggests a way to do it search-and-replace without another tool _and_ recommends [`fastmod`](https://github.com/facebookincubator/fastmod). I'd expect any new FAQ item to be integrated into what's there already, instead of bolted on without awareness of related FAQ items.

Finally, the license choice of `rg-edit` seems odd. I don't think this alone would stop me from otherwise recommending it, but it gives me pause.

PR #3302, opened 2026-03-16
globset: fix matches_all false positive when globs share a strategy bucket

### LINKED ISSUE #3290: GlobSet: `matches_all` returns true when multiple globs share the same match strategy

`master` branch, commit `4519153e5e461527f4bca45b042fff45c4ec6fb9`

build from source

macOS 26.3

### Describe your bug.

`GlobSet::matches_all` / `matches_all_candidate` are documented as returning `true` only when **all globs in the set** match the candidate path.

In the current implementation, `matches_all_candidate` checks whether **each populated strategy bucket** matches, not whether **each individual glob** matches. That produces false positives whenever two or more globs are placed into the same strategy bucket.

```rust
use globset::{Glob, GlobSetBuilder};

fn main() {
    let mut builder = GlobSetBuilder::new();
    builder.add(Glob::new("*.rs").unwrap());
    builder.add(Glob::new("*.c").unwrap());
    let set = builder.build().unwrap();

    // Expected: false, because "*.c" does not match "foo.rs".
    // Actual: true.
    assert!(!set.matches_all("foo.rs"));
}
```

This also reproduces with other same-strategy pairs, for example:

- `"foo"` + `"bar"` (literal strategy)
- `"src/"` + `"test/"` (prefix strategy)
- two suffix-based globs

`matches_all("foo.rs")` returns `true`.

`matches_all("foo.rs")` should return `false`, because not every glob in the set matches `"foo.rs"`.

xvchris:
When two or more globs are placed into the same internal strategy bucket (e.g. two extension globs like `*.rs` and `*.c`), `matches_all` returned `true` even when only one of them matched the path. The previous implementation checked whether each strategy bucket had _any_ match rather than whether each _individual glob_ matched.

Fix by delegating to `matches_candidate_into` and checking that the count of matched glob indices equals `self.len`.

Fixes #3290

### DIFF
--- a/crates/globset/src/lib.rs
+++ b/crates/globset/src/lib.rs
@@ -383,17 +383,14 @@ impl GlobSet {
     /// Returns ture if all globs in this set match the path given.
     ///
     /// This takes a Candidate as input, which can be used to amortize the cost
-    /// of peparing a path for matching.
+    /// of preparing a path for matching.
     ///
     /// This will return true if the set of globs is empty, as in that case all
     /// `0` of the globs will match.
     pub fn matches_all_candidate(&self, path: &Candidate<'_>) -> bool {
-        for strat in &self.strats {
-            if !strat.is_match(path) {
-                return false;
-            }
-        }
-        true
+        let mut matches = vec![];
+        self.matches_candidate_into(path, &mut matches);
+        matches.len() == self.len
     }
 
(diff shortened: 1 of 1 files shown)


BurntSushi:
I'm going to go with #3427 to avoid the allocation here. Thank you though!

PR #3307, opened 2026-03-19
fix: handle inline (?x) flag with comments in multi-pattern wrapping

### LINKED ISSUE #2677: Inline flag (?x) with comment doesn't work due to automatic wrapping of patterns inside non-capturing group

- [X] I have a different issue.

ripgrep 14.0.3

Using ripgrep_14.0.3-1_amd64.deb

Ubuntu 20.04

### Describe your bug.

Using `(?x)` to add comments isn't working due to automatic wrapping of the supplied pattern within `(?:)`.

Any pattern which uses `(?x)` and a comment.

```bash
$ echo 'fig a#b 123' | rg -o '(?x)\d+ #extract digits'
regex parse error:
    (?:(?x)\d+ #extract digits)
    ^
error: unclosed group
```

I think automatic wrapping of patterns inside non-capturing group was introducted for this issue: https://github.com/BurntSushi/ripgrep/issues/2480. If possible, don't modify the pattern when there's a comment. Or may be add a note in the documentation for this corner case.

FWIW, `GNU grep` doesn't support multiple patterns via `-e` or `-f` for PCRE:

```bash
$ echo 'fig a#b 123' | grep -oP '(?x)\d+ #extract digits'
123

$ echo 'MyText' | grep -P -e '(?i)notintext' -e 'text'
grep: the -P option only supports a single pattern
```

themavik:
## Summary
- Fix multi-pattern wrapping to handle `(?x)` inline flag with comments
- Previously, wrapping `(?x)pattern # comment` in `(?:...)` caused the comment to consume the closing paren

## Root Cause
When patterns are wrapped in `(?:...)` for multi-pattern support, a `# comment` in extended mode (`(?x)`) extends to end of line, consuming the closing `)` and causing an "unclosed group" error.

## Testing
- Verified `(?x)foo # comment` works correctly with multi-pattern wrapping
- Added regression test r2677
- Existing test suite passes (330 tests)

Fixes #2677

Made with [Cursor](https://cursor.com)

### DIFF
--- a/crates/pcre2/src/matcher.rs
+++ b/crates/pcre2/src/matcher.rs
@@ -49,11 +49,18 @@ impl RegexMatcherBuilder {
         let mut builder = self.builder.clone();
         let mut pats = Vec::with_capacity(patterns.len());
         for p in patterns.iter() {
-            pats.push(if self.fixed_strings {
-                format!("(?:{})", pcre2::escape(p.as_ref()))
+            let pat = p.as_ref();
+            let wrapped = if self.fixed_strings {
+                format!("(?:{})", pcre2::escape(pat))
+            } else if pat.contains("(?x") {
+                // In (?x) mode, # starts a comment to EOL. Adding newline
+                // before the closing ) prevents the comment from consuming it.
+                format!("(?:{}
+)", pat)
             } else {
-                format!("(?:{})", p.as_ref())
-            });
+                format!("(?:{})", pat)
+            };
+            pats.push(wrapped);
         }
         let mut singlepat = if patterns.is_empty() {
             // A way to spell a pattern that can never match anything.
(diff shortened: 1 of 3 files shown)


crates/pcre2/src/matcher.rs
```diff
@@ -49,11 +49,18 @@ impl RegexMatcherBuilder {
         let mut builder = self.builder.clone();
         let mut pats = Vec::with_capacity(patterns.len());
         for p in patterns.iter() {
-            pats.push(if self.fixed_strings {
-                format!("(?:{})", pcre2::escape(p.as_ref()))
+            let pat = p.as_ref();
+            let wrapped = if self.fixed_strings {
+                format!("(?:{})", pcre2::escape(pat))
+            } else if pat.contains("(?x") {
```
ltrzesniewski, line 55: Let's just say this is not the correct way of handling the issue: what about stuff like `(?ix)`, or `(?-x)` later on, etc... The AI just tried to handle a very specific case.

PR #3314, opened 2026-03-23
fix: add -- separator before filename in compression tools

### LINKED ISSUE #3222: Calls to compression tools need to separate file names from options

ripgrep 15.1.0 (rev 57c190d56e)

features:-pcre2

PCRE2 is not available in this build of ripgrep.

Cargo build from the main branch of git

Fedora 43

### Describe your bug.

I have a directory full of gzipped files with names or path names that start with a dash. They can be named like '-10:29.log.gz'. When I search in them using `rg -z ...`, I get the following error:

```
-------------------------------------------------------------------------------
gzip: invalid option -- '0'
Try `gzip --help' for more information.
-------------------------------------------------------------------------------
```

The reason it's happening is that ripgrep calls the compression tools without separating options from file names with the customary `--` separator. gzip is treating the filename as a string of options, doesn't recognize a letter (or digit) in them, and bails.

This is likely the case with other compression tools.

Create a file named '-10.txt' with some text, gzip it (for example, `gzip ./-10.txt`), and try to `rg -z sometext`.

```
-------------------------------------------------------------------------------
gzip: invalid option -- '0'
Try `gzip --help' for more information.
-------------------------------------------------------------------------------
```

I would expect ripgrep to search within such a file.

mango766:
When searching compressed files with names that start with a dash (e.g., `-10.txt.gz`), ripgrep fails because the compression tool interprets the filename as command options.

Here's the issue that this PR addresses:
```
$ echo "hello" > ./-10.txt
$ gzip ./-10.txt
$ rg -z .
gzip: invalid option -- '0'
Try `gzip --help' for more information.
```

This happens because ripgrep calls compression tools like `gzip -d -c -10.txt.gz`, where `-10.txt.gz` gets parsed as invalid options.

The fix is simple: add `--` before the filename argument. This is the standard POSIX "end of options" marker that tells the tool everything following it should be treated as positional arguments, not options.

After this fix:
```
$ echo "hello" > ./-10.txt
$ gzip ./-10.txt
$ rg -z hello
-10.txt.gz:1:hello
```

This aligns with how most shell commands handle filenames that might be confused with options.

Fixes #3222

### DIFF
--- a/crates/cli/src/decompress.rs
+++ b/crates/cli/src/decompress.rs
@@ -226,7 +226,7 @@ impl DecompressionReaderBuilder {
         let Some(mut cmd) = self.matcher.command(path) else {
             return DecompressionReader::new_passthru(path);
         };
-        cmd.arg(path);
+        cmd.arg("--").arg(path);
 
         match self.command_builder.build(&mut cmd) {
             Ok(cmd_reader) => Ok(DecompressionReader { rdr: Ok(cmd_reader) }),


jafd:
Prior art: https://github.com/BurntSushi/ripgrep/pull/3224

PR #3321, opened 2026-03-26
Fix parent ignore matching for multiple explicit roots

### LINKED ISSUE #3320: Order-dependent .rgignore bug with multiple explicit search roots

ripgrep 15.1.0

features:+pcre2
simd(compile):+NEON
simd(runtime):+NEON

PCRE2 10.45 is available (JIT is available)

Homebrew

macOS 15.7.4

### Describe your bug.

`rg` appears to handle `.rgignore` inconsistently when multiple explicit search roots are provided.

In a minimal repro with:

- `.rgignore` containing `beta/**/*.svg`
- `alpha/a.txt` containing `AWS`
- `beta/x.svg` containing `AWS`

the command:

```sh
rg --files-with-matches -n 'AWS' alpha beta
```

incorrectly returns `beta/x.svg`, even though it matches the ignore rule.

If I run the same search with the roots reversed:

```sh
rg --files-with-matches -n 'AWS' beta alpha
```

then `beta/x.svg` is correctly ignored.

So the bug seems to be that `.rgignore` handling is order-dependent when multiple explicit search roots are passed to ripgrep.

Create this minimal repro directory:

```txt
.
├── .rgignore
├── alpha/
│   └── a.txt
└── beta/
    └── x.svg
```

With these contents:

`.rgignore`
```txt
beta/**/*.svg
```

`alpha/a.txt`
```txt
AWS
```

`beta/x.svg`
```txt
AWS
```

Then run these commands from that directory:

```sh
rg --files-with-matches -n 'AWS' beta
rg --files-with-matches -n 'AWS' alpha beta
rg --files-with-matches -n 'AWS' beta alpha
```

You can also reproduce it with this shell script:

```sh
mkdir repro && cd repro
mkdir -p alpha beta
printf 'beta/**/*.svg\n' > .rgignore
printf 'AWS\n' > alpha/a.txt
printf 'AWS\n' > beta/x.svg

rg --files-with-matches -n 'AWS' beta
rg --files-with-matches -n 'AWS' alpha beta
rg --files-with-matches -n 'AWS' beta alpha
```

The result depends on the order of the explicit search roots.

Command:

```sh
rg --files-with-matches -n 'AWS' alpha beta
```

Actual output:

```txt
alpha/a.txt
beta/x.svg
```

But `beta/x.svg` should be ignored by `.rgignore` because it matches:

```txt
beta/**/*.svg
```

If I swap the root order:

```sh
rg --files-with-matches -n 'AWS' beta alpha
```

the output changes to:

```txt
alpha/a.txt
```

which is what I expect.

`--debug` for the working order shows that ripgrep does recognize the ignore rule and ignores `beta/x.svg`:

```sh
rg --debug --files-with-matches -n 'AWS' beta alpha
```

Relevant debug output:

```txt
rg: DEBUG|ignore::walk|crates/ignore/src/walk.rs:1942: ignoring beta/x.svg: Ignore(IgnoreMatch(Gitignore(Glob { from: Some("/path/to/repro/.rgignore"), original: "beta/**/*.svg", actual: "beta/**/*.svg", is_whitelist: false, is_only_dir: false })))
```

So the bug appears to be that ignore handling is order-dependent when multiple explicit search roots are provided.

Both commands should produce the same result regardless of the order of the explicit search roots, and both should respect `.rgignore`.

Given this ignore rule:

```txt
beta/**/*.svg
```

`beta/x.svg` should never be searched.

So both of these commands:

```sh
rg --files-with-matches -n 'AWS' alpha beta
rg --files-with-matches -n 'AWS' beta alpha
```

should return:

```txt
alpha/a.txt
```

and should not report `beta/x.svg`.

markusylisiurunen:
Fixes #3320.

This fixes an order-dependent bug in parent ignore matching when ripgrep is given multiple explicit search roots.

The root cause was in `Ignore::add_parents`. Absolute-parent matchers were cached only by ancestor directory path, but they also carried root-specific state in `absolute_base`. When a later explicit root reused a cached matcher built for an earlier root, parent ignore rules could be evaluated relative to the wrong root, which made `.rgignore` / parent ignore behavior depend on the order of the explicit roots.

The fix keeps the existing per-directory cache, but stops reusing the cached matcher node directly across roots. Instead, on a cache hit, it rebinds the current parent chain and `absolute_base` while reusing the compiled matcher data. The cache entry is then refreshed to keep reuse working across successive roots.

Tests added:

- a low-level regression in `crates/ignore/src/dir.rs` covering reuse across multiple explicit roots
- a traversal regression in `crates/ignore/src/walk.rs` covering both root orders

Validation:

- `cargo test -p ignore`
- repro from #3320 now behaves correctly in both single-threaded and parallel runs

### DIFF
--- a/crates/ignore/src/dir.rs
+++ b/crates/ignore/src/dir.rs
@@ -100,7 +100,9 @@ struct IgnoreInner {
     ///
     /// Note that this is never used during matching, only when adding new
     /// parent directory matchers. This avoids needing to rebuild glob sets for
-    /// parent directories if many paths are being searched.
+    /// parent directories if many paths are being searched. When reusing a
+    /// cached matcher for a different explicit root, the traversal specific
+    /// state is rebound to the new parent chain.
     compiled: Arc<RwLock<HashMap<OsString, Weak<IgnoreInner>>>>,
     /// The path to the directory that this matcher was built from.
     dir: PathBuf,
@@ -214,7 +216,15 @@ impl Ignore {
             let mut compiled = self.0.compiled.write().unwrap();
             if let Some(weak) = compiled.get(parent.as_os_str()) {
                 if let Some(prebuilt) = weak.upgrade() {
-                    ig = Ignore(prebuilt);
+                    let rebound = ig.add_child_path_from_prebuilt(
+                        &prebuilt,
+                        absolute_base.clone(),
+                    );
+                    compiled.insert(
+                        parent.as_os_str().to_os_string(),
+                        Arc::downgrade(&rebound.0),
+                    );
+                    ig = rebound;
                     continue;
                 }
             }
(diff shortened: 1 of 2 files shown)


BurntSushi:
I went with https://github.com/BurntSushi/ripgrep/pull/3420 instead

PR #3282, opened 2026-03-07
fix: use -- separator when passing file paths to decompression tools

luinbytes:
## Summary

Fixes #3222 - Files with dash-prefixed names (e.g., `-10.txt.gz`) were causing decompression tools to fail with errors like `gzip: invalid option -- '0'` when using the `-z` flag.

## The Problem

When searching compressed files with `-z`, ripgrep spawns external decompression tools (gzip, bzip2, xz, etc.) and passes the file path as an argument:

```
gzip -d -c ./-10.txt.gz
```

This fails because gzip interprets `-10.txt.gz` as options, not a filename.

## The Fix

Added `--` separator before the file path to signal end of options:

```
gzip -d -c -- ./-10.txt.gz
```

## Testing

Verified locally:
```bash
$ echo "test content" > ./-10.txt
$ gzip ./-10.txt
$ rg -z test .
./-10.txt.gz:test content
```

## Changes

- `crates/cli/src/decompress.rs`: Single line change to add `--` separator

This is a minimal, safe fix that follows standard POSIX conventions for handling dash-prefixed filenames.

PR #3334, opened 2026-03-30
fix(cli): use '--' separator when passing file paths to decompression tools

### LINKED ISSUE #3222: Calls to compression tools need to separate file names from options

xingzihai:
## Summary

Fixes an issue where compressed files starting with `-` could not be searched because decompression tools would interpret the filename as a command-line option.

## Problem

When using `rg -z` to search compressed files with names starting with `-` (e.g., `-10.txt.gz`), the decompression tool (gzip, zstd, etc.) would incorrectly parse the filename as an option flag.

Example:
```bash
rg -z 'pattern' '-file.gz'  # Fails: gzip treats '-file.gz' as options
```

## Solution

Add `--` argument separator before the file path when invoking decompression tools. This tells the tool to treat all subsequent arguments as filenames, not options.

```rust
cmd.arg("--");  // Add separator before file path
cmd.arg(path);
```

## Changes

- `crates/cli/src/decompress.rs` - Added `--` separator in `DecompressionReaderBuilder::build()`

## Testing

- `cargo build` ✅
- `cargo test` ✅ (329 passed, 0 failed)
- Manual testing with files starting with `-` ✅

Fixes #3222

### DIFF
--- a/crates/cli/src/decompress.rs
+++ b/crates/cli/src/decompress.rs
@@ -226,6 +226,9 @@ impl DecompressionReaderBuilder {
         let Some(mut cmd) = self.matcher.command(path) else {
             return DecompressionReader::new_passthru(path);
         };
+        // Use '--' to separate options from the file path, which ensures
+        // that file paths starting with '-' are not interpreted as options.
+        cmd.arg("--");
         cmd.arg(path);
 
         match self.command_builder.build(&mut cmd) {


jafd:
[Previously](https://github.com/BurntSushi/ripgrep/pull/3314), [previously](https://github.com/BurntSushi/ripgrep/pull/3282), [previously](https://github.com/BurntSushi/ripgrep/pull/3224). I'm betting someone will make another one in a couple days.

PR #3340, opened 2026-03-31
macOS: improve default explicit-file mmap searches

kmad:
### Summary

  On macOS, small explicit-file searches now use the existing auto-mmap path by default. The mmap
  path also gets sequential-read advice (SEQUENTIAL / WILLNEED) for full-file scans, and inverted
  matches continue to use the buffered path where mmap is a worse tradeoff.

  The rest of the branch is limited to a few small searcher-path simplifications plus regression
  coverage for the new macOS behavior and existing matcher invariants.

  ### What Changed

  - enable the existing auto-mmap path on macOS for small explicit-file searches
  - add madvise(SEQUENTIAL) / madvise(WILLNEED) on the mmap path for sequential full-file scans
  - skip mmap for inverted matches
  - simplify a few hot searcher paths:
      - cache is_line_by_line_fast()
      - cache the binary-detection byte
      - use simpler loop bounds in hot loops
      - add a fast path for single-byte line terminators in without_terminator
  - add regression coverage for:
      - macOS default auto-mmap behavior
      - matcher invariant panics

  ### Why

  The main win here is narrow but real: on Apple Silicon/macOS, large single-file explicit searches
  get faster without changing CLI usage.

  Local benchmarking (on an M1 Macbook Pro) against master showed the clearest gains on large single-file scans:

  | Benchmark | Before | After | Improvement |
  | --- | ---: | ---: | ---: |
  | Simple literal | 0.0658s | 0.0572s | +13.1% |
  | No-match scan | 0.0421s | 0.0271s | +35.6% |

  This PR is not claiming a broad speedup across all ripgrep workloads. Directory-heavy and more CPU-
  bound regex workloads were much smaller wins or effectively flat in my testing.

  ### Commit Map

  - 4abea6e perf: improve macOS explicit-file mmap path
  - e442434 perf: simplify hot searcher paths
  - b9d9d2f test: add regression coverage for macOS mmap and matcher invariants

  ### Validation

  - cargo test --release -p grep-matcher
  - cargo test --release --test integration auto_mmap
  - cargo test --release mmap --test integration
  - cargo test --release line_numbers --test integration

Codeveloped with claude-opus-4-6 and gpt-5.4-high

### DIFF
--- a/crates/matcher/tests/test_matcher.rs
+++ b/crates/matcher/tests/test_matcher.rs
@@ -79,6 +79,18 @@ fn shortest_match() {
 }
 
 #[test]
+#[should_panic]
+fn match_new_panics_on_invalid_range() {
+    Match::new(2, 1);
+}
+
+#[test]
+#[should_panic]
+fn match_offset_panics_on_overflow() {
+    Match::new(1, 2).offset(usize::MAX);
+}
+
+#[test]
 fn captures() {
     let matcher = matcher(r"(?P<a>\w+)\s+(?P<b>\w+)");
     assert_eq!(matcher.capture_count(), 3);
--- a/crates/searcher/src/lines.rs
+++ b/crates/searcher/src/lines.rs
@@ -119,6 +119,16 @@ pub(crate) fn without_terminator(
     bytes: &[u8],
     line_term: LineTerminator,
 ) -> &[u8] {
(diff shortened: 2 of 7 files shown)


BurntSushi:
It seems like there are a bunch of unrelated changes in this PR. Please back those out so that I can evaluate the actual change here.

PR #3344, opened 2026-04-02
ignore: fix panic recovery after filter_entry panic

lopopolo:
## Summary

This fixes panic recovery in `ignore::Walk` after a panicking `filter_entry` predicate.

Before this change, `Walk::next` called `filter_entry` on a directory before it pushed the child ignore matcher. If the predicate panicked and the caller resumed iteration with `catch_unwind`, the later `WalkEvent::Exit` still popped an ignore frame. That could drop ignore rules for later siblings and leak paths that `.ignore` should have hidden. The new regression test demonstrates the failure by leaking `x` and `x/file`.

The fix moves the child ignore-state transition ahead of the user callback so the ignore stack stays balanced across a recovered panic.

## Context

I found this bug with OpenAI Codex while auditing ripgrep for bugs in the same unwind-safety class as [`RUSTSEC-2026-0078`](https://rustsec.org/advisories/RUSTSEC-2026-0078.html). @lopopolo reviewed and validated the reproducer, patch, and PR.

## Validation

- `cargo test -p ignore filter_entry_panic_should_not_drop_ignore_rules_for_later_siblings -- --nocapture`
- `cargo test -p ignore --lib`
- `cargo fmt --all --check`

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -2049,8 +2049,12 @@ mod tests {
     use std::ffi::OsStr;
     use std::fs::{self, File};
     use std::io::Write;
+    use std::panic::{AssertUnwindSafe, catch_unwind};
     use std::path::Path;
-    use std::sync::{Arc, Mutex};
+    use std::sync::{
+        Arc, Mutex,
+        atomic::{AtomicBool, Ordering},
+    };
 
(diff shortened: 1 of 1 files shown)


BurntSushi:
Hmmm I'm not sure about this one. When I run the regression test on current master, it passes.

Issue #1992, opened 2021-09-16
Parse errors in .gitignore do not sanitize some terminal escape sequences before printing to stderr


Kimundi:
#### What version of ripgrep are you using?

```
ripgrep 11.0.2
-SIMD -AVX (compiled)
+SIMD +AVX (runtime)
```

#### How did you install ripgrep?

`apt install ripgrep` in a Ubuntu 20.04 WSL2 install.

#### What operating system are you using ripgrep on?

Ubuntu 20.04 in a WSL2

#### Describe your bug.

The line printed from a .gitignore in case of an parse errors might contain terminal escape sequences that mess up the terminal state.

#### What are the steps to reproduce the behavior?

I'm not sure which kind of terminal escape sequences these are, so in case this is terminal emulator specific:

- Be on windows
- Use the new "Windows Terminal"
- Be logged into a WSL2 instance with it (bash or fish)
- Have a `.gitignore` containing the following bytes: `\x5b\x5d\xc2\x96\xc2\x97\xc2\x98\xc2\x99\xc2\x9a\x0a`
- run `rg foo` or any other command that causes it to parse `.gitignore` files.
- You should now be able to observe that some terminal control sequences got send to the terminal

#### What is the actual behavior?

```
> hexdump -C .gitignore
00000000  5b 5d c2 96 c2 97 c2 98  c2 99 c2 9a 0a           |[]...........|
0000000d
> rg foo
./.gitignore: line 1: error parsing glob '[]': unclosed character class; missing ']'
^[[?1;0c
> [?1;0c
```

![grafik](https://user-images.githubusercontent.com/2903206/133650874-39677ecb-4a7f-43cc-81e8-9910e3e90aca.png)

#### What is the expected behavior?

Any terminal escape sequences should get sanitized out from the line, at least if printed to an interactive terminal.

BurntSushi:
I can't seem to reproduce this:

![no-color-mangling](https://user-images.githubusercontent.com/456674/134041804-0a6ef063-a8d5-44b6-8722-79120668ba4c.png)

Either way, I appreciate the bug report, but I think I'm going to mark this as `wontfix`. ripgrep passes through content it reads, as-is, pretty much everywhere. The same would happen if you were searching a file and ripgrep printed a match with ANSI escape codes, for example. Similarly for file names and probably other stuff. I'm not really sure it's ripgrep's place to get super paranoid about this and start cleansing everything it prints. It certainly doesn't seem like other tools do it.

PR #3371, opened 2026-04-18
Sanitize control bytes

curious-rabbit:
Filenames and file contents can contain terminal control sequences like OSC52 clipboard writes, OSC8 hyperlink injection, cursor moves that spoof output. When ripgrep prints them to a tty, the terminal acts on them. This PR sanitizes the relevant sequences before they reach the terminal.

Sanitization is gated on whether stdout is an interactive tty. Pipe output is left byte-exact so `rg | xargs`, `rg | jq`, `rg --files |
xargs -0`, and similar pipelines see exactly the bytes they see today.

Implementation is a 256-byte lookup table with a four-way branch. The fast path returns Cow::Borrowed without allocating. On a 50 MB / ~80k-match Rust corpus the tty path runs within the noise of the unsanitized baseline.

Library API change is additive: one is_terminal(bool) method per printer (Standard, Summary, PathPrinter), default false. Existing embedders see no behavior change unless they opt in.

Covers: matched-line content, printed paths in standard/summary/path-list output, and stderr error messages from .gitignore parsing (see issue #1992). JSON output is unaffected.

@BurntSushi I know you had some resevationas bout filters in the past but I hope this approach with interactive vs piped matches how you want ripgrep to behave while we can also address the relevant risks. There is no measurable performance impact

### DIFF
--- a/crates/core/flags/hiargs.rs
+++ b/crates/core/flags/hiargs.rs
@@ -549,6 +549,7 @@ impl HiArgs {
         builder
             .color_specs(self.colors.clone())
             .hyperlink(self.hyperlink_config.clone())
+            .is_terminal(self.is_terminal_stdout)
             .separator(self.path_separator.clone())
             .terminator(self.path_terminator.unwrap_or(b'\n'));
         builder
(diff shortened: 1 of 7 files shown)


curious-rabbit:
Any chance for a decision on this?

PR #3384, opened 2026-04-26
doc: remove dangling ignore error link

### LINKED ISSUE #3235: Dangling link in docs for `ignore::Error::io_error`
In [docs](https://docs.rs/ignore/0.4.25/ignore/enum.Error.html#method.io_error) for `ignore::Error::io_error`, there's the following paragraph:
> This is the original [std::io::Error](https://doc.rust-lang.org/stable/std/io/struct.Error.html) and is not the same as [impl From<Error> for std::io::Error](https://docs.rs/ignore/0.4.25/ignore/struct.Error.html#impl-From%3CError%3E) which contains additional context about the error.

The second link however leads to "This resource does not exist" page (which is understandable, since there's no `struct Error` in `ignore`), and, furthermore, I can't see the implementation it could refer to, either - this link seems to be dangling from [the very beginning](https://github.com/BurntSushi/ripgrep/commit/873abecbf1c4fbac2e070ac33a25e7a23fea6700). Is this some artifact of the first unmerged implementation, or something that is actually expected to exist?

MestreY0d4-Uninter:
## Summary

- remove the `ignore::Error::io_error` rustdoc paragraph that references a non-existent `impl From<Error> for std::io::Error`
- remove the now-unused rustdoc link references for that paragraph

Closes #3235.

## Validation

Passed locally:

- `cargo fmt --all -- --check`
- `cargo doc -p ignore --no-deps`
- `cargo test -p ignore`
- `cargo check`

Notes:

- I also tried `cargo clippy --all-targets -- -D warnings`, but with my local Rust 1.94 toolchain it reports warnings in untouched crates unrelated to this rustdoc-only change.
- I also tried `cargo test --all`, but my local checkout ran out of disk space while compiling full-workspace dependencies (`No space left on device`).

### DIFF
--- a/crates/ignore/src/lib.rs
+++ b/crates/ignore/src/lib.rs
@@ -192,16 +192,10 @@ impl Error {
     /// This method returns a borrowed value that is bound to the lifetime of the [`Error`]. To
     /// obtain an owned value, the [`into_io_error`] can be used instead.
     ///
-    /// > This is the original [`std::io::Error`] and is _not_ the same as
-    /// > [`impl From<Error> for std::io::Error`][impl] which contains
-    /// > additional context about the error.
-    ///
     /// [`None`]: https://doc.rust-lang.org/stable/std/option/enum.Option.html#variant.None
     /// [`std::io::Error`]: https://doc.rust-lang.org/stable/std/io/struct.Error.html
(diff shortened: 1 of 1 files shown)


MestreY0d4-Uninter:
Closing this in favor of #3386, which fixes the same dangling rustdoc link with a narrower one-line change while preserving the explanatory docs.

PR #3396, opened 2026-05-10
resolve relative paths in GitignoreBuilder::add relative to root

joshvoigts:
`GitignoreBuilder::add` takes a path and does `File::open(path)` directly without joining with the root. The docs say paths should be relative to the root given to `GitignoreBuilder::new`, but the code ignores the root entirely.

So if you do:
```rust
let builder = GitignoreBuilder::new("/some/repo");
builder.add(".gitignore");  // Tries to open .gitignore relative to CWD, instead of /some/repo
```

Can be fixed by joining the path with `self.root`. Absolute paths pass through `join` unchanged  (continue to work as they previously did).

### DIFF
--- a/crates/ignore/src/gitignore.rs
+++ b/crates/ignore/src/gitignore.rs
@@ -386,13 +386,17 @@ impl GitignoreBuilder {
     ///
     /// The file given should be formatted as a `gitignore` file.
     ///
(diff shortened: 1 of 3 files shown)


joshvoigts:
Closing, misunderstood the implementation.

Issue #3293, opened 2026-03-11
Performance penalty of probing ignore files under every subdirectory (especially on Windows), and a possible solution


AlanIWBFT:
On Windows, both `fs::metadata()` and `Path::exists()` go to `std::sys::fs::windows::File::open` and eventually `CreateFileW` which is very heavy:

<img width="50%" alt="Image" src="https://github.com/user-attachments/assets/beecc5d3-cfaf-41cc-99f1-73cc9d49cc75" />

While `GetFileAttributesW` is slightly lighter, there is actually a better alternative. Windows prefers fetching metadata of directories in batches using `FindFirstFileExW/FindNextFileW`, which maps to `fs::read_dir`. So, if we prefetch the content of the entire directory using `read_dir`, use it to check the existence of ignore files, and visit the entries later, it can be potentially much faster as we avoid touching files individually. And it turned out true. I prototyped one here: https://github.com/AlanIWBFT/ripgrep/commit/4e8fb872a06231f8d61e1b6dd70d5f9a14396399. The effectiveness of this optimization depends on the ratio of files/directories. More dirs = faster.

One thing worth mentioning is only the multi-thread path is touched in the prototype as it is already using `read_dir`.
## Benchmark results

I tested it on three large repos: Unreal Engine 5.7.3, LLVM 22.1.0 and Chromium 147.0.7727.1. Patterns:

Unreal Engine 5.7.3: `FLightmapRenderer::FLightmapRenderer`
LLVM 22.1.0: `opt<CompileArgsFrom> CompileArgsFrom`
Chromium 147.0.7727.1: `#define GL_CHROMIUM_pixel_transfer_buffer_object 1`

These patterns are picked so that exactly 1 match will be found across the repository, avoiding noise from outputting to terminal.

5 warmup runs and 10 benchmark runs. 

### Windows (Windows 11 10.0.26200.7922)

Unreal Engine 5.7.3:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
| batch-fetch | 2.418 ± 0.017 | 2.386 | 2.436 | 1.00 |
| original | 3.644 ± 0.026 | 3.606 | 3.683 | **1.51** ± 0.02 |

LLVM 22.1.0:

| Command | Mean [s] | Min [s] | Max [s] | Relative |
|:---|---:|---:|---:|---:|
[...]

PR #3410, opened 2026-05-21
Ignore files probing optimization: fetch directory entry metadata in batches

AlanIWBFT:
This PR contains my fix to the issue #3293 I brought up. I was holding back making a PR (I made a fix but only linked in in the issue body) as nowadays owners are overwhelmed with AI generated PRs. But then someone popped up and made a PR with almost identical code: https://github.com/BurntSushi/ripgrep/pull/3405 . So I feel forced to do it, especially when that branch contains a suspicious name `codex/prefetch-ignore-files`.
You can see the GitHub verified GPG signature showing initial submission date being Mar 11, 2026.

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -1531,7 +1531,9 @@ impl Work {
                 ),
             }
         }
-        let (ig, err) = self.ignore.add_child_with_entries(self.dent.path(), &result.entries);
+        let (ig, err) = self
+            .ignore
+            .add_child_with_entries(self.dent.path(), &result.entries);
         self.ignore = ig;
         self.dent.err = err;
         Ok(result)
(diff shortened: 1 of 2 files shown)


crates/ignore/src/walk.rs
```diff
@@ -1712,6 +1747,12 @@ impl<'s> Worker<'s> {
                 return state;
             }
         }
+        for err in readdir.errors {
+            let state = self.visitor.visit(Err(err));
+            if state.is_quit() {
+                return state;
+            }
+        }
```
BurntSushi, line 1755: This is slightly different than the status quo, where valid entries and errors were inter-mingled.

I think this is okay though, since we never guaranteed an ordering. And it's probably better UX to surface all valid results before all errors.

BurntSushi (approved):
Fantastic change, thank you. I made some small tweaks, but this otherwise LGTM.

Issue #3395, opened 2026-05-08
Add hierarchy-aware mode to include hierarchical parent context in search results


layzerar:
I’d like to propose a new `--hierarchy` option for ripgrep that prints hierarchical parent lines for each match, giving search results more useful context without substantially increasing output size.

For example, given `test.md`:

```md
list content:
  1. list item (level 1)
  2. list item (level 1)
    - inner list item: foobar (level 2)
```

Running:

```sh
rg --hierarchy foobar test.md
```

could produce:

```text
test.md
1: list content:
3:   2. list item (level 1)
4:     - inner list item: foobar (level 2)
```

The idea is that when a matching line is found, ripgrep would also print the relevant ancestor lines that establish its hierarchy. In this example, the match is inside a nested Markdown list, so the output includes the top-level context line and the parent list item.

This would be useful for more than Markdown. Many common file formats and programming languages encode useful structure through indentation or simple block hierarchy, including Python, C, C++, Java, YAML, TOML, configuration files, logs with grouped sections, and documentation formats. Current context options like `-A`, `-B`, and `-C` are line-distance based, which can be noisy or miss the actual structural context. A hierarchy-aware mode could provide more precise context with fewer extra lines.

The implementation would not need to perform full syntax parsing. A simple indentation/block-based heuristic would already be valuable:

- Track previous lines that appear to be structural ancestors.
- When a match is found, emit the nearest relevant parent lines before the matching line.
- Deduplicate ancestor lines when multiple matches share the same hierarchy.
- Preserve normal ripgrep behavior unless `--hierarchy` is explicitly enabled.

This should keep the feature relatively non-invasive and avoid significant performance impact compared with language-specific parsing. It would provide a middle ground between minimal search output and verbose surrounding-context output, making ripgrep results easier to understand when matches occur inside nested structures.

yash3605:
@BurntSushi I can try Implementing it. if you're up for it !!. Although it seems it'll require some core changes but I can refactor it based on your review.

BurntSushi:
I don't think this is a good fit for ripgrep. This is the kind of feature that will likely beget many more feature requests to tweak how it works. It just expands the scope too much.

PR #3413, opened 2026-05-23
Feat: Add --hierarchy flag for showing parent context lines

yash3605:
Implemented Feature suggested in Issue: #3395. 

I was able to do it using a simple **Stack**. what I did was that - when we're reading file, we check the **indentation** of each line. if the indentation of current line is less than the previous ones then we Pop them from the stack otherwise we push. and when I match is found the Stack is printed as it is.

what I couldn't do was implementing the Stack thing via _StandardSink_ or _Sink_. because the code base is modular and tightly coupled and I was facing some ownership issues when borrowing _StandardSink_ and others. So I had to create the stack  separately and then pass it forward by some workarounds like creating a **SinkWrapper** for Sink and then Pass the Map I have made for the output.

Same with Printing the final result. I could have used _WriteColor_ trait but I was facing same Ownership issues and I couldn't borrow another mutable reference so here I went with plain Println!() statement.

**Before Hierarchy:** 

**After Hierarchy (in Debug mode):**

**After Hierarchy (in build/release):**

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,18 +4,18 @@ version = 4
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.3"
+version = "1.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8e60d3430d3a69478ad0993f19238d2df97c507009a52b3c10addcd7f6bcb916"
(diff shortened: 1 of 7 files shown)


yash3605:
This last ci/test error is due to Github Actions Infrastructure. I have fixed most issues in ci.

BurntSushi:
See https://github.com/BurntSushi/ripgrep/issues/3395#issuecomment-4925067103

PR #3414, opened 2026-05-25
core: add --base64 and --base64url flags

lionello:
This feature allows you to search for strings in base64 or base64url-encoded data, for example leaked credentials:

> $ rg admin --base64    
> coredns/plugin/test/file.go
> 53:AwwKa3ViZS**1hZG1pbj**CCASIwDQYJKoZIhvcNAQEBBQADggEPADCCAQoCggEBAMTw

It works by adding all possible patterns (with all lead/trail bytes) to the patterns list, before passing it to the engine. Setting any of the options implies `--fixed-strings`:

- `--base64` uses the standard alphabet (RFC 4648, +/)
- `--base64url` uses the URL- and filename-safe alphabet (RFC 4648, -_)

(The difference between the two really only appears when searching for text with many bits set, eg. `"?"`.)

Disclaimer: the original prototype was written by me by hand in Golang, but I've used Claude Code to help me ramp up on the ripgrep codebase and its use of Rust idioms. In particular, it wrote the base64-encoder, instead of adding a new dep on the `base64` crate.

There's room for improvement: when searching for a string, the generated patterns will contain the variants that are preceded by 1 and by 2 bytes (2 resp. 4 trailing bits), but the final (fixed-string) patterns do not require any prefix. Because of this, short search strings (length≤2) will have many false positives.

By the way, the existing `fn base64_standard` could not be reused, because it pads the result with `=`, which we don't want here. There's some dedup options, but not worth it IMHO.

### DIFF
--- a/crates/core/base64.rs
+++ b/crates/core/base64.rs
@@ -1,0 +1,285 @@
+/*!
+Enumerate every fixed-string encoding of a literal byte query under a base64
+alphabet.
+
+Base64 packs 8-bit bytes into 6-bit groups, so the same literal can appear in
+encoded output at three different bit offsets (0, 2, or 4 leading bits owned
+by a neighboring byte). For each offset we enumerate the unknown leading and
+trailing bits, encode the resulting padded buffer, and trim the chars that
+cover those unknown bits. The disjunction of all such encodings is what we
+want to search for; we return the literals themselves so callers can feed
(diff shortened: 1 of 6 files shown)


BurntSushi:
This isn't a good fit for ripgrep. You should have a wrapper script do this for you.

PR #3420, opened 2026-05-26
ignore: scope compiled parent matchers by root

### LINKED ISSUE #3419: Nondeterminism in ignore::WalkBuilder parallel multi-root walk

14.1.1

Cargo

macOS 26.5

### Describe your bug.

`ignore::WalkBuilder` appears to produce nondeterministic results for a parallel multi-root walk when one root has a scoped ignore rule that should not apply to another root.

Minimal layout:

```text
.git/
.gitignore
noxfile.py
src/scikit_build_core/build/metadata.py
tests/test_metadata.py
```

`.gitignore`:

```gitignore
tests/**/build/
```

The source build file should always be included: the ignore rule only matches `build/` directories below `tests/`.

Minimal Rust repro with `ignore = "0.4.25"`:

```rust
use std::path::Path;
use std::sync::atomic::{AtomicBool, Ordering};

fn source_build_file_was_seen(root: &Path) -> bool {
    let source_build_file = root.join("src/scikit_build_core/build/metadata.py");
    let seen = AtomicBool::new(false);

    let mut builder = ignore::WalkBuilder::new(root.join("src"));
    builder.current_dir(root);
    builder.standard_filters(true);
    builder.add(root.join("tests"));
    builder.add(root.join("noxfile.py"));
    builder.threads(12);

    builder.build_parallel().run(|| {
        Box::new(|entry| {
            if let Ok(entry) = entry {
                if entry.path() == source_build_file {
                    seen.store(true, Ordering::Relaxed);
                }
            }

            ignore::WalkState::Continue
        })
    });

    seen.load(Ordering::Relaxed)
}

fn main() {
    let root = Path::new(env!("CARGO_MANIFEST_DIR"));
    let mut seen = 0;
    let mut missing = 0;

    for _ in 0..100 {
        if source_build_file_was_seen(root) {
            seen += 1;
        } else {
            missing += 1;
        }
    }

    println!("seen={seen} missing={missing}");
}
```

Observed output on my machine:

```text
seen=14 missing=86
```

The same issue reproduces via ripgrep:

```sh
rg --threads 12 --files src tests noxfile.py
```

Across 100 identical runs, `src/scikit_build_core/build/metadata.py` was present 59 times and missing 41 times.

Control cases were stable:

```sh
rg --threads 1 --files src tests noxfile.py
# source file present 100/100

rg --threads 12 --files src
# source file present 100/100
```

So this looks specific to the parallel multi-root walk. The expected behavior is that `tests/**/build/` never excludes `src/scikit_build_core/build/metadata.py`, regardless of walk scheduling.

This is causing nondeterminism in ty too: it makes it so some files in the scikit-build-core project are sometimes checked and sometimes ignored by ty. The repro case in this issue is a simplified version of the shape in scikit_build_core.

Inconsistent output from `rg --threads 12 --files src tests noxfile.py`.

Rules should be applied consistently

jelle-openai:
Fixes #3419.

### DIFF
--- a/crates/ignore/src/dir.rs
+++ b/crates/ignore/src/dir.rs
@@ -91,7 +91,9 @@ struct IgnoreOptions {
 
 /// Ignore is a matcher useful for recursively walking one or more directories.
 #[derive(Clone, Debug)]
-pub(crate) struct Ignore(Arc<IgnoreInner>);
+pub(crate) struct Ignore {
+    inner: Arc<IgnoreInner>,
+}
 
 #[derive(Clone, Debug)]
 struct IgnoreInner {
@@ -153,22 +155,22 @@ struct IgnoreInner {
 impl Ignore {
     /// Return the directory path of this matcher.
     pub(crate) fn path(&self) -> &Path {
-        &self.0.dir
+        &self.inner.dir
     }
 
     /// Return true if this matcher has no parent.
     pub(crate) fn is_root(&self) -> bool {
-        self.0.parent.is_none()
+        self.inner.parent.is_none()
     }
 
     /// Returns true if this matcher was added via the `add_parents` method.
     pub(crate) fn is_absolute_parent(&self) -> bool {
-        self.0.is_absolute_parent
+        self.inner.is_absolute_parent
     }
 
     /// Return this matcher's parent, if one exists.
     pub(crate) fn parent(&self) -> Option<Ignore> {
-        self.0.parent.clone()
+        self.inner.parent.clone()
     }
 
     /// Create a new `Ignore` matcher with the parent directories of `dir`.
@@ -179,10 +181,10 @@ impl Ignore {
         &self,
         path: P,
     ) -> (Ignore, Option<Error>) {
-        if !self.0.opts.parents
-            && !self.0.opts.git_ignore
-            && !self.0.opts.git_exclude
-            && !self.0.opts.git_global
+        if !self.inner.opts.parents
+            && !self.inner.opts.git_ignore
+            && !self.inner.opts.git_exclude
+            && !self.inner.opts.git_global
         {
             // If we never need info from parent directories, then don't do
             // anything.
@@ -211,10 +213,10 @@ impl Ignore {
         let mut errs = PartialErrorBuilder::default();
         let mut ig = self.clone();
         for parent in parents.into_iter().rev() {
-            let mut compiled = self.0.compiled.write().unwrap();
+            let mut compiled = self.inner.compiled.write().unwrap();
             if let Some(weak) = compiled.get(parent.as_os_str()) {
                 if let Some(prebuilt) = weak.upgrade() {
-                    ig = Ignore(prebuilt);
+                    ig = Ignore { inner: prebuilt };
                     continue;
                 }
             }
@@ -223,13 +225,13 @@ impl Ignore {
             igtmp.is_absolute_parent = true;
             igtmp.absolute_base = Some(absolute_base.clone());
             igtmp.has_git =
-                if self.0.opts.require_git && self.0.opts.git_ignore {
+                if self.inner.opts.require_git && self.inner.opts.git_ignore {
                     parent.join(".git").exists() || parent.join(".jj").exists()
                 } else {
                     false
                 };
             let ig_arc = Arc::new(igtmp);
-            ig = Ignore(ig_arc.clone());
+            ig = Ignore { inner: ig_arc.clone() };
             compiled.insert(
                 parent.as_os_str().to_os_string(),
                 Arc::downgrade(&ig_arc),
@@ -251,13 +253,13 @@ impl Ignore {
         dir: P,
     ) -> (Ignore, Option<Error>) {
         let (ig, err) = self.add_child_path(dir.as_ref());
-        (Ignore(Arc::new(ig)), err)
+        (Ignore { inner: Arc::new(ig) }, err)
     }
 
     /// Like add_child, but takes a full path and returns an IgnoreInner.
     fn add_child_path(&self, dir: &Path) -> (IgnoreInner, Option<Error>) {
-        let check_vcs_dir = self.0.opts.require_git
-            && (self.0.opts.git_ignore || self.0.opts.git_exclude);
+        let check_vcs_dir = self.inner.opts.require_git
+            && (self.inner.opts.git_ignore || self.inner.opts.git_exclude);
         let git_type = if check_vcs_dir {
             dir.join(".git").metadata().ok().map(|md| md.file_type())
         } else {
@@ -267,44 +269,45 @@ impl Ignore {
             check_vcs_dir && (git_type.is_some() || dir.join(".jj").exists());
 
         let mut errs = PartialErrorBuilder::default();
-        let custom_ig_matcher = if self.0.custom_ignore_filenames.is_empty() {
-            Gitignore::empty()
-        } else {
-            let (m, err) = create_gitignore(
-                &dir,
-                &dir,
-                &self.0.custom_ignore_filenames,
-                self.0.opts.ignore_case_insensitive,
-            );
-            errs.maybe_push(err);
-            m
-        };
-        let ig_matcher = if !self.0.opts.ignore {
+        let custom_ig_matcher =
+            if self.inner.custom_ignore_filenames.is_empty() {
+                Gitignore::empty()
+            } else {
+                let (m, err) = create_gitignore(
+                    &dir,
+                    &dir,
+                    &self.inner.custom_ignore_filenames,
+                    self.inner.opts.ignore_case_insensitive,
+                );
+                errs.maybe_push(err);
+                m
+            };
+        let ig_matcher = if !self.inner.opts.ignore {
             Gitignore::empty()
         } else {
             let (m, err) = create_gitignore(
                 &dir,
                 &dir,
                 &[".ignore"],
-                self.0.opts.ignore_case_insensitive,
+                self.inner.opts.ignore_case_insensitive,
             );
             errs.maybe_push(err);
             m
         };
-        let gi_matcher = if !self.0.opts.git_ignore {
+        let gi_matcher = if !self.inner.opts.git_ignore {
             Gitignore::empty()
         } else {
             let (m, err) = create_gitignore(
                 &dir,
                 &dir,
                 &[".gitignore"],
-                self.0.opts.ignore_case_insensitive,
+                self.inner.opts.ignore_case_insensitive,
             );
             errs.maybe_push(err);
             m
         };
 
-        let gi_exclude_matcher = if !self.0.opts.git_exclude {
+        let gi_exclude_matcher = if !self.inner.opts.git_exclude {
             Gitignore::empty()
         } else {
             match resolve_git_commondir(dir, git_type) {
@@ -313,7 +316,7 @@ impl Ignore {
                         &dir,
                         &git_dir,
                         &["info/exclude"],
-                        self.0.opts.ignore_case_insensitive,
+                        self.inner.opts.ignore_case_insensitive,
                     );
                     errs.maybe_push(err);
                     m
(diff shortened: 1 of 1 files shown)


crates/ignore/src/dir.rs
```diff
@@ -211,8 +220,12 @@ impl Ignore {
         let mut errs = PartialErrorBuilder::default();
         let mut ig = self.clone();
         for parent in parents.into_iter().rev() {
+            let key = CompiledMatcherKey {
+                dir: parent.as_os_str().to_os_string(),
+                absolute_base: absolute_base.clone(),
+            };
             let mut compiled = self.0.compiled.write().unwrap();
-            if let Some(weak) = compiled.get(parent.as_os_str()) {
+            if let Some(weak) = compiled.get(&key) {
```
MichaReiser: I'm not very familiar with this code base, but this makes me wonder if it's still worth caching here. 

for `roots: /tmp/project/src, /tmp/project/tests`:

* Before: The entry for `/tmp/project` was shared
* After: Each root uses its own pre-built `Ignore`, making the caching here useless except for the case where someone calls `add_parents` twice with the exact same paths

The caching here also served two purposes:

* Avoid parsing the same ignore files multiple times 
* Deduplicate errors (because we only built the ignore once)

Now, the main issue why sharing the `Ignore` across base paths is incorrect is because we set `ignore.absolute_base` to `Some(absolute_base)`. 

I haven't traced through all the changes necessary but I think a better solution is to decouple `absolute_path` from what we store in the cache, e.g. by:

* Moving `absolute_path` from `IgnoreInner` to the `Ignore` struct (makes cloning more expensive)
* Use a different struct for caching

BurntSushi: Yeah I think @MichaReiser is right here. In particular, I think this solution will fall apart when there are a lot of roots. For example, `rg foo bar/**` where `bar` contains many paths. 

I'm on mobile so I don't have a good idea right now of what the simplest change is yet. But Micha's ideas at a very high level seem plausible.

jelle-openai: Thanks, I pushed a followup aligning with Micha's first suggestion. This largely restores the performance of the pre-existing code while still fixing the bug.

crates/ignore/src/dir.rs
```diff
@@ -91,7 +91,12 @@ struct IgnoreOptions {
 
 /// Ignore is a matcher useful for recursively walking one or more directories.
 #[derive(Clone, Debug)]
-pub(crate) struct Ignore(Arc<IgnoreInner>);
+pub(crate) struct Ignore(
+    Arc<IgnoreInner>,
+    // Parent matchers are cached independently of the root they are used
+    // from, but matching them still needs the current root to rewrite paths.
+    Option<Arc<PathBuf>>,
```
MichaReiser: If using a named field is too disruptive, can we document what `root` means in this context.

jelle-openai: Moved to named fields, wasn't too bad

crates/ignore/src/dir.rs
```diff
@@ -557,29 +575,40 @@ impl Ignore {
...
 
     /// Returns the first absolute path of the first absolute parent, if
     /// one exists.
     fn absolute_base(&self) -> Option<&Path> {
-        self.0.absolute_base.as_ref().map(|p| &***p)
+        self.absolute_base.as_ref().map(|p| &***p)
     }
 }
 
 /// An iterator over all parents of an ignore matcher, including itself.
```
MichaReiser: This comment here is now misplaced

MichaReiser (approved):

crates/ignore/src/dir.rs
jelle-openai: Thanks, moved

BurntSushi:
I got this benchmark from Codex that I think I believe. It finds a regression on the initial commit in this PR, but not on master or on the last commit in this PR. Do:

```
$ mkdir /dev/shm/ripgrep-pr3420
$ cd /dev/shm/ripgrep-pr3420
$ touch create
$ chmod +x create
$ $EDITOR create
$ cat create
#!/usr/bin/env bash

set -euo pipefail

D=./

mkdir -p "$D/.git"

for i in $(seq 1 2000); do
  printf 'ignored_%04d/**\n' "$i"
done > "$D/.gitignore"

for r in $(seq -w 0 49); do
  mkdir -p "$D/root$r/src"
  for f in $(seq -w 0 9); do
    printf 'needle-free content %s %s\n' "$r" "$f" \
      > "$D/root$r/src/file$f.txt"
  done
done
$ ./create
$ tree -a | wc -l
606
$ $ hyperfine --warmup 3 --runs 10 --ignore-failure '/path/to/master/target/release/rg -n NOMATCH root*' '/path/to/commit-13134/target/release/rg -n NOMATCH root*' '/path/to/commit-3e32/target/release/rg -n NOMATCH root*'

[.. snip ..]

Summary
  /tmp/rg-master-src/target/release/rg -n NOMATCH root* ran
    1.00 ± 0.26 times faster than /tmp/rg-3e32-src/target/release/rg -n NOMATCH root*
   25.51 ± 5.77 times faster than /tmp/rg-13134-src/target/release/rg -n NOMATCH root*
```

I find this quite promising!

crates/ignore/src/dir.rs
```diff
@@ -575,29 +587,46 @@ impl Ignore {
...
     /// one exists.
     fn absolute_base(&self) -> Option<&Path> {
-        self.inner.absolute_base.as_ref().map(|p| &***p)
+        self.absolute_base.as_ref().map(|p| &***p)
+    }
+}
+
+#[derive(Clone, Copy)]
+pub(crate) struct IgnoreRef<'a> {
+    inner: &'a IgnoreInner,
```
MichaReiser, line 602: We use a named field here to reduce churn in the diff

MichaReiser:
I reworded the commits. I also had codex write a few benchmarks and verified that none of them regress (beyond noise). We could consider adding those benchmarks to the repo, but I won't do so as part of this PR. 

@BurntSushi I think this is good to go

<details>

<summary>Benchmarks</summary>

```rust
/*!
This module benchmarks ignore file matching.
*/
#![feature(test)]

extern crate test;

use std::{
    fmt::Write as _,
    fs,
    path::{Path, PathBuf},
    sync::atomic::{AtomicUsize, Ordering},
    time::{SystemTime, UNIX_EPOCH},
};

use ignore::{WalkBuilder, WalkState};

const GITIGNORE_PATTERNS: usize = 100;
const DEEP_DIRS: usize = 64;
const LAYERED_DIRS: usize = 32;
const LAYERED_PARENTS: usize = 8;
const MANY_ROOTS: usize = 64;
const PARTIALLY_SHARED_GROUPS: usize = 8;
const PARTIALLY_SHARED_ROOTS_PER_GROUP: usize = 8;
const WIDE_DIRS: usize = 128;

struct TempDir(PathBuf);

impl TempDir {
    fn new() -> TempDir {
        let nanos =
            SystemTime::now().duration_since(UNIX_EPOCH).unwrap().as_nanos();
        let path = std::env::temp_dir()
            .join(format!("ignore-bench-{}-{nanos}", std::process::id(),));
        fs::create_dir(&path).unwrap();
        TempDir(path)
    }

    fn path(&self) -> &Path {
        &self.0
    }
}

impl Drop for TempDir {
    fn drop(&mut self) {
        fs::remove_dir_all(&self.0).unwrap();
    }
}

struct ParentGitignore {
    _temp_dir: TempDir,
    builder: WalkBuilder,
    roots: usize,
}

impl ParentGitignore {
    fn new(roots: usize) -> ParentGitignore {
        let gitignore = gitignore_patterns("ignored");
        ParentGitignore::with_paths(&gitignore, |project| {
            (0..roots).map(|i| project.join(format!("root-{i:03}"))).collect()
        })
    }

    fn empty(roots: usize) -> ParentGitignore {
        ParentGitignore::with_paths("", |project| {
            (0..roots).map(|i| project.join(format!("root-{i:03}"))).collect()
        })
    }

    fn partially_shared() -> ParentGitignore {
        let gitignore = gitignore_patterns("ignored");
        ParentGitignore::with_paths(&gitignore, |project| {
            (0..PARTIALLY_SHARED_GROUPS)
                .flat_map(|group| {
                    (0..PARTIALLY_SHARED_ROOTS_PER_GROUP).map(move |root| {
                        project
                            .join(format!("group-{group:03}"))
                            .join(format!("root-{root:03}"))
                    })
                })
                .collect()
        })
    }

    fn deep_kept() -> ParentGitignore {
        ParentGitignore::with_paths("ignored/**\n", |project| {
            let root = project.join("root");
            let mut current = root.clone();
            for i in 0..DEEP_DIRS {
                current = current.join(format!("dir-{i:03}"));
                fs::create_dir_all(&current).unwrap();
            }
            fs::write(current.join("file"), "").unwrap();
            vec![root]
        })
    }

    fn layered_parent_gitignores_misses() -> ParentGitignore {
        let gitignore = gitignore_patterns("ignored-project");
        ParentGitignore::with_paths(&gitignore, |project| {
            let mut parent = project.to_path_buf();
            for layer in 0..LAYERED_PARENTS {
                parent = parent.join(format!("parent-{layer:03}"));
                fs::create_dir_all(&parent).unwrap();
                fs::write(
                    parent.join(".gitignore"),
                    gitignore_patterns(&format!("ignored-{layer:03}")),
                )
                .unwrap();
            }
            let root = parent.join("root");
            for i in 0..LAYERED_DIRS {
                let dir = root.join(format!("dir-{i:03}"));
                fs::create_dir_all(&dir).unwrap();
                fs::write(dir.join("file"), "").unwrap();
            }
            vec![root]
        })
    }

    fn wide_kept() -> ParentGitignore {
        ParentGitignore::with_paths("ignored/**\n", |project| {
            let root = project.join("root");
            for i in 0..WIDE_DIRS {
                let dir = root.join(format!("dir-{i:03}"));
                fs::create_dir_all(&dir).unwrap();
                fs::write(dir.join("file"), "").unwrap();
            }
            vec![root]
        })
    }

    fn with_parent_matches(roots: usize) -> ParentGitignore {
        ParentGitignore::with_paths("root-*/ignored/\n", |project| {
            let paths = (0..roots)
                .map(|i| project.join(format!("root-{i:03}")))
                .collect::<Vec<_>>();
            for path in &paths {
                fs::create_dir_all(path.join("ignored")).unwrap();
                fs::create_dir_all(path.join("kept")).unwrap();
                fs::write(path.join("ignored/file"), "").unwrap();
                fs::write(path.join("kept/file"), "").unwrap();
            }
            paths
        })
    }

    fn with_paths(
        gitignore: &str,
        paths: impl FnOnce(&Path) -> Vec<PathBuf>,
    ) -> ParentGitignore {
        let temp_dir = TempDir::new();
        let project = temp_dir.path().join("project");
        fs::create_dir_all(project.join(".git")).unwrap();
        fs::write(project.join(".gitignore"), gitignore).unwrap();

        let paths = paths(&project);
        for path in &paths {
            fs::create_dir_all(path).unwrap();
        }

        let mut builder = WalkBuilder::new(&paths[0]);
        for path in &paths[1..] {
            builder.add(path);
        }
        builder.max_depth(Some(0)).git_global(false).git_exclude(false);
        ParentGitignore { _temp_dir: temp_dir, builder, roots: paths.len() }
    }

    fn walk(&self) {
        assert_eq!(self.roots, self.builder.build().count());
    }

    fn walk_parallel(&self, expected_entries: usize) {
        let entries = AtomicUsize::new(0);
        self.builder.build_parallel().run(|| {
            Box::new(|_| {
                entries.fetch_add(1, Ordering::Relaxed);
                WalkState::Continue
            })
        });
        assert_eq!(expected_entries, entries.load(Ordering::Relaxed));
    }

    fn walk_entries(&self, expected_entries: usize) {
        assert_eq!(expected_entries, self.builder.build().count());
    }
}

fn gitignore_patterns(prefix: &str) -> String {
    let mut gitignore = String::new();
    for i in 0..GITIGNORE_PATTERNS {
        writeln!(gitignore, "{prefix}/{i:03}/**").unwrap();
    }
    gitignore
}

#[bench]
fn add_parents_one_root(b: &mut test::Bencher) {
    let fixture = ParentGitignore::new(1);
```

</details>

BurntSushi (approved):
I love the fix! Rather elegant. Thank you @jelle-openai and @MichaReiser for getting this over the finish line. :-)

BurntSushi:
This PR is on crates.io in `ignore 0.4.26`.

MichaReiser:
> This PR is on crates.io in `ignore 0.4.26`.

Wow, this wask quick. Thank you

PR #3433, opened 2026-06-03
docs: update GUIDE to use `lexopt`

BurntSushi:
... we migrated off of `clap` long ago.


MichaReiser (approved):
TIL about `--type`. Nice feature

PR #3451, opened 2026-06-27
ignore: do not read .gitignore files above a Git boundary

### LINKED ISSUE #3434: rg reads outer repo gitignore even though it ignores it

ripgrep 15.1.0

features:+pcre2
simd(compile):+NEON
simd(runtime):+NEON

PCRE2 10.45 is available (JIT is available)

Homebrew.  Also tested on master.

mac latest

### Describe your bug.

rg is processing a file it doesn't use (gitignore).  This leads to confusing debugging because it creates globs it does not apply to the file walk.

```
mkdir outer; cd outer
echo bar > .gitignore
mkdir inner; cd inner
mkdir .git; echo foo > bar
```

rg correctly ignores the outer gitignore, since it's past the git boundary
```
/tmp/outer/inner $ rg foo
bar
1:foo
```

But it still processes it, and lists it.  Why?
```
/tmp/outer/inner $ rg --debug foo 2>&1 | grep -F .gitignore
rg: DEBUG|ignore::gitignore|crates/ignore/src/gitignore.rs:398: opened gitignore file: /private/tmp/outer/.gitignore
rg: DEBUG|globset|crates/globset/src/lib.rs:527: glob set patterns — literals: [], basenames: ["bar"], extensions: [], prefixes: [], suffixes: [], required_extensions: [], regexes: []
```
(the last line is a custom log I added)

ripgrep should not open outer repo gitignore when it's past git boundary.  For performance and also for easier debugging.

Nipan83:
### What does this PR do?

When searching inside a nested Git repo, ripgrep already **ignores** parent `.gitignore` rules above the repository boundary (`saw_git` in `matched_ignore`). It still **opened and parsed** those files while building absolute parent matchers in `Ignore::add_parents`, which shows up under `--debug` and wastes work (see #3434).

This change skips reading **git-specific** ignore files (`.gitignore` and `.git/info/exclude`) for parent directories strictly above the nearest VCS root on the search path (and when `require_git` is on and there is no VCS at all). **`.ignore` and custom ignore filenames are still read** from those parents, matching existing behavior for non-git ignore files.

Parent matchers that omit git rules are **not** inserted into the path-only parent matcher cache, so another search root that legitimately needs that parent’s `.gitignore` does not reuse an incomplete matcher.

### How was this tested?

- New unit test: `absolute_parent_skips_gitignore_above_git_boundary` in `crates/ignore/src/dir.rs`
- `cargo test -p ignore --lib`
- Integration filters: `ignore_git_parent`, `ignore_git_parent_stop`, `ignore_git_parent_stop_file`, `ignore_ripgrep_parent_no_stop`, `no_parent_ignore_git`

### Issue

Fixes #3434

### AI use

Coding assistance was used to implement and verify this change. I have reviewed the diff and can explain and defend it. Per the project AI policy, I will answer review questions myself rather than pasting AI-generated replies.

### DIFF
--- a/crates/ignore/src/dir.rs
+++ b/crates/ignore/src/dir.rs
@@ -93,11 +93,13 @@ struct IgnoreOptions {
 #[derive(Clone, Debug)]
 pub(crate) struct Ignore {
     inner: Arc<IgnoreInner>,
-    // Parent matchers are cached independently of the path being walked, but
-    // matching them still needs the canonicalized path originally passed to
-    // `add_parents`. For example, when walking `/tmp/project/src`, parent
-    // matchers use `/tmp/project/src` to rewrite `foo.py` before matching it
-    // against ignore files from `/tmp/project` and its ancestors.
+    // Parent matchers (`IgnoreInner`) are cached by directory path and reused
+    // across multiple search roots, but path rewriting for absolute parents
+    // must use the canonicalized path originally passed to `add_parents` for
+    // *this* root. Storing that base on `Ignore` (not in the cached inner)
+    // keeps matching correct when several roots share parent matchers—e.g.
+    // `rg pat src tests` must rewrite `src/invalid` relative to `…/src`, not
+    // whatever root was prepared first. See issues #3376, #3419, and #3320.
     absolute_base: Option<Arc<PathBuf>>,
 }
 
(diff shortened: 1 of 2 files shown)


crates/ignore/src/dir.rs
```diff
@@ -239,10 +278,13 @@ impl Ignore {
...
-                parent.as_os_str().to_os_string(),
-                Arc::downgrade(&ig_arc),
-            );
+            if read_git_ignore_files {
+                let mut compiled = self.inner.compiled.write().unwrap();
+                compiled.insert(
+                    parent.as_os_str().to_os_string(),
+                    Arc::downgrade(&ig_arc),
+                );
+            }
```
BurntSushi, line 287: I think this whole thing makes me a bit nervous. In particular:

* Caching here is reduced in some cases.
* The mutex guarding `self.inner.compiled` is now acquired twice instead of once.
* This significantly increases complexity.

In other words, I'm not sure this change is actually worth it unfortunately.

BurntSushi:
I will take the tests added in this PR though, thank you! In the future, that probably should be a separate PR.

PR #3452, opened 2026-06-28
zsh completion updates

okdana:
some over-due completion updates. see commit messages for explanations. feel free to squash, reword, etc

ps: re: _numbers: i think it's been long enough that it's ok to rely on this, but if you're worried about bc it would be easy to add a compatibility shim


BurntSushi (approved):
Lovely! The commit breakdown was super helpful. And you followed my conventions, so I'm happy to merge these as-is. :-)

PR #3458, opened 2026-07-01
feat[ignore]: Always include depth in errors from WalkDir

### LINKED ISSUE #3453: [ignore] Include depth in WithPath error.

https://github.com/sharkdp/fd/issues/1017 would be significantly easier to fix if `ignore::Error::WithPath` included the depth, relative to the root of the path the error occurred on.  Specifically, in the event we get an error for a broken symlink, it allows us to know the depth of the broken symlink, which informs whether or not we should output anything for it.

tmccombs:
Include the depth in I/O errors from WalkDir as well, so that the consumer can handle the error differently depending on depth. For example, this would allow ignore broken symlinks that don't match certain depth criteria.

Fixes: #3453

### DIFF
--- a/crates/ignore/src/lib.rs
+++ b/crates/ignore/src/lib.rs
@@ -294,7 +294,10 @@ impl Error {
             };
         }
         let path = err.path().map(|p| p.to_path_buf());
-        let mut ig_err = Error::Io(std::io::Error::from(err));
+        let mut ig_err = Error::WithDepth {
+            depth,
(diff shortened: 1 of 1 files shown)


BurntSushi (approved):
OK let's see what happens!

Note that I would suggest writing your code in a way that doesn't care about the nesting order.

(I want to re-iterate how terrible of a design this is for an error. I don't know what I was thinking.)

BurntSushi:
This PR is on crates.io in `ignore 0.4.27`.

PR #3461, opened 2026-07-04
feat[ignore]: Add path() to error

tmccombs:
Add a path() method to Error, which allows getting the path without needing to pattern match on WithPath.


BurntSushi:
Hmmm I'm not so sure about this one. This is omitting `Error::Io` and `Error::Loop` and `Error::Partial`. In the latter two cases, it's not clear what `Path` should be provided. I think this one might be better as a helper function defined as needed for your specific use case.

PR #3462, opened 2026-07-05
Add `--in`/`--in0` options to read input paths from files or stdin

### LINKED ISSUE #273: Implement such `--files-from` option

### LINKED ISSUE #3459: Read input paths from a file or stdin
I'd like to suggest a new option, such as `--input=INPUTFILE` or `--input-file=INPUTFILE` (name TBD, the shorter the better), very similar to the existing `--file=PATTERNFILE`, which would read the list of files or directories to be searched from `INPUTFILE`. The value `-` would mean the list should be read from stdin.

Here are some use cases:

- Search from an existing (potentially long) list of files, which was previously generated, and could be reused across several `rg` calls.

- Search for files containing multiple patterns, but not necessarily on the same line, by chaining `rg --files-with-matches` calls.

  As an example: searching for files containing `foo`, `bar` and `baz`:
  ```bash
  rg foo -l | rg bar --input=- -l | rg baz --input=-
  ```
  The output would be a nicely presented list of `baz` matches, with headings. This also shows why a short option name would be useful.

  This would be an equivalent to the following on Linux:
  ```bash
  rg -0l foo | xargs -0 rg bar -0l | xargs -0 rg baz
  ```

  But my main usage is on Windows, which doesn't have `xargs`, so the same command on PowerShell would be:

  ```powershell
  rg baz (rg -l bar (rg -l foo))
  ```

  Not only is it less readable (IMO), but it also doesn't support *long lists of files* to search from, as command-line arguments are limited to 32 kb on Windows:

  Image

- Probably some other usages from #875

Of course, `--input=-` would have to be incompatible with `--file=-` and stdin input.

I know you don't like adding more options, and for full disclosure I already have a script which handles this, but I believe this would be a useful feature for many people.

If you'd like, I think I can write and submit a *simple* implementation in `rg::flags::hiargs::Paths` which would be based on `rg::flags::hiargs::Patterns` (which handles `--file`). The main issue I see with this approach is that stdin would need to be drained before the search begins, but this is already the case with `xargs` anyway. I just wanted to check with you first if that's something you'd be willing to accept.

ltrzesniewski:
This PR adds `--in` and `--in0` command-line options, which read files or directories to be searched from their arguments, similar to as if they were specified on the command line. ~If the file contains relative paths, their paths are resolved relatively to the input file.~ The value `-` means the list of files should be read from stdin.

e.g. having a file `files.txt` with the following content:

```
foo.txt
bar.txt
```

Then running `rg pattern --in files.txt` is *equivalent* to running `rg pattern foo.txt bar.txt`.

`--in` reads from a text file (supporting LF and CRLF line terminators), while `--in0` reads from a null-delimited file. Both flags can be used multiple times. ~Empty entries are ignored.~

These features are designed for reusing input files and ripgrep calls chaining, such as:

```bash
$ rg foo -l | rg bar --in=- -l | rg baz --in=-
$ rg foo -l0 | rg bar --in0=- -l0 | rg baz --in0=-
```

These commands will find all the files that contain `foo`, `bar` and `baz`, and will highlight all of the `baz` matches, with headings and line numbers. The second one is equivalent to:

```bash
rg -0l foo | xargs -0 rg bar -0l | xargs -0 rg baz
```

Implementation notes:
- ~Currently, the paths read from the input file are treated *exactly* as if they were specified on the command line, which means relative paths are relative to the *current working directory*, not the directory of the input file. This may be surprising behavior, but is consistent with the way `xargs` works. Please tell me if you'd prefer for the paths to be relative to the input file instead (or the CWD when stdin is used). I really don't know which one is better TBH, so I defer to your opinion.~
  **Decision:** interpret everything as if it were a command-lime argument.
- This implies the exit code will always be 2 if there's any non-existent path in an input file.
- There is no data streaming: all the paths are read into memory before searching. This keeps the implementation relatively simple.
- The implementation is loosely based on `--file`, but I was reluctant to create a `grep_cli::input` module similar to `grep_cli::pattern` for this kind of functionality, so I put everything in `rg::flags::hiargs::Paths::read_from_input`. This feels a bit messy/unidiomatic to me TBH, don't hesitate to suggest some other approach if you think it would be better.
- There is a little mixup between `std::io::Result` and `anyhow::Result` in `read_from_input` due to the signatures `bstr::io::BufReadExt` expects. I'm not sure if that's ok or not.
- If a file passed to `--in` contains a NUL byte, it's almost certainly an error, and is reported as such.

Related issues:
- Closes #3459
- Closes #273
- #875

### DIFF
--- a/crates/core/flags/complete/rg.zsh
+++ b/crates/core/flags/complete/rg.zsh
@@ -175,6 +175,10 @@ _rg() {
     "--no-ignore-files[don't respect --ignore-file flags]"
     $no'--ignore-files[respect --ignore-file files]'
 
+    + input
+    '*--in=[specify text file containing paths to search]: :_files'
+    '*--in0=[specify binary file containing NUL-terminated paths to search]: :_files'
+
     + '(invert-match)'
     {-v,--invert-match}'[invert matching]'
     $no"--no-invert-match[don't invert matching]"
--- a/crates/core/flags/defs.rs
+++ b/crates/core/flags/defs.rs
@@ -26,8 +26,8 @@ use crate::flags::{
     lowargs::{
         BinaryMode, BoundaryMode, BufferMode, CaseMode, ColorChoice,
         ContextMode, EncodingMode, EngineChoice, GenerateMode, IndexMode,
-        LoggingMode, LowArgs, MmapMode, Mode, PatternSource, SearchMode,
-        SortMode, SortModeKind, SpecialMode, TypeChange,
+        InputSource, LoggingMode, LowArgs, MmapMode, Mode, PatternSource,
+        SearchMode, SortMode, SortModeKind, SpecialMode, TypeChange,
     },
 };
 
@@ -83,6 +83,8 @@ pub(super) const FLAGS: &[&dyn Flag] = &[
     &IgnoreCase,
     &IgnoreFile,
     &IgnoreFileCaseInsensitive,
+    &In,
+    &In0,
     &IncludeZero,
     &Index,
     &IndexCrud,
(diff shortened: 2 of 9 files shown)


ltrzesniewski:
Actually I think I prefer the paths to be relative to the input file, as this makes more sense to me. I pushed a second commit which changes the behavior. Feel free to drop it if you'd rather have the `xargs` behavior.

crates/core/flags/defs.rs
```diff
@@ -3325,6 +3327,176 @@ fn test_ignore_file_case_insensitive() {
...
+    fn doc_short(&self) -> &'static str {
+        r"Read paths to search from the given text file."
+    }
+    fn doc_long(&self) -> &'static str {
+        r"
+Includes paths to search from the given file, with one path per line.
+.sp
+When this flag is used multiple times or in combination with the \flag{in0}
+flag, then all provided files are searched in addition to the paths in the
+positional arguments.
```
BurntSushi: I think this should be a little more specific. Something like:

```
When this flag is used, it behaves as if the file paths in the file provided
were passed as positional arguments in the same order as they are
in the file and relative to other positional arguments.
```

crates/core/flags/defs.rs
```diff
@@ -3325,6 +3327,176 @@ fn test_ignore_file_case_insensitive() {
...
+.sp
+Empty lines are ignored, and newlines (both LF and CRLF) are not counted as
```
BurntSushi: `find` does not ignore empty lines and instead errors. I think we should do the same here. The one exception is that we should treat this file as line (or NUL) _terminated_, not delimited. That is, if the file has a trailing newline or NUL, then that's fine and doesn't indicate the start of another file path if nothing follows it.

(I note that empty lines are not ignored with ripgrep's `-f/--file` flag. They are treated as an empty string, which creates a regex that matches at all positions.)

For this flag, I think an empty line is just something we should error on. It's _also_ the conservative choice. We can change an error into something else that isn't an error later, but we can't do the reverse.

crates/core/flags/defs.rs
```diff
@@ -3325,6 +3327,176 @@ fn test_ignore_file_case_insensitive() {
...
+part of the path. The rest of the path name is decoded as UTF-8 on Windows
+and taken as-is on Unix systems.
+Use the \flag{in0} flag to use \fBNUL\fP as separator.
```
BurntSushi: I would use the word "terminator" instead.

crates/core/flags/defs.rs
```diff
@@ -3325,6 +3327,176 @@ fn test_ignore_file_case_insensitive() {
...
+Includes paths to search from the given file, separated by \fBNUL\fP bytes.
+.sp
+When this flag is used multiple times or in combination with the \flag{in}
+flag, then all provided files are searched in addition to the paths in the
+positional arguments.
+.sp
+Empty entries are ignored. The path names are decoded as UTF-8 on Windows
+and taken as-is on Unix systems.
+.sp
+When \fIINPUTFILE\fP is \fB-\fP, then \fBstdin\fP will be read for the files.
```
BurntSushi: Same comments apply as above.

crates/core/flags/defs.rs
```diff
@@ -3325,6 +3327,176 @@ fn test_ignore_file_case_insensitive() {
...
+    let args =
+        parse_low_raw(["--in=foo", "--in0", "bar", "--in", "baz"]).unwrap();
+    assert_eq!(
+        vec![
+            InputSource::TextFile(PathBuf::from("foo")),
+            InputSource::BinaryFile(PathBuf::from("bar")),
+            InputSource::TextFile(PathBuf::from("baz")),
+        ],
+        args.inputs
+    );
```
BurntSushi, line 3560: Can you add tests that mingle `--in`, `--in0` and other positional paths?

crates/core/flags/hiargs.rs
```diff
@@ -1137,6 +1143,123 @@ impl Paths {
...
+        // - Files or stdin: needs to handle `stdin_consumed`
+        // - Unix or Windows: different rules for `PathBuf`
+
+        let (path, flag) = match input {
+            InputSource::TextFile(path) => (path, "--in"),
+            InputSource::BinaryFile(path) => (path, "--in0"),
+        };
+
+        #[cfg(not(any(unix, windows)))]
+        anyhow::bail!("{flag} is not supported on this platform");
```
BurntSushi: No, this can't be...

crates/core/flags/hiargs.rs
```diff
@@ -1137,6 +1143,123 @@ impl Paths {
...
+            }
+            #[cfg(unix)]
+            {
+                // Unix allows any byte sequence in a file name except
+                // for NUL, so we can just use the raw bytes.
+                use std::{ffi::OsStr, os::unix::ffi::OsStrExt};
+                let _ = path; // Not used on Unix
+                output.push(PathBuf::from(OsStr::from_bytes(item)));
+            }
+            #[cfg(windows)]
```
BurntSushi: I don't see any Windows specific code here, so I think you can just use `cfg(not(unix))`.

crates/core/flags/hiargs.rs
```diff
@@ -1137,6 +1143,123 @@ impl Paths {
...
+                        "error reading {} {}: {}",
+                        flag,
+                        path.display(),
+                        err
+                    ))
+                })?;
+                output.push(PathBuf::from(s));
+            }
+            Ok(())
+        };
```
BurntSushi, line 1272: This entire routine can almost be simplified to just [`ByteSlice::to_path`](https://docs.rs/bstr/latest/bstr/trait.ByteSlice.html#method.to_path).

crates/core/flags/hiargs.rs
```diff
@@ -1068,22 +1069,27 @@ struct Paths {
...
         let mut paths = Vec::with_capacity(low.positional.len());
+        for input in low.inputs.drain(..) {
+            Self::read_from_input(state, &input, &mut paths)?
+        }
         for osarg in low.positional.drain(..) {
             let path = PathBuf::from(osarg);
             if state.stdin_consumed && path == Path::new("-") {
                 anyhow::bail!(
-                    "error: attempted to read patterns from stdin \
+                    "error: attempted to read patterns or inputs from stdin \
```
BurntSushi: "or input file paths"

crates/core/flags/hiargs.rs
```diff
@@ -1068,22 +1069,27 @@ struct Paths {
...
         // We require a `&Patterns` even though we don't use it to ensure that
-        // patterns have already been read from LowArgs. This let's us safely
+        // patterns have already been read from LowArgs. This lets us safely
         // assume that all remaining positional arguments are intended to be
         // file paths.
 
         let mut paths = Vec::with_capacity(low.positional.len());
+        for input in low.inputs.drain(..) {
+            Self::read_from_input(state, &input, &mut paths)?
+        }
```
BurntSushi: I think we want users to be able to mingle `--in` and `--in0` with positional paths and have them processed in the same order as given. This means, I think, that `low.positional` needs to become an `InputSource` and `InputSource` needs to grow a third single-file-path variant. To do this, I think this needs to be tweaked:

https://github.com/BurntSushi/ripgrep/blob/48b0c795f4feb37343b2832d991c5c6a3900c08a/crates/core/flags/parse.rs#L259-L228

crates/core/flags/hiargs.rs
```diff
@@ -1137,6 +1143,123 @@ impl Paths {
...
+            input: &InputSource,
+            mut for_each_path: impl FnMut(&[u8]) -> std::io::Result<()>,
+        ) -> anyhow::Result<()> {
+            let mut reader = std::io::BufReader::new(read);
+            match input {
+                InputSource::TextFile(path) => {
+                    reader.for_byte_line(|line| {
+                        if line.contains(&0u8) {
+                            return Err(std::io::Error::other(format!(
+                                "--in {}: file contains a NUL byte, \
```
BurntSushi: Error messages should just start with a file path. The `--in` can come after the colon.

BurntSushi: Also, `anyhow::bail!` should be used here. And anywhere else you're building an `io::Error` like this.

crates/core/flags/lowargs.rs
```diff
@@ -637,6 +638,15 @@ pub(crate) enum MmapMode {
     Never,
 }
 
+/// Represents a source of input paths to be searched.
+#[derive(Debug, Eq, PartialEq)]
+pub(crate) enum InputSource {
+    /// A text file with newline-separated paths. Comes from the `--in` flag.
+    TextFile(PathBuf),
+    /// A text file with NUL-separated paths. Comes from the `--in0` flag.
+    BinaryFile(PathBuf),
```
BurntSushi: These should be `LineTerminated` and `NulTerminated`.

tests/feature.rs
```diff
@@ -1172,3 +1172,143 @@ rgtest!(stop_on_nonmatch, |dir: Dir, mut cmd: TestCommand| {
...
+);
+
+// See: https://github.com/BurntSushi/ripgrep/issues/3459
+rgtest!(
+    input_from_in0_stdin_consumed_by_file,
+    |_dir: Dir, mut cmd: TestCommand| {
+        cmd.arg("--file").arg("-").arg("--in0").arg("-").arg("match");
+        cmd.assert_exit_code(2);
+    }
+);
```
BurntSushi: There are enough tests here that this warrants its own module.

Can you include tests that mingle `--in`, `--in0` _and_ positional file paths?

BurntSushi (changes requested):
Nice work for an initial pass!

crates/core/flags/complete/rg.zsh
```diff
@@ -174,6 +174,10 @@ _rg() {
     "--no-ignore-files[don't respect --ignore-file flags]"
     $no'--ignore-files[respect --ignore-file files]'
 
+    + input # Read input paths from file options
```
okdana: you can get rid of the group comment, i've removed them all now

crates/core/flags/complete/rg.zsh
```diff
@@ -174,6 +174,10 @@ _rg() {
...
+    '*--in=[text file containing paths to search]: :_files'
```
okdana: these should both start with 'specify ...'

crates/core/flags/hiargs.rs
ltrzesniewski: The issue I have with this is that `for_byte_line`/`for_byte_record` expect a function which returns `io::Result<bool>`. Maybe I'm stupid, but I don't see a way to use them with `anyhow`...

BurntSushi: Oh whoops, that's my bad. You aren't stupid. Thanks for pushing back. :-)

ltrzesniewski:
Thank you for your reviews @BurntSushi and @okdana!

I took all your comments into account and rebased the PR. This version is basically a major rewrite, but it feels much better and more idiomatic to me thanks to your insights. 🙂

ltrzesniewski:
I just rebased again because I noticed there were conflicts, and a new test needed an update.

PR #3464, opened 2026-07-07
feat[ignore]: Add depth to more errors

tmccombs:
Add depth to the Error objects in more cases where we know the appropriate depth.


BurntSushi (approved):
Thank you!

PR #3465, opened 2026-07-08
update AI policy link to point to the one in the ripgrep project

ClementTsang:
Noticed this while I was getting inspiration for my own AI policy document, doesn't _really_ change much but eh might as well point to the right url.


BurntSushi:
Thank you!

PR #3472, opened 2026-07-16
ignore: add incremental checking

BurntSushi:
This adds a new increment checker for ignore files, overrides and file
type selection. This effectively makes it possible to check individual
file paths as to whether they are ignored or not by any ignore file in
the directory tree (including parent gitignores). This is particularly
useful in the context of implementing a file watcher, where one wants to
make a decision about whether a newly added file should be included
*without* doing a full re-traversal of the directory tree.

This implementation reuses pretty much all of the existing gitignore
logic without making any changes to it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,7 @@ jobs:
         include:
         - build: pinned
           os: ubuntu-latest
-          rust: 1.85.0
+          rust: 1.96.0
         - build: stable
           os: ubuntu-latest
           rust: stable
--- a/crates/ignore/src/incremental.rs
+++ b/crates/ignore/src/incremental.rs
@@ -1,0 +1,1286 @@
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+    sync::OnceLock,
+};
+
+use crate::{
+    Error, Match, PartialErrorBuilder, dir::Ignore, pathutil::is_hidden_path,
+};
+
+/// A cached matcher for checking paths against hierarchical ignore files.
+///
+/// An `IncrementalIgnore` is built from a [`crate::WalkBuilder`]. Unlike a
+/// recursive walk, it can check individual paths while still respecting the
+/// ignore files in every relevant parent directory. Matchers for directories
+/// are compiled on first use and then retained for later queries.
+/// Each matcher corresponds to exactly one root configured on the builder,
+/// and paths passed to it are interpreted relative to that root.
+/// A matcher for the special `-` root representing standard input is inert
+/// and always returns a non-match.
+///
+/// The matcher checks path-based filters in the same precedence order as
+/// a traversal. This includes glob overrides, `.ignore`, `.gitignore`,
+/// `.git/info/exclude`, global and explicitly added ignore files, custom
+/// ignore file names and file type selections. It does not apply filters that
+/// require a directory entry or other traversal state, such as custom entry
+/// predicates. Hidden-file detection, minimum and maximum depth limits and the
+/// maximum file size are applied.
+///
+/// A matcher is a snapshot at directory granularity. Once the ignore files in
+/// a directory have been loaded, edits to those files are not observed. Build
+/// a new matcher to reload them.
+///
+/// # Warning
+///
+/// The incremental path checking here necessarily needs to do a lot more work
+/// per path matched. Callers should _not_ use this to run directory traversal.
+/// This is intended to avoid the work of re-traversing an entire directory
+/// tree when only a few changes are detected. (For example, in response to
+/// file additions or deletions.)
+///
+/// # Example
+///
+/// ```rust,no_run
+/// use ignore::WalkBuilder;
+///
+/// let mut builder = WalkBuilder::new(".");
+/// builder.add_custom_ignore_filename(".rgignore");
+/// let mut matchers = builder.build_matchers();
+/// let matcher = &mut matchers[0];
+///
+/// if matcher.matched("src/generated.rs", false).is_ignore() {
+///     println!("ignored");
+/// }
+/// ```
+#[derive(Clone, Debug)]
+pub struct IncrementalIgnore {
+    /// The root exactly as it was given to `WalkBuilder`.
+    root: PathBuf,
+    /// The normalized root used only by the opt-in normalization routine.
+    normalized_root: OnceLock<Option<PathBuf>>,
+    /// The matcher for the configured root directory, loaded on first use.
+    ignore: RootIgnore,
+    /// Directory paths relative to `root`, excluding the root itself.
+    dirs: HashMap<PathBuf, CachedDir>,
+    /// Options for additional filtering beyond gitignore.
+    options: IncrementalIgnoreOptions,
+}
+
+/// The options for a matcher, mostly meant to duplicate as much as we can from
+/// `WalkParallel`.
+#[derive(Clone, Debug)]
+pub(crate) struct IncrementalIgnoreOptions {
+    pub(crate) min_depth: Option<usize>,
+    pub(crate) max_depth: Option<usize>,
+    pub(crate) max_filesize: Option<u64>,
+    pub(crate) hidden: bool,
+    pub(crate) follow_links: bool,
+}
+
+#[derive(Clone, Debug)]
+enum RootIgnore {
+    Unloaded(Ignore),
+    Loaded(Ignore),
+    NotDirectory,
+    Stdin,
+}
+
+/// Cached traversal state for a directory relative to the configured root.
+///
+/// The presence of an entry means that the directory and every ancestor
+/// between it and the root have already been checked.
+#[derive(Clone, Debug)]
+enum CachedDir {
+    /// The directory may be descended into. The matcher includes the ignore
+    /// rules loaded through this directory and is therefore the matcher to use
+    /// for its children.
+    Allowed(Ignore),
+    /// The directory may not be descended into because it was ignored by a
+    /// path rule or hidden-file filtering. Every descendant is consequently
+    /// ignored, and ignore files inside this directory are not loaded.
+    Ignored,
+}
+
+impl IncrementalIgnore {
+    pub(crate) fn new(
+        root: PathBuf,
+        ignore: Ignore,
+        options: IncrementalIgnoreOptions,
+    ) -> IncrementalIgnore {
+        // File traversal special cases `-` to search stdin, so we recognize
+        // it here for completeness too. In particular, we really want
+        // `WalkBuilder::build_matchers` to return a matcher for every root,
+        // even when it's a simple file (handled automatically) or when it's
+        // stdin (necessarily special cased).
+        //
+        // If callers need to search a file or directory named `-`, then they
+        // can use `./-`. As is the case for file traversal too.
+        let ignore = if root == Path::new("-") {
+            RootIgnore::Stdin
+        } else {
+            RootIgnore::Unloaded(ignore)
+        };
+        IncrementalIgnore {
+            root,
+            normalized_root: OnceLock::new(),
+            ignore,
+            dirs: HashMap::new(),
+            options,
+        }
+    }
+
+    /// Return the root that paths matched by this matcher are relative to.
+    pub fn root(&self) -> &Path {
+        &self.root
+    }
+
+    /// Normalize `path` and return it relative to this matcher's root.
+    ///
+    /// This returns `None` when `path` cannot be made absolute or
+    /// when it is known to be outside this matcher's root. Unlike
+    /// [`IncrementalIgnore::matched`], this performs absolute path conversion,
+    /// lexical normalization and allocation. It is intended as an opt-in
+    /// convenience for callers that do not already have root-relative paths.
+    ///
+    /// Note that `.` is interpreted relative to the process level current
+    /// working directory. It is _not_ interpreted relative to the root of
+    /// this matcher.
+    ///
+    /// Note also that this may reject paths that only differ in casing. For
+    /// example, if the root path for this matcher is `/FOO` but the provided
+    /// path is `/foo/bar`, then this may return `None`. Callers must ensure
+    /// casing is consistent between the path provided and the root path for
+    /// this matcher.
+    pub fn normalize<P: AsRef<Path>>(&self, path: P) -> Option<PathBuf> {
+        if matches!(self.ignore, RootIgnore::Stdin) {
+            return None;
+        }
+        let path = normalize_absolute(path.as_ref())?;
+        let root = self
+            .normalized_root
+            .get_or_init(|| normalize_absolute(&self.root))
+            .as_ref()?;
+        path.strip_prefix(root).ok().map(Path::to_path_buf)
+    }
+
+    /// Match a root-relative path against ignore files in its directory and
+    /// all relevant parent directories.
+    ///
+    /// `is_dir` should be true when `path` should be matched as a directory.
+    ///
+    /// For the return value, use [`IncrementalMatch::is_ignore`],
+    /// [`IncrementalMatch::is_whitelist`] or [`IncrementalMatch::is_none`] to
+    /// inspect it.
+    ///
+    /// Matchers for previously unseen directories are loaded and cached during
+    /// this call. Errors encountered while loading ignore files are logged. To
+    /// receive those errors, use [`IncrementalIgnore::matched_with_errors`].
+    ///
+    /// `path` must be relative to this matcher's root and must not contain a
+    /// parent directory (`..`) component. Behavior is unspecified when these
+    /// preconditions are violated. Callers with an absolute path or with a
+    /// path containing `.` or `..` may use [`IncrementalIgnore::normalize`]
+    /// to get a path satisfying these preconditions. In all cases, a relative
+    /// path is *assumed* to be relative to the root of this ignore matcher.
+    ///
+    /// In general, it is intended that callers doing recursive directory
+    /// traversal on the root of this matcher may provide relative paths to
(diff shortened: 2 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         include:
         - build: pinned
           os: ubuntu-latest
-          rust: 1.85.0
+          rust: 1.96.0
```
MichaReiser, line 56: Do we have/need a CI job that checks whether the library crates like `ignore` compile with rust version X (I don't know what the MSRV is, or is it intentional that there's none?)

crates/ignore/src/pathutil.rs
```diff
@@ -2,58 +2,63 @@ use std::{ffi::OsStr, path::Path};
...
+/// ## All other platforms
+///
+/// This only returns true if the base name of the path starts with a `.`.
+pub(crate) fn is_hidden_entry(dent: &DirEntry) -> bool {
+    #[cfg(unix)]
+    fn imp(dent: &DirEntry) -> bool {
+        use std::os::unix::ffi::OsStrExt;
+
+        if let Some(name) = file_name(dent.path()) {
+            name.as_bytes().get(0) == Some(&b'.')
```
MichaReiser: Nit: You should still be able to use `starts_with` here

crates/ignore/src/pathutil.rs
```diff
@@ -2,58 +2,63 @@ use std::{ffi::OsStr, path::Path};
...
+        // This looks like we're doing an extra stat call, but on Windows, the
+        // directory traverser reuses the metadata retrieved from each directory
+        // entry and stores it on the DirEntry itself. So this is "free."
+        if let Ok(md) = dent.metadata() {
+            if file::is_hidden(md.file_attributes() as u64) {
+                return true;
+            }
+        }
+        if let Some(name) = file_name(dent.path()) {
+            name.to_str().map(|s| s.starts_with(".")).unwrap_or(false)
```
MichaReiser: What's the reason that we use `to_str` here and not `as_bytes`?

crates/ignore/src/incremental.rs
```diff
@@ -0,0 +1,1041 @@
...
+};
+
+/// A cached matcher for checking paths against hierarchical ignore files.
+///
+/// An `IncrementalIgnore` is built from a [`crate::WalkBuilder`]. Unlike a
+/// recursive walk, it can check individual paths while still respecting the
+/// ignore files in every relevant parent directory. Matchers for directories
+/// are compiled on first use and then retained for later queries.
+/// Each matcher corresponds to exactly one root configured on the builder,
+/// and paths passed to it are interpreted relative to that root.
```
MichaReiser, line 18: This may become clear while reviewing but one question I had when reading this is how nested paths are handled. E.g. what if we have `src` and `src/tests`

crates/ignore/src/incremental.rs
```diff
@@ -0,0 +1,1041 @@
...
+    /// This returns `None` when `path` cannot be made absolute or
+    /// when it is known to be outside this matcher's root. Unlike
+    /// [`IncrementalIgnore::matched`], this performs absolute path conversion,
+    /// lexical normalization and allocation. It is intended as an opt-in
+    /// convenience for callers that do not already have root-relative paths.
+    ///
+    /// Note that `.` is interpreted relative to the process level current
+    /// working directory. It is _not_ interpreted relative to the root of
+    /// this matcher.
+    pub fn normalize<P: AsRef<Path>>(&self, path: P) -> Option<PathBuf> {
```
MichaReiser, line 155: Is the intention here that an API consumer first calls `normalize` and then `matched` on the returned path? 

A newtype wrapper would sort of be nice to encode the "is normalized" (it also has a `from_normalized(Path)` method) but I assume you decided against it because it is a lot of ceremony because it probably requires a `PathBuf` and `Path` version

crates/ignore/src/incremental.rs
```diff
@@ -0,0 +1,1041 @@
...
+    ///
+    /// The empty path represents the explicitly configured root and also
+    /// returns non-match, consistent with recursive traversal where a root is
+    /// always treated as being at depth zero.
+    pub fn matched<P: AsRef<Path>>(
+        &mut self,
+        path: P,
+        is_dir: bool,
+    ) -> IncrementalMatch {
+        let (matched, err) = self.matched_with_errors(path, is_dir);
```
MichaReiser, line 199: Should this have a debug assertion that `path` is indeed a prefix of self.normalized_root`.

crates/ignore/src/incremental.rs
```diff
@@ -0,0 +1,1041 @@
...
+    /// `path` must be relative and must not contain a parent directory (`..`)
+    /// component. A non-match is returned when any of these preconditions
+    /// are violated. Callers with an absolute path or with a path containing
+    /// `.` or `..` may use [`IncrementalIgnore::normalize`] to get a path
+    /// satisfying these preconditions. In all cases, a relative path is
+    /// *assumed* to be relative to the root of this ignore matcher.
+    ///
+    /// In general, it is intended that callers doing recursive directory
+    /// traversal on the root of this matcher may provide relative paths to
+    /// this routine *without* calling [`IncrementalIgnore::normalize`].
```
MichaReiser, line 189: Is this an optimization that ty benefits from? I wonder how useful it is in the described use case of a file watcher and whether it's an API we can add later if needed.

crates/ignore/src/incremental.rs
```diff
@@ -0,0 +1,1041 @@
...
+    pub fn normalize<P: AsRef<Path>>(&self, path: P) -> Option<PathBuf> {
+        if matches!(self.ignore, RootIgnore::Stdin) {
+            return None;
+        }
+        let path = normalize_absolute(path.as_ref())?;
+        let root = self
+            .normalized_root
+            .get_or_init(|| normalize_absolute(&self.root))
+            .as_ref()?;
+        path.strip_prefix(root).ok().map(Path::to_path_buf)
```
MichaReiser, line 164: Is the reason that we allocate here that `AsRef` has no lifetime? Could we return a `Option<&Path>` otherwise and leave it up to the caller whether they want/need to clone

crates/ignore/src/incremental.rs
```diff
@@ -0,0 +1,1041 @@
...
+                .count();
+            satisfies_min =
+                self.options.min_depth.is_none_or(|min| depth >= min);
+            let satisfies_max =
+                self.options.max_depth.is_none_or(|max| depth <= max);
+            edge_max = self.options.max_depth.is_some_and(|max| depth == max);
+            // When we have a file that isn't past our min depth, we can give
+            // up right away.
+            if !is_dir && !satisfies_min {
+                return IncrementalMatch::ignore().no_within_depth();
```
MichaReiser: `not_within_depth`?

crates/ignore/src/incremental.rs
```diff
@@ -0,0 +1,1041 @@
...
+                    return IncrementalMatch::ignore();
+                }
+                Ok(_) => {}
+                Err(err) => {
+                    // Record the error but otherwise fall through
+                    let err = Error::from(err);
+                    errs.push(err.with_path(path));
+                }
+            }
+        }
```
MichaReiser: Should we move this after ignore matching, given that it always requires a file system access, whereas matching ignore and hidden only may require a file system access?

crates/ignore/src/incremental.rs
```diff
@@ -0,0 +1,1041 @@
...
+                    return IncrementalMatch::none(is_dir);
+                }
+                std::path::Component::CurDir => continue,
+                std::path::Component::Normal(_) => {}
+            }
+            if components.peek().is_none() {
+                break;
+            }
+            dir.push(component.as_os_str());
+            match self.dirs.get(&dir) {
```
MichaReiser, line 365: Here an idea for a perf optimization. Currently, we iterate over each component after `root`. I do believe this is unnecessary if a `self.dirs.get(normalized_relative.parent())` returns a cached `Ignored` or `Allowed`. We then already did the check whether all paths between `root` and `path` are allowed/ignored, so we can use that cached result. It's only a cache miss where we need to do all the work of checking every single component. It also avoids the `dir` allocation

crates/ignore/src/incremental.rs
```diff
@@ -0,0 +1,1041 @@
...
+            .peekable();
+        if components.peek().is_none() {
+            return IncrementalMatch::none(is_dir);
+        }
+
+        let mut dir = PathBuf::new();
+        let Some(mut ignore) = self.root_ignore(errs) else {
+            return IncrementalMatch::none(is_dir);
+        };
+
```
MichaReiser: Nit: Move `dir` after the early return?

MichaReiser (approved):
This is great. 

I don't know what Rust versioning policy you use, but it may make sense to set up some CI to ensure we follow it (unless it's, whatever ripgrep uses). 

I look forward to the ty PR. It should help me get a better sense for how the API is used in a real world context. 

Codex did find a few issues. I haven't verified them myself, but we should look into them

**Hidden ancestor**

For 

```
project/
└── .hidden/
    └── visible.rs
```

The incremental matcher returns `true` when testing for `.hidden/visible.rs`. 

**Ignored directories and `should_descend`**

```
project/
├── .ignore
└── build/
    └── generated.rs 
```

The incremental matcher returns `true` for `build`. This may be more difficult to fix?


**API Contract for non normalized paths**

Matching `matcher.matched("a/../visible.rs", false)` can return `ignored` instead of `None` if `a` has already been tested and is ignored.

The same API limitation applies for `max_filesize`, and depth. 

I'm not too concerned about this. But it may be worth documenting that the result is not guaranteed to be `none` 

**Windows hidden-attribute check inspects the wrong file**

```rust
matcher.matched("src\\generated.rs", false)
```

Calls

```rust
Path::new("src\\generated.rs").metadata()
```

which is relative to the current working directory, not to root

**Follow symlinks mismatch**

```rust
builder.follow_links(false);
builder.max_filesize(Some(100));
```

```
project/
├── small-link -> large-target
└── large-target               # 500 bytes
```

```
normal walker → includes small-link
incremental   → ignores small-link
```

The issue is that `path.metadata` follows symlinks.


**Case insensitive file systems**

`normalize` can reject paths that only differ in casing. E.g. the `root` is 

```
/private/tmp/Mixed/Root
```

and the path is 

```
/PRIVATE/TMP/MIXED/ROOT/File.TXT
```

I think we should just document this requirement. 

**`.gitignore` casing on case insensitve file systems**

On a case-insensitive filesystem, create:

```
project/
├── .IGNORE
├── .GITIGNORE
├── .GIT/
├── .RGIGNORE
├── ignored-by-ignore
├── ignored-by-gitignore
└── ignored-by-custom
```

The filesystem successfully resolves:

```
.ignore
.gitignore
.git
.rgignore
```

to those entries.

parallel walking first enumerates the directory and does a case-sensitive comparison for `entry.file_name() == ".ignore"` etc. The incremental walker only tests if the file exists, which will return true regardless of casing on the file system. 

I'm not sure this is something we should bother about. To my knowledge, the only solution for this is to read the directory and also compare the entries. 

## Code reuse

Codex also claims that we can share a few more helpers between walking/incremental. I leave it up to you to decide whether you find this useful

```diff
diff --git a/crates/ignore/src/dir.rs b/crates/ignore/src/dir.rs
@@
-    pathutil::{is_hidden_entry, strip_prefix},
+    pathutil::{is_hidden_entry, is_hidden_path, strip_prefix},
@@
+    /// Like `matched_dir_entry`, but works with a path instead.
+    pub(crate) fn matched_path<'a>(
+        &'a self,
+        path: &Path,
+        is_dir: bool,
+    ) -> Match<IgnoreMatch<'a>> {
+        let m = self.matched(path, is_dir);
+        if m.is_none() && self.inner.opts.hidden && is_hidden_path(path) {
+            return Match::Ignore(IgnoreMatch::hidden());
+        }
+        m
+    }
@@
-    pub(crate) fn is_hidden(&self) -> bool {
-        self.opts.hidden
-    }

diff --git a/crates/ignore/src/incremental.rs b/crates/ignore/src/incremental.rs
@@
-    Error, Match, PartialErrorBuilder, dir::Ignore, pathutil::is_hidden_path,
+    Error, Match, PartialErrorBuilder, dir::Ignore, walk::skip_filesize,
@@
-    pub(crate) hidden: bool,
+    pub(crate) follow_links: bool,
@@
-        if !is_dir && let Some(max_filesize) = self.options.max_filesize {
+        let mut mat = self.matched_with_errors_ignore(relative, is_dir, errs);
+        if !mat.is_ignore()
+            && !is_dir
+            && let Some(max_filesize) = self.options.max_filesize
+        {
             let path = self.root().join(relative);
-            match path.metadata() {
-                Ok(md) if md.len() > max_filesize => {
-                    return IncrementalMatch::ignore();
-                }
-                Ok(_) => {}
-                Err(err) => {
-                    let err = Error::from(err);
-                    errs.push(err.with_path(path));
-                }
+            let metadata = if self.options.follow_links {
+                path.metadata()
+            } else {
+                path.symlink_metadata()
+            }
+            .ok();
+            if skip_filesize(max_filesize, &path, &metadata) {
+                mat = IncrementalMatch::ignore();
             }
         }
-
-        let mut mat = self.matched_with_errors_ignore(relative, is_dir, errs);
-        if mat.is_none() && self.options.hidden && is_hidden_path(relative) {
-            mat = IncrementalMatch::ignore();
-        }
@@
-            if ignore.matched(&path, true).is_ignore() {
+            if ignore.matched_path(&path, true).is_ignore() {
@@
-            ignore.matched(&path, is_dir).map(|_| ()),
+            ignore.matched_path(&path, is_dir).map(|_| ()),
@@
     fn from_match(mat: Match<()>, is_dir: bool) -> IncrementalMatch {
-        IncrementalMatch { mat, should_descend: is_dir, is_within_depth: true }
+        let should_descend = is_dir && !mat.is_ignore();
+        IncrementalMatch { mat, should_descend, is_within_depth: true }
     }

diff --git a/crates/ignore/src/walk.rs b/crates/ignore/src/walk.rs
@@
-            hidden: self.ig_builder.is_hidden(),
+            follow_links: self.follow_links,
@@
-fn skip_filesize(
+pub(crate) fn skip_filesize(
```

.github/workflows/ci.yml
BurntSushi: That's just part of this CI run. It uses the `--workspace` flag, so this will check `ignore` too, which is part of ripgrep's workspace.

BurntSushi: There's no listed MSRV in the `ignore/Cargo.toml` mostly because I just never put one there. :P

crates/ignore/src/pathutil.rs
BurntSushi: This is a `OsStr`, so there is only `as_encoded_bytes()`.  I think that's fine here. This is just ancient code that predates such niceties.

crates/ignore/src/incremental.rs
BurntSushi: The caller has to handle that. :P Indeed, your nice hack in ty does, and that part of your hack lives.

The alternative is that everyone always pays the cost of trying to find the matching root on every match call. Which is maybe fine, but seemed like an inappropriate thing to impose on callers at the library level. Particularly since I think there may be some use cases with only one root.

crates/ignore/src/incremental.rs
BurntSushi: Generally, no. If your file paths are already relative to the root, then you don't need to call normalize. But if they're absolute, then yeah, you need a normalize call. (This is what is done in ty in my branch.)

BurntSushi: Yes, much ceremony for a newtype wrapper here.

crates/ignore/src/incremental.rs
BurntSushi: Do you mean the other way around? No, `path` ought to be _relative_ to the root. The caller is responsible for upholding that precondition. We don't really have a way to check that. (Other than a newtype wrapper, but that wrapper would probably want an unchecked constructor for cases where the caller already knows their path is root relative.)

crates/ignore/src/incremental.rs
BurntSushi: I don't think it does no. And I honestly don't know how much it matters given all of the other work being done during matching. But I'd rather the API be a bit more flexible even if I don't have concrete use cases in front of me just yet.

It's also possible that matching becomes faster in the future, but not having this optimization at the API level imposes a constraint on some amount of work that has to be done during matching. I think. Like, ideally callers _don't_ pass absolute paths because they are bigger and require more work to handle. Generally... anyway.

crates/ignore/src/incremental.rs
BurntSushi: That's because `path` comes from `normalize_absolute` and that uses `std::path::absolute` which returns a `std::path::PathBuf`. e.g., You could pass a `/root/whatever/foo/./bar` and get back `foo/bar` assuming that `/root/whatever` is your root.

Maybe this is too defensive, I'm not sure. We could assume that root doesn't need normalization I think. Then we could do a simple path prefix strip. I'll try that out and see how it feels.

crates/ignore/src/incremental.rs
BurntSushi: Fair enough. I did it to be consistent with `no_descend`.

crates/ignore/src/incremental.rs
BurntSushi: Yup I think this makes sense to me! And it seems like a pretty common case. I've added this.

BurntSushi:
This PR is on crates.io in `ignore 0.4.30`.

PR #3475, opened 2026-07-16
ignore: fix deadlock when visitor panics

### LINKED ISSUE #3009: ignore: parallel walker occasionally hangs if visitor panics

BurntSushi:
When a visitor panics and the workers shutdown, we indicate as such and
quit everything. This effectively acts as a sort of poison that will
cause all other workers to stop even when in their idle loop.

Fixes #3009

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -1445,21 +1445,28 @@ impl WalkParallel {
         let quit_now = Arc::new(AtomicBool::new(false));
         let active_workers = Arc::new(AtomicUsize::new(threads));
         let stacks = Stack::new_for_each_thread(threads, stack);
+        // Collect all of the workers first. In the case that
+        // `builder.build()` panics, we want that to happen and
+        // propagate before we actually start to run any of the
+        // workers.
+        let workers: Vec<_> = stacks
+            .into_iter()
+            .map(|stack| Worker {
+                visitor: builder.build(),
+                stack,
+                quit_now: quit_now.clone(),
+                active_workers: active_workers.clone(),
+                max_depth: self.max_depth,
+                min_depth: self.min_depth,
+                max_filesize: self.max_filesize,
+                follow_links: self.follow_links,
+                skip: self.skip.clone(),
+                filter: self.filter.clone(),
+            })
+            .collect();
         std::thread::scope(|s| {
-            let handles: Vec<_> = stacks
+            let handles: Vec<_> = workers
                 .into_iter()
-                .map(|stack| Worker {
-                    visitor: builder.build(),
-                    stack,
-                    quit_now: quit_now.clone(),
-                    active_workers: active_workers.clone(),
-                    max_depth: self.max_depth,
-                    min_depth: self.min_depth,
-                    max_filesize: self.max_filesize,
-                    follow_links: self.follow_links,
-                    skip: self.skip.clone(),
-                    filter: self.filter.clone(),
-                })
                 .map(|worker| s.spawn(|| worker.run()))
                 .collect();
             for handle in handles {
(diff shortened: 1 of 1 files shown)


MichaReiser (approved):
Nice, this makes sense. 

Codex found one edge case. It's not relevant for ty but maybe something you want to fix regardless, while we're at it

> Codex
[P1] Handle visitor-builder panics after workers start
> `builder.build()` runs while earlier workers are already spawned. If the third build call panics, no `Worker` exists to set `quit_now`, while `active_workers` still includes unspawned workers; the scope hangs indefinitely. Construct all workers before spawning or install a parent-side shutdown guard.

PR #3476, opened 2026-07-16
Add rust-version to published sub-crates

### LINKED ISSUE #3473: Specify rust-version in Cargo.toml
Crates like [`ignore`](https://crates.io/crates/ignore) are missing MSRV metadata.

albatrossflyon-coder:
Fixes #3473

The workspace root `Cargo.toml` already declares `rust-version = "1.85"`, and every sub-crate uses `edition = "2024"` (which mechanically requires rustc >= 1.85), but none of the 9 published sub-crates carried their own `rust-version` field — so it was missing from their individual `Cargo.toml` on crates.io even though the workspace itself documents the MSRV.

Added `rust-version = "1.85"` to:
- `crates/cli` (grep-cli)
- `crates/globset` (globset)
- `crates/grep` (grep)
- `crates/ignore` (ignore)
- `crates/matcher` (grep-matcher)
- `crates/pcre2` (grep-pcre2)
- `crates/printer` (grep-printer)
- `crates/regex` (grep-regex)
- `crates/searcher` (grep-searcher)

Verified `cargo metadata --no-deps` and `cargo check --workspace` both succeed with no changes needed elsewhere.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,8 +24,8 @@ exclude = [
 ]
 build = "build.rs"
 autotests = false
-edition = "2024"
-rust-version = "1.96"
+edition.workspace = true
+rust-version.workspace = true
 
 [[bin]]
 bench = false
@@ -48,6 +48,10 @@ members = [
   "crates/searcher",
   "crates/ignore",
 ]
+
+[workspace.package]
+edition = "2024"
+rust-version = "1.96"
 
 [dependencies]
 anyhow = "1.0.75"
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -11,8 +11,8 @@ repository = "https://github.com/BurntSushi/ripgrep/tree/master/crates/cli"
 readme = "README.md"
 keywords = ["regex", "grep", "cli", "utility", "util"]
 license = "Unlicense OR MIT"
-edition = "2024"
-rust-version = "1.85"
+edition.workspace = true
+rust-version.workspace = true
 
 [dependencies]
 bstr = { version = "1.6.2", features = ["std"] }
--- a/crates/cli/Cargo.toml
+++ b/crates/cli/Cargo.toml
@@ -12,6 +12,7 @@ readme = "README.md"
 keywords = ["regex", "grep", "cli", "utility", "util"]
 license = "Unlicense OR MIT"
 edition = "2024"
+rust-version = "1.85"
 
 [dependencies]
 bstr = { version = "1.6.2", features = ["std"] }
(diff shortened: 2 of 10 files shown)


BurntSushi:
Can this use `rust-version.workspace = true`? That way we don't have to repeat all of the version numbers.

xtqqczze:
We don’t currently define `[workspace.package]` in the root `Cargo.toml`. We could introduce it just for `edition` and `rust-version`, but it would be better to do so as part of a broader refactoring.

albatrossflyon-coder:
Good call — pushed a follow-up commit that centralizes `edition`/`rust-version` in a new `[workspace.package]` table in the root `Cargo.toml`, with each of the 9 sub-crates now using `edition.workspace = true` / `rust-version.workspace = true` instead of hardcoding the values. Values are unchanged (2024 / 1.85), just no longer duplicated 10 times. `cargo check -p ripgrep` and `cargo metadata` both pass clean.

xtqqczze:
Conflicts with https://github.com/BurntSushi/ripgrep/commit/c035d23c26da7cf9d0eb96944b69a212240e17f1.

albatrossflyon-coder:
Rebased onto master to resolve the conflict with `c035d23` (MSRV bump to 1.96). `[workspace.package].rust-version` now reads `1.96` to match; `cargo check` and `cargo metadata` both pass clean on 1.97.1.

Cargo.toml
```diff
@@ -49,6 +49,10 @@ members = [
   "crates/ignore",
 ]
 
+[workspace.package]
```
xtqqczze, line 52: nit: Could we move `[workspace.package]` to the top of the file?

BurntSushi: Please move it above `[[bin]]` but below the `[package]` section.

BurntSushi:
I think the main problem preventing me from merging this PR is that Rust 1.96 is fine for ripgrep, but it might be too new for some of the library crates like `ignore`.

So I guess for `globset` and `ignore` (the only two non-`grep` crates), could you set their `rust-version` to something lower? Let's go with Rust 1.88 for now, since I think `ignore` will build for that.

albatrossflyon-coder:
Done — globset and ignore now pin `rust-version = \"1.88\"` directly instead of inheriting the workspace's 1.96. Verified both build clean under an actual Rust 1.88.0 toolchain (not just an assumption), so 1.88 is a real, confirmed floor for these two, not a guess.

BurntSushi:
I don't think it's necessarily a floor. But I think this is fine for now.

PR #3492, opened 2026-07-26
fix(ignore): preserve escaped trailing spaces in gitignore patterns

baima365-web:
## Summary

`.gitignore` lines like `foo\   ` (escaped space followed by more spaces) were over-trimmed: only a single trailing `\ ` was preserved, so the pattern collapsed to `foo\` and failed with `dangling '\'`.

Git keeps the escaped trailing space. Match that by trimming only **unescaped** trailing spaces.

### Fix

- Replace `line.trim_right()` with logic that only trims unescaped trailing spaces
- Preserve escaped spaces (preceded by backslash) while removing unescaped ones

Fixes #3485

### DIFF
--- a/crates/ignore/src/gitignore.rs
+++ b/crates/ignore/src/gitignore.rs
@@ -465,9 +465,17 @@ impl GitignoreBuilder {
         if line.starts_with("#") {
             return Ok(self);
         }
-        if !line.ends_with("\\ ") {
-            line = line.trim_right();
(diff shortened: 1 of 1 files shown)


ltrzesniewski:
I didn't read the surrounding code, but you should probably check if the number of preceding backslashes is odd.

MichaReiser:
Closing in favor of https://github.com/BurntSushi/ripgrep/pull/3515

PR #2566, opened 2023-07-23
feat: dont do needless search when we reach max depth

sigmaSd:
fix https://github.com/BurntSushi/ripgrep/issues/2565

I can improve the code, if in general it seems reasonable,
seems to work in my testing

PR #3496, opened 2026-07-29
ignore: skip loading unreachable ignore files

### LINKED ISSUE #2565: perf(ignore): don't search subdirs for git/ignore files if max depth is reached
When searching for git/ignore files (the one set with `.ignore` `.git_ignore` `.git_exclude`), the crate should not search the sub-directories if its at the max depth

example:
```
mkdir a
for ((i=1; i<=100; i++)); do     mktemp --directory XXX; done
```
a.rs
```rs
use ignore::WalkBuilder;

fn main() {
    let arg = std::env::args().nth(1).unwrap();
    let files: Vec<_> = WalkBuilder::new(&arg)
        .hidden(false)
        .follow_links(false) // We're scanning over depth 1
        .max_depth(Some(1))
        .build()
        .collect();
}
```
```
cargo b --release
strace target/release/a a
```

you can see that we do a lot of unneeded syscalls because we're reading git/ignore files even though we won't search inside those directorates
```
openat(AT_FDCWD, "/home/mrcool/dev/rust/others/ripgrep/crates/ignore/a/Vwe", O_RDONLY|O_NONBLOCK|O_CLOEXEC|O_DIRECTORY) = 4
newfstatat(4, "", {st_mode=S_IFDIR|0700, st_size=4096, ...}, AT_EMPTY_PATH) = 0
statx(AT_FDCWD, "/home/mrcool/dev/rust/others/ripgrep/crates/ignore/a/Vwe/.git", AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7fffdb02f030) = -1 ENOENT (Aucun fichier ou dossier de ce type)
statx(AT_FDCWD, "/home/mrcool/dev/rust/others/ripgrep/crates/ignore/a/Vwe/.ignore", AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7fffdb02ed60) = -1 ENOENT (Aucun fichier ou dossier de ce type)
statx(AT_FDCWD, "/home/mrcool/dev/rust/others/ripgrep/crates/ignore/a/Vwe/.gitignore", AT_STATX_SYNC_AS_STAT, STATX_ALL, {stx_mask=STATX_ALL|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFREG|0644, stx_size=0, ...}) = 0
openat(AT_FDCWD, "/home/mrcool/dev/rust/others/ripgrep/crates/ignore/a/Vwe/.gitignore", O_RDONLY|O_CLOEXEC) = 5
close(5)                                = 0
statx(AT_FDCWD, "/home/mrcool/dev/rust/others/ripgrep/crates/ignore/a/Vwe/.git/info/exclude", AT_STATX_SYNC_AS_STAT, STATX_ALL, 0x7fffdb02ed60) = -1 ENOENT (Aucun fichier ou dossier de ce type)
close(4)                                = 0
```
this is the best case (empty dirs), in practice its worse because directory can have those files, and this crate will have to issue read syscalls for them

without that search, it would be only 3 syscalls per directory

**suggestion:**
I'm suggestion to not search for those files inside subdirs if we reach the max depth
Unless I'm missing something it should keep the exact same functionality while having all the performance boost

**motivation:**
you can checkout https://github.com/helix-editor/helix/issues/7715 where disabling this behavior improved the walking from ~4000 syscalls to ~900 and from 20 seconds to 100 ms in my old pc with hdd drive
so ~3000 wasted syscalls for ~300 files

jif-oai:
Fixes #2565.

The serial walker currently loads ignore files even when a directory is already skipped or we've reached the max depth. Since we won't visit any children, that work seems unnecessary.

I see that #2566 tried to fix this by adding a new matcher method and threading an `is_leaf` flag through the ignore internals. This instead reuses the existing `add_child_with_entries(path, &[])`, so the change stays in `walk.rs` and the ignore stack still works as before. Let me know if the scoping is not important here

Local benchmarks show:
- 6–9% faster on a real repo.
- ~3× faster on nested ignored directories.
- 22–28× faster on ignore-heavy or sorted searches.
- No change to parallel traversal.

Also adds a regression test for malformed ignore files at the max depth

### DIFF
--- a/crates/ignore/src/walk.rs
+++ b/crates/ignore/src/walk.rs
@@ -636,6 +636,7 @@ impl WalkBuilder {
             it: None,
             ig_root: ig_root.clone(),
             ig: ig_root.clone(),
+            max_depth: self.max_depth,
             max_filesize: self.max_filesize,
             skip: self.skip.clone(),
             filter: self.filter.clone(),
@@ -1118,6 +1119,7 @@ pub struct Walk {
     it: Option<WalkEventIter>,
     ig_root: Ignore,
     ig: Ignore,
+    max_depth: Option<usize>,
     max_filesize: Option<u64>,
     skip: Option<Arc<Handle>>,
     filter: Option<Filter>,
@@ -1228,12 +1230,17 @@ impl Iterator for Walk {
                         self.it.as_mut().unwrap().it.skip_current_dir();
                         // Still need to push this on the stack because
                         // we'll get a WalkEvent::Exit event for this dir.
-                        // We don't care if it errors though.
-                        let (igtmp, _) = self.ig.add_child(ent.path());
+                        // Its ignore files cannot apply to any visited entry.
+                        let (igtmp, _) =
+                            self.ig.add_child_with_entries(ent.path(), &[]);
                         self.ig = igtmp;
                         continue;
                     }
-                    let (igtmp, err) = self.ig.add_child(ent.path());
+                    let (igtmp, err) = if self.max_depth == Some(ent.depth()) {
+                        self.ig.add_child_with_entries(ent.path(), &[])
+                    } else {
+                        self.ig.add_child(ent.path())
+                    };
                     self.ig = igtmp;
                     ent.err = err;
                     return Some(Ok(ent));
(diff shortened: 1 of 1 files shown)


crates/ignore/src/walk.rs
```diff
@@ -2427,6 +2434,27 @@ mod tests {
...
+            .build()
+            .find_map(|result| {
+                let entry = result.unwrap();
+                (entry.path() == leaf).then_some(entry)
+            })
+            .unwrap();
+
+        assert!(entry.error().is_none());
+        assert_paths(td.path(), &builder, &["leaf"]);
+    }
```
BurntSushi, line 2456: Nice, clever test.

BurntSushi (approved):
This makes sense to me! Thank you.

BurntSushi:
On my checkout of the chromium repository (which is loaded with gitignores), I do see a small but measurable improvement here:

```
$ hyperfine 'rg-master -j1 --max-depth 6 regex-internals' 'rg -j1 --max-depth 6 regex-internals'
Benchmark 1: rg-master -j1 --max-depth 6 regex-internals
  Time (mean ± σ):     867.7 ms ±   6.9 ms    [User: 377.1 ms, System: 488.8 ms]
  Range (min … max):   860.4 ms … 879.6 ms    10 runs

Benchmark 2: rg -j1 --max-depth 6 regex-internals
  Time (mean ± σ):     842.3 ms ±   3.7 ms    [User: 367.7 ms, System: 472.9 ms]
  Range (min … max):   836.4 ms … 846.8 ms    10 runs

Summary
  rg -j1 --max-depth 6 regex-internals ran
    1.03 ± 0.01 times faster than rg-master -j1 --max-depth 6 regex-internals
```

Nice find!

PR #3528, opened 2026-09-07
ignore: treat lone ! gitignore line as a no-op

### LINKED ISSUE #3527: A bare "!" line in a .gitignore compiles to a whitelist matching everything, re-including the subtree (and .git/)

```

features:+pcre2

```

#### Describe your bug.

never descends into.

```bash
d=$(mktemp -d); cd "$d"
git init -q
mkdir a
touch a/.venv

git check-ignore -v a/.venv
```

```
$ git check-ignore -v a/.venv
.gitignore:1:.venv	a/.venv

a/.gitignore
a/.venv
.gitignore

```

git handles it.

`a/.gitignore` can decide.

|---|---|---|

  doesn't.

#### Where I hit it

exactly this kind of file.

a candidate fix through it.

— Midas

wellparth:
Closes #3527

## Problem
A `.gitignore` line consisting of a bare `!` compiles (after stripping the negation marker) into an empty glob. That empty glob is then treated as needing the `**/` prefix (no literal slash), producing `**/` — a whitelist matching every path. Because last-match-wins in a gitignore, it re-includes the entire subtree below the file, even paths excluded by an ancestor `.gitignore` (and, when hidden files are enabled, `.git/` itself). `git` treats a bare `!` as a no-op: after stripping the `!` there is no pattern left.

## Fix
In `GitignoreBuilder::add_line`, skip the line early when it is empty after stripping the `!` (negation) marker and any leading `/`. This also covers a lone `/` line, which git likewise treats as a no-op.

## Tests
* Unit tests in `crates/ignore/src/gitignore.rs` encoding the issue's control table: bare `!` after an ignore rule does not cancel it; `!` before an ignore rule does not matter; a lone `!` ignores nothing.
* Regression test in `tests/regression.rs` (`r3527`) reproducing the issue end to end with `rg --files --hidden`.

## Verification
* `cargo test --workspace` passes (all suites).
* `cargo test --features pcre2 --test integration r3527` passes.
* Manual repro matches `git check-ignore`:
  * before: `rg --files --hidden` lists `a/.venv`
  * after: `a/.venv` is excluded; everything else unchanged


KaizenShogun:
I opened #3527, so take this as an interested party's measurement rather than a review — and partly as a heads-up: there are **two** PRs open for this bug, this one and #3531, off the same base commit (`3fce3b5b`), and neither thread mentions the other.

I built both branches and the base, and ran them against `git check-ignore` (2.55.0) over 8,953 queries derived from 33 repositories' real `.gitignore` files:

| build | divergences |
|---|---|
| `3fce3b5b` (base) | 11 |
| #3528 | **0** |
| #3531 | **0** |

Compared as sets rather than counts: same 11 fixed, nothing broken, and the two branches' result sets are identical. (All 11 come from `psf/black`'s nested test data, where a `!` line sits under an ancestor `.gitignore`.) The official 14.1.1 binary gives the same 11 as the base build, which is the control that says the harness isn't inventing anything.

The diffs aren't the same shape — #3531 returns early inside the `!` branch; this one returns after the whole prefix block, so it also covers a lone `/` and `!/` — but I couldn't make that difference visible. Over nine hand-built cases (`!`, `/`, `!/`, each alone and with a rule before or after it, plus a legitimate `!.venv` re-inclusion as a control) both branches answer identically and both match git. `3fce3b5b` already treats a lone `/` as a no-op.

So it's a tie on behaviour, and the choice is about what ships alongside: this PR has an integration test in `tests/regression.rs`, four unit tests and a CHANGELOG entry; #3531 has one unit test.

PR #3529, opened 2026-09-08
Implement glob-not `-G`

### LINKED ISSUE #3522: Feature: add `-G` to exclude files by glob pattern
A very common use case for me is to exclude some file or directory from a search: documentation, translations, etc.. Notice that these are under version control, so `.gitignore` doesn't help me.

I can of course use `-g '!docs'` for that, but because `!` has special meaning in bash, it needs extra quotes.

A simpler way, and analogous to `-t` vs. `-T`, would be to add `-G pattern` as an alias for `-g '!pattern'`.

Happy to try my hand at a PR if you think this is a good idea.

yilmazhasan:
New options  `-G` / `--glob-not` are added as a negation of `-g` / `--glob`:

Closes #3522

E.g. `-G tests` is equivalent to `-g '!tests'`

Moreover, negation of glob-not is also possible in case:

E.g. `-G '!tests' is equivalent to `-g tests``

This is inspired as analogous to how `-T/--type-not` relates to `-t/--type`


tests/misc.rs
```diff
@@ -338,6 +338,26 @@ rgtest!(glob_negate, |dir: Dir, mut cmd: TestCommand| {
     eqnice!("file.py:Sherlock\n", cmd.stdout());
 });
 
+rgtest!(glob_not, |dir: Dir, mut cmd: TestCommand| {
+    dir.create("sherlock", SHERLOCK);
+    dir.remove("sherlock");
```
ttencate: What's the point of creating a file and then immediately removing it again?

tests/misc.rs
```diff
@@ -338,6 +338,26 @@ rgtest!(glob_negate, |dir: Dir, mut cmd: TestCommand| {
...
+    dir.create("file.py", "Sherlock");
+    dir.create("file.rs", "Sherlock");
+    cmd.arg("-G").arg("*.rs").arg("Sherlock");
+
+    eqnice!("file.py:Sherlock\n", cmd.stdout());
+});
+
+rgtest!(glob_not_long, |dir: Dir, mut cmd: TestCommand| {
+    dir.create("sherlock", SHERLOCK);
+    dir.remove("sherlock");
```
ttencate: Same here.

crates/core/flags/defs.rs
```diff
@@ -2723,6 +2724,93 @@ fn test_glob_case_insensitive() {
...
+        r"Exclude files and directories matching GLOB."
+    }
+    fn doc_long(&self) -> &'static str {
+        r#"
+Exclude files and directories for searching that match the given glob. This
+always overrides any other ignore logic. Multiple globs may be used. Globbing
+rules match \fB.gitignore\fP globs. This is equivalent to \fB\-\-glob\fP with
+a \fB!\fP prefix, i.e. \flag{glob-not}\fB=\fP\fIGLOB\fP is the same as
+\fB\-\-glob=!\fP\fIGLOB\fP. If multiple globs match a file or directory, the
+glob given later in the command line takes precedence.
```
ttencate, line 2763: Maybe refer to the new flag in the docs for `-g` as well, for discoverability?

GUIDE.md
```diff
@@ -306,6 +306,19 @@ is a bit non-standard, but it was chosen to be consistent with how globs in
...
+which is exactly the same as `-g` except that it always negates the given
+glob. That is, `-G '*.toml'` is equivalent to `-g '!*.toml'`, but doesn't
+require quoting to protect the `!` from your shell:
+
+```
+$ rg lexopt -G '*.toml'
+[Results from files except for `*.toml`]
+```
+
+This is analogous to how `-T`/`--type-not` relates to `-t`/`--type`.
```
ttencate, line 320: Seems superfluous to me.

GUIDE.md
```diff
@@ -306,6 +306,19 @@ is a bit non-standard, but it was chosen to be consistent with how globs in
 `.gitignore` files, a `!` prefix means whitelist, and on the command line, a
 `!` means blacklist.)
 
+Since `!` is special to most shells, negated globs given to `-g` typically
+need to be quoted. As a convenience, ripgrep also provides `-G`/`--glob-not`,
+which is exactly the same as `-g` except that it always negates the given
+glob. That is, `-G '*.toml'` is equivalent to `-g '!*.toml'`, but doesn't
```
ttencate: This is maybe not the best example, since it also contains a `*` so it still needs to be quoted 🙂 How about `-G build`?

ttencate (commented):
Thanks! Did a quick review pass, hope it's useful.

tests/misc.rs
yilmazhasan: Couldn't be sure, as it was used in `file_types_negate`.

There looks no point of having that though.

Maybe it's worth to remote that from `file_types_negate` too.

crates/core/flags/defs.rs
yilmazhasan: That's added to `-g` docs with an example

GUIDE.md
yilmazhasan: In my shell, I didn't need quote for `*` though.

Still it makes sense. Just to be on safe side, used a new example `tests`.

ttencate: Yeah, at least bash by default only expands * if it matches any files. So it works depending on what's in your current working directory...

PR #3531, opened 2026-09-12
ignore: treat a bare "!" gitignore line as a no-op, not a whitelist-e…

### LINKED ISSUE #3527: A bare "!" line in a .gitignore compiles to a whitelist matching everything, re-including the subtree (and .git/)

voidstackloop:
…verything

`GitignoreBuilder::add_line` stripped the leading `!` to detect a negated (whitelist) pattern, then used whatever was left as the glob. For a line that is just `!` with nothing after it, that leaves an empty pattern, which the "no literal slash" branch then prefixes with `**/`, producing a glob that matches every path and is marked `is_whitelist: true`.

Concretely, in a subdirectory's `.gitignore`, a bare `!` line re-included everything below it, overriding an ancestor `.gitignore`'s exclusions and letting the walker descend into `.git/` itself, which it otherwise never does.

git treats this line as a no-op: stripping `!` leaves no pattern, so nothing is negated and the parent's rules keep applying. Match that: return early (skip adding a glob at all) when the stripped line is empty, the same way an all-whitespace or comment-only line is already skipped earlier in this function.

Added a regression test (`ig45`) and verified the exact repro from the issue end-to-end with the built `rg` binary: a subdirectory's bare `!` no longer resurrects a file excluded by an ancestor `.gitignore`.

Closes #3527


KaizenShogun:
I opened #3527, so take this as an interested party's measurement rather than a review — and partly as a heads-up: there are **two** PRs open for this bug, this one and #3528, off the same base commit (`3fce3b5b`), and neither thread mentions the other.

I built both branches and the base, and ran them against `git check-ignore` (2.55.0) over 8,953 queries derived from 33 repositories' real `.gitignore` files:

| build | divergences |
|---|---|
| `3fce3b5b` (base) | 11 |
| #3528 | **0** |
| #3531 | **0** |

Compared as sets rather than counts: same 11 fixed, nothing broken, and the two branches' result sets are identical. (All 11 come from `psf/black`'s nested test data, where a `!` line sits under an ancestor `.gitignore`.) The official 14.1.1 binary gives the same 11 as the base build, which is the control that says the harness isn't inventing anything.

The diffs aren't the same shape — this one returns early inside the `!` branch; #3528 returns after the whole prefix block, so it also covers a lone `/` and `!/` — but I couldn't make that difference visible. Over nine hand-built cases (`!`, `/`, `!/`, each alone and with a rule before or after it, plus a legitimate `!.venv` re-inclusion as a control) both branches answer identically and both match git. `3fce3b5b` already treats a lone `/` as a no-op, so the narrower guard costs nothing here.

So it's a tie on behaviour, and the choice is about what ships alongside: #3528 has an integration test in `tests/regression.rs`, four unit tests and a CHANGELOG entry; this one has one unit test.

MichaReiser:
See https://github.com/BurntSushi/ripgrep/issues/3527#issuecomment-5756924252

PR #3536, opened 2026-09-16
hyperlink: percent-encode invalid UTF-8 bytes in file URLs

### LINKED ISSUE #3526: Hyperlink URI contains raw non-UTF-8 path bytes on Unix

Reproduced with `ripgrep 14.1.0` (Ubuntu 24.04, apt). The relevant code path looks unchanged on master as of today — see the two links below.

`apt install ripgrep` inside `ubuntu:24.04`. Also checked 15.1.0 (Homebrew) on macOS, where the case cannot be constructed because APFS rejects filenames that are not valid UTF-8 (`EILSEQ`). So this is Unix-with-a-permissive-filesystem only.

Linux (any filesystem that allows arbitrary bytes in a filename).

### Describe your bug.

On Unix, path bytes that are **not valid UTF-8** are copied verbatim into the hyperlink URI. The result is neither a URI (RFC 3986 is ASCII-only) nor an IRI (RFC 3987 needs decodable characters), and it puts a byte outside 32–126 into an OSC 8 sequence, which [the OSC 8 specification linked from `--hyperlink-format`'s own man page entry](https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda) says is undefined behaviour: *"the parameters and the URI must not contain any bytes outside of the 32–126 range. If they do, the behavior is undefined. Bytes outside of this range in the URI must be URI-encoded."*

I want to be clear that I am **not** reporting the general non-ASCII behaviour. `HyperlinkPath::encode` leaves `128..` unencoded on purpose and the comment above it explains why — RFC 8089 §4 does not mandate an encoding, and `UrlCreateFromPathW` does not encode non-ASCII on Windows. For a valid UTF-8 path that reasoning holds and produces a usable IRI. This report is about the narrower case that reasoning does not cover: bytes that are not part of any valid UTF-8 sequence, where there is nothing to be lenient *about*.

```bash
docker run --rm ubuntu:24.04 bash -c '
  apt-get update -qq && apt-get install -y -qq ripgrep
  cd "$(mktemp -d)"
  printf "hit\n" > "$(printf "bad\xff.txt")"
  rg --color=always --hyperlink-format=default -H hit . | od -An -c | head -3'
```

```
 033   ]   8   ;   ;   f   i   l   e   :   /   /   4   1   1   f
   b   8   2   5   a   6   0   1   /   t   m   p   /   t   m   p
   .   l   Q   D   o   a   c   c   x   q   n   /   b   a   d 377
```

The `377` at the end is the raw `0xFF` from the filename, inside the URI.

Either of these would resolve it, and both are consistent with the existing design:

1. **Percent-encode only the bytes that cannot be part of valid UTF-8.** `%FF` is lossless, unambiguous, round-trips exactly, and is explicitly permitted by RFC 8089. It does not affect the documented non-ASCII behaviour, because a valid UTF-8 path contains no such byte — so the Windows rationale in the comment is untouched.
2. **Refuse to build a hyperlink**, which is what the Windows branch already does for the analogous case: `from_path` bails with *"path is not valid UTF-8"* when `to_str()` returns `None`. The Unix branch takes `path.as_os_str().as_bytes()` with no equivalent check, so the two platforms disagree about a case they both can describe.

Relevant code, for reference:

- [`HyperlinkPath::encode`](https://github.com/BurntSushi/ripgrep/blob/master/crates/printer/src/hyperlink/mod.rs) — the `128..` arm of the match
- the `#[cfg(unix)]` `from_path` just above it, versus the `#[cfg(windows)]` one that gives up on non-UTF-8

### One small documentation point

The `--hyperlink-format` man page entry says of `{path}`:

> The path is guaranteed to be absolute and percent encoded such that it is valid to put into a URI.

For any non-ASCII path that is not accurate — the result is an IRI, not a URI — and the guarantee is what led me to check this in the first place. The doc comment on `encode` is precise about what actually happens; the user-facing text is not. Softening it (or pointing at RFC 8089 the way the code comment does) would help, independently of whatever you decide about the invalid-UTF-8 case.

Thanks for ripgrep — this came out of writing up how `rg` handles text encodings, and the rest of what I measured was exactly as documented.

nikolas-sapa:
Hyperlink path encoding copied every byte >= 0x80 raw into the OSC 8 URL, so filenames with undecodable bytes (e.g. 0xFF) produced raw non-UTF-8 output in the terminal sequence.

This keeps valid UTF-8 sequences as-is (IRI behavior per RFC 8089, matching UrlCreateFromPathW on Windows) and percent-encodes only bytes that cannot form valid UTF-8: lone continuations, truncated sequences, overlongs, surrogates, and 0xF5-0xFF. The --hyperlink-format docs are softened to match. Covered by a new unit test (encode_percent_encodes_invalid_utf8); I could not run an invalid-byte end-to-end test on APFS, which rejects such filenames, so unit coverage is the proof here.

Fixes #3526.

> Note: I used an AI coding assistant for drafting; the change and description above are my own and I have verified the tests.


ltrzesniewski (commented):
I won't comment on the code itself, but I found some test files back from the day, and I can confirm the hyperlinks still work correctly on Windows.

Image

This monstrosity also works fine:

image

nikolas-sapa:
Thanks for testing this on Windows — good to know both the standard and the heavily-encoded path render as working hyperlinks there. That matches the intent of the change: bytes forming valid UTF-8 stay unencoded (IRI-style, like `UrlCreateFromPathW`), and only bytes that cannot be part of valid UTF-8 get percent-encoded, since raw bytes outside 32-126 are undefined behavior in an OSC 8 sequence.

nikolas-sapa:
Thank you for testing the old cases on Windows — really helpful to have those confirmed still working.

PR #3539, opened 2026-09-19
ignore: ignore empty gitignore negation patterns

### LINKED ISSUE #3527: A bare "!" line in a .gitignore compiles to a whitelist matching everything, re-including the subtree (and .git/)

BrocodeADI:
Fixes #3527

### Summary

A bare `!` in a `.gitignore` pattern was parsed as an empty pattern after
the negation prefix was stripped.

The empty pattern then fell through to glob construction and was expanded
to `**/` with whitelist semantics, causing the bare `!` to match entries.

Git treats a negation with no remaining pattern as a no-op.

This change:

- returns early when the pattern becomes empty after stripping `!`/`/`
- preserves normal negation patterns
- adds regression coverage for `!`, `!/`, ordering, and escaped `\!`

### Testing

- `cargo test -p ignore`
- `cargo test -p globset`
- `cargo fmt --check`
- `cargo clippy -p ignore --all-targets`
- `git diff --check`
- `cargo test -p ripgrep --test integration`


MichaReiser:
See https://github.com/BurntSushi/ripgrep/issues/3527#issuecomment-5756924252

